PR #55, opened 2021-06-11
WIP: actions: build python package and docker image

thijs-obj:
Change: Build python package and docker image on each push.

This is still **WIP** because I am not sure what to do with this. Two concerns:
1) Might not be that useful to build the python package on every push.
2) Perhaps we should only do this on merges to master. Perhaps we should also add some caching of the docker layers, that should be doable by caching /var/lib/docker, and having the current date in the cache key.

Still thinking a bit about both


thijs-obj:
closing/removing as part of clean-up

PR #270, opened 2021-11-09
Bach: Create Sample on filter

jansentom:
Allow creating a sample using a custom filter vs. Bernoulli sample.
```
bt_sample = bt.get_sample(table_name='test_data_sample', filter=bt.skating_order % 2 == 0, overwrite=True)
# do super fast stuff
bt_full = bt.get_unsampled()
bt_full.head(10)  # run on ful dataset, epic.
```

### DIFF
--- a/bach/bach/dataframe.py
+++ b/bach/bach/dataframe.py
@@ -404,10 +404,15 @@ class DataFrame:
             exists.
         :param: seed: seed number used to generate the sample.
         """
+        original_node = self.get_current_node(name='before_sampling')
+        df = self
         if filter:
-            df = self[filter]
+            from bach.series import SeriesBoolean
+            if not isinstance(filter, SeriesBoolean):
+                raise TypeError('Filter parameter needs to be a SeriesBoolean instance.')
+            # Boolean argument implies return type of self[filter], help mypy a bit
+            df = cast('DataFrame', self[filter])
         elif sample_percentage is None:
-            df = self
             sample_percentage = 50
 
         if df._group_by is not None:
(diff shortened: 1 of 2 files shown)


bach/bach/dataframe.py
```diff
@@ -396,40 +397,53 @@ def get_sample(self,
...
+            if sample_percentage:
+                sql = f'''
+                    create temporary table tmp_table_name on commit drop as
+                    ({df.view_sql()});
+                    create table {table_name} as
+                    (select * from tmp_table_name
+                    tablesample bernoulli({sample_percentage}) repeatable ({seed}))
+                '''
+            else:
+                sql = f'create table {table_name} as ({df.view_sql()})'
```
jansentom, line 435: @thijs-obj can you make this a temp table before merging? That way we can use it in the web demo notebook without too aweful permissions for the user.

jansentom: nevermind, did it already

bach/bach/dataframe.py
```diff
@@ -406,7 +407,7 @@ def get_sample(self,
         """
         original_node = self.get_current_node(name='before_sampling')
         df = self
-        if filter:
+        if filter is not None:
```
thijs-obj, line 410: not strictly needed, but found it weird to read `if filter` for something where `filter` is sort of a boolean. A boolean could be false, of course a `SeriesBoolean` is actually always True, but still a bit weird to read imho

thijs-obj (approved):

bach/bach/dataframe.py
jansentom: makes sense, thanks.

PR #274, opened 2021-11-09
Bach: Add Series.unique() and Series[slice]

jansentom:
giving us:
```
# get the unique values
bt.municipality.unique().head()

# slice the first three 
first_muni = bt.municipality[:3] 

# or the first three unique ones
first_unique_muni = bt.municipality.unique()[:3]
```

Update: removed value counts from this MR, as it had quite some work to be done and unique() is ready

### DIFF
--- a/bach/bach/partitioning.py
+++ b/bach/bach/partitioning.py
@@ -290,8 +290,6 @@ class Window(GroupBy):
             self._frame_clause = f'{mode.name} BETWEEN {start_boundary.frame_clause(start_value)}'\
                                 f' AND {end_boundary.frame_clause(end_value)}'
 
-        if not isinstance(order_by, list) or not all(isinstance(ob, SortColumn) for ob in order_by):
-            raise ValueError("order_by must be a list of SortColumns")
         self._order_by = order_by
 
     @property
@@ -340,8 +338,6 @@ class Window(GroupBy):
         """
         Given the window_func generate a statement like:
             {window_func} OVER (PARTITION BY .. ORDER BY ... frame_clause)
-        :note: Currently, there is no single value expression detection here, so this might break
-            anything depending on that.
         """
         # TODO implement NULLS FIRST / NULLS LAST, probably not here but in the sorting logic.
         order_by = self._get_order_by_expression()
--- a/bach/bach/partitioning.py
+++ b/bach/bach/partitioning.py
@@ -290,6 +290,8 @@ class Window(GroupBy):
             self._frame_clause = f'{mode.name} BETWEEN {start_boundary.frame_clause(start_value)}'\
                                 f' AND {end_boundary.frame_clause(end_value)}'
 
+        if not isinstance(order_by, list) or not all(isinstance(ob, SortColumn) for ob in order_by):
+            raise ValueError("order_by must be a list of SortColumns")
         self._order_by = order_by
 
     @property
@@ -338,6 +340,8 @@ class Window(GroupBy):
         """
         Given the window_func generate a statement like:
             {window_func} OVER (PARTITION BY .. ORDER BY ... frame_clause)
+        :note: Currently, there is no single value expression detection here, so this might break
+            anything depending on that.
         """
         # TODO implement NULLS FIRST / NULLS LAST, probably not here but in the sorting logic.
         order_by = self._get_order_by_expression()
--- a/bach/bach/series/series.py
+++ b/bach/bach/series/series.py
@@ -385,13 +385,20 @@ class Series(ABC):
         :note: This will maintain Expression.is_single_value status
         """
         # This will give us a dataframe that contains our series as a materialized column in the base_node
-        df = series.to_frame().materialize()
-        fmt_string = f'{operation} (SELECT {{}} FROM {{}})' if operation else f'(SELECT {{}} FROM {{}})'
-        expr = IndependentSubqueryExpression.construct(fmt_string,
-                                                       Expression.column_reference(series.name),
-                                                       Expression.model_reference(df.base_node))
+        if series.expression.is_independent_subquery:
+            expr = series.expression
+        else:
+            df = series.to_frame()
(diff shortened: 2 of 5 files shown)


bach/bach/series/series.py
```diff
@@ -845,6 +854,57 @@ def nunique(self, partition: WrappedPartition = None, skipna: bool = True):
...
+        :param skipna: Skip na values, automatically done by bins mode
+        :note: this returns a Series with a different name: value_counts or frequency
+        :note: this returns a Series grouped by either the value or the bin
+        :note: binning works based on n-tiles, all values types that can be sorted, can be binned
+        """
+        from bach.partitioning import GroupBy, Window
+        if partition:
+            raise ValueError('Can not use group_by in combination with value_counts(). Materialize() first.')
+        if bins:
+            value_counts = self.window_ntile(
```
jansentom, line 893: This kinda combines everything we've built. Pretty neat.

bach/bach/partitioning.py
```diff
@@ -290,6 +290,8 @@ def __init__(self,
             self._frame_clause = f'{mode.name} BETWEEN {start_boundary.frame_clause(start_value)}'\
                                 f' AND {end_boundary.frame_clause(end_value)}'
 
+        if not isinstance(order_by, list) or not all(isinstance(ob, SortColumn) for ob in order_by):
```
thijs-obj, line 293: Is this just a precaution, or something that actually went wrong in some instance? If the latter, then when should perhaps have the same check in DataFrame

bach/bach/series/series.py
```diff
@@ -845,6 +854,57 @@ def nunique(self, partition: WrappedPartition = None, skipna: bool = True):
...
+                     normalize: bool = False, sort: bool = True,
+                     ascending: bool = False, bins: int = None,
+                     skipna: bool = True):
+        """
+        Get the unique values in this series and how often they occur, or get binned frequency
+        :param normalize: default False, if True then the object returned will contain the relative
+                frequencies of the unique values.
+        :param sort: default True: sort by frequencies.
+        :param ascending: default False, Sort in ascending order.
+        :param bins: Rather than count values, group them into half-open bins.
```
thijs-obj, line 883: We should mention here that this binning is different from pandas. I think it's better than pandas, but people might disagree :)

It seems pandas takes the min and max of the series, and then creates the specified number of buckets with each bucket covering a similarly wide range of values regardless of how many values will end up in each bucket. Whereas we create buckets with the same number of values in each bucket.

Example:
values: `1, 2, 3, 4, 5, 10`
number of bins: 2

pandas result: bucket one (range 1 to 5): `1, 2, 3, 4, 5` and  bucket two (range 6 to 10): `10`
our result: bucket one (lowest three values): `1, 2, 3` and bucket two (highest three values): `4, 5, 10`

thijs-obj: This also explains why pandas has the `normalize` parameter: one bucket might contain much more values than the next. Whereas in our case the normalized values will be pretty useless, e.g. if there are `N` items and `X` buckets and `N` is a multiple of `X` then the normalized value is always `N/X` in our implementation and if `N` is not a multiple of `X` then it is close to that value.

thijs-obj: perhaps we should check with @hendrik-obj what to do with this

bach/bach/series/series.py
thijs-obj: This is pretty neat indeed

bach/tests/functional/bach/test_series.py
```diff
@@ -43,6 +43,37 @@ def test_series__getitem__():
         non_existing_value_ref.value
 
 
+def test_positional_slicing():
```
thijs-obj, line 46: Below I wrote a comment that a test relies on a certain data-ordering which is not guaranteed by the database, same goes for this test. There is no order by clause in the generated sql

bach/bach/partitioning.py
jansentom: I made a mistake in value_counts() that took me a while to figure out. Could have also run mypy probably

thijs-obj (changes requested):
I think we need to make some improvements to the docstring and tests and/or have sorting by default

bach/bach/series/series.py
```diff
@@ -845,6 +854,57 @@ def nunique(self, partition: WrappedPartition = None, skipna: bool = True):
...
             skipna=skipna)
 
+    def unique(self, partition: WrappedPartition = None, skipna: bool = True):
+        """
+        Get the unique values in this series, currently by grouping the series involved.
+        :note: this returns a Series with a different name; _unique gets appended
+        :note: this retruns a Series with a different grouping, namely on this series.
+        """
+        from bach.partitioning import GroupBy
+        if partition:
```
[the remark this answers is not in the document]
jansentom: yes, and we later might support it. It's possible, but a lot of work (because we have to stack group_by's)

bach/tests/functional/bach/test_series.py
```diff
@@ -172,6 +203,101 @@ def test_type_agnostic_aggregation_functions():
...
+            ['Súdwest-Fryslân', 'Súdwest-Fryslân'], ['Harlingen', 'Harlingen'], ['Waadhoeke', 'Waadhoeke'],
+            ['De Friese Meren', 'De Friese Meren']
+        ]
+    )
+
+
+def test_unique_subquery():
+    bt = get_bt_with_test_data(full_data_set=True)
+
+    uq = bt.municipality.unique()[:2]
```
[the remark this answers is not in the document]
jansentom: (I actually wrote the above comment yesterday, and did some thinking):
We need a solution for ordering that is multi-column, even for a single series. I've added a comment in the docstring for now, and put the ordering on the TODO list)

jansentom:
I've removed value_counts(), as there was too much discussion on it. Merging the rest.

PR #275, opened 2021-11-09
wip: Improve docs sankey

hendrik-obj:


vincenthoogsteder (approved):
liked it, very clear now. Will also update it in the live demo NB

PR #278, opened 2021-11-09
fix sankey clicking

hendrik-obj:


vincenthoogsteder:
tested, works nice, lgtm

PR #282, opened 2021-11-10
Bach/Analysis: add fix for mixed environments for README.md

jansentom:


jansentom:
Merging this, can do no harm as we need to rename all packages very soon anyway.

PR #284, opened 2021-11-11
Objectiv bach analytics tweaks

thijs-obj:
This is not quite finished, as discussed this morning: we should rename all of this probably to objectiv_bach_analytics or something like that. Did not yet quite have time for that


analysis/objectiv_bach/objectiv_bach/__init__.py
```diff
@@ -5,9 +5,3 @@
 
 from objectiv_bach.series import *
 from objectiv_bach.stack.util import basic_feature_model
-from bach.types import _registry
-
-# automatically register our own types
-# TODO use proper decorator for this.
```
thijs-obj, line 11: I added the decorators.

Now that I used it for the first time, I am doubting a bit whether this is the way to go tho. As we still need to do some import that load the `series_objectiv.py` file (even if indirectly) and then suddenly the types are registered, it's a bit heavy on the magic.

thijs-obj:
small tweaks, please have a look @jansentom

analysis/objectiv_bach/objectiv_bach/__init__.py
jansentom: I like it actually. Better than it was, and if you import this (indirectly), you need the types to be registered for this to be useful.

analysis/objectiv_bach/requirements.txt
```diff
@@ -3,5 +3,5 @@ psycopg2-binary
 sqlalchemy
 pandas
 numpy
-bach
+-e ../../bach
```
jansentom, line 6: oh, nice.

jansentom (approved):
Better than it was. Added the discussion about the magic to our TODO

PR #285, opened 2021-11-11
Bach improve escaping

thijs-obj:


bach/bach/dataframe.py
```diff
@@ -1132,6 +1132,7 @@ def to_pandas(self, limit: Union[int, slice] = None) -> pandas.DataFrame:
             sql = self.view_sql(limit=limit)
             dtype = {name: series.dtype_to_pandas for name, series in self.all_series.items()
                      if series.dtype_to_pandas is not None}
+            sql = sql.replace('%', '%%')
```
thijs-obj, line 1135: WIP: we should do this in more places. Just fixing it in view_sql is also not really a solution cause we might use the output of that in another way too.

bach/bach/dataframe.py
```diff
@@ -1513,3 +1522,14 @@ def dict_name_series_equals(a: Dict[str, 'Series'], b: Dict[str, 'Series']):
             len(a) == len(b) and list(a.keys()) == list(b.keys())
             and all(ai.equals(bi) for (ai, bi) in zip(a.values(), b.values()))
     )
+
+
+def _escape_parameter_characters(conn: Connection, raw_sql: str) -> str:
```
thijs-obj, line 1527: Maybe this is a bit overrefactored future proofing. From the connection we can tell what the escape character is, so that's why I added it here as a parameter.

But then for  now we'll just always assume it is '%', and we ignore the `conn` parameter. Because if we actually look at the connection, then that might break in our online sandboxed notebook where we use a fake connection object.

Perhaps there is also some magic builtin function that does the same as this, but couldn't find it just now.

jansentom (approved):
LGTM, nice improvement

PR #288, opened 2021-11-16
Readme docs alignment

jansenbob:
I overhauled the main readme to align with the docs changes.

- removed the bulleted key features in favor of a better introduction with an image
- moved the installation instructions to the docs to have one place to maintain it
- made a more compact layout for the meta information to make it visually a bit less distracting

Note: the pipeline image is rather large at this moment. this will be fixed in the docs (as that's where it comes from)

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ Check out [objectiv.io](https://www.objectiv.io) to learn more.
 
 * [Objectiv Docs](https://www.objectiv.io/docs) - Objectiv's official documentation.
 * [Objectiv on Slack](https://join.slack.com/t/objectiv-io/shared_invite/zt-u6xma89w-DLDvOB7pQer5QUs5B_~5pg) - Get help & join the discussion on where to take Objectiv next.
-* [Contribution Guide](CONTRIBUTING.md) - Report bugs, request features and contribution information.
+* [Contribution Guide](https://www.objectiv.io/docs/the-project/contribute) - Report bugs, request features and contribution information.
 * [Github Roadmap](https://github.com/objectiv/objectiv-analytics/projects/2) - Learn about Objectiv's future plans.
 
 
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ with the goal of collecting better data and more effective modeling.
 
 ![Objectiv Pipeline](https://www.objectiv.io/docs/img/objectiv-pipeline.svg "Objectiv Pipeline")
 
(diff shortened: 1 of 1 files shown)


ivarpruijn (approved):
Looks good!

README.md
```diff
@@ -1,5 +1,3 @@
-Go to [Objectiv Docs](https://objectiv.io/docs/) for detailed installation & usage instructions
-
 ![Objectiv Logo](https://objectiv.io/docs/img/logo-objectiv-large.svg "Objectiv Logo")
```
jansentom, line 1: imho, this is not very developer friendly. Okay to have this in the main README, but the rest should be image-free (unless ascii art, that rules)

README.md
```diff
@@ -8,66 +6,26 @@ structure and validate data. With Objectiv, you create a
...
-The below commands assume that you have `docker-compose` [installed](https://docs.docker.com/compose/install/).
-```bash
-docker-compose pull  # pull pre-built images from gcr
-docker-compose up    # spin up Objective pipeline
-```
-This will spin up three images:
-* `objectiv_collector` - Endpoint that the Objectiv-tracker can send events to (http://localhost:5000).
-* `objectiv_postgres` - Database to store data.
-* `objectiv_notebook` - Jupyter notebook that can be used to query the data (http://localhost:8888).
+![Objectiv Pipeline](https://www.objectiv.io/docs/img/objectiv-pipeline.svg "Objectiv Pipeline")
```
jansentom, line 9: an sich a nice figure, but I'd write "data store" instead of "postgres data store", to make it more future proof and attractive.

README.md
```diff
@@ -8,66 +6,26 @@ structure and validate data. With Objectiv, you create a
...
 
-SECURITY WARNING: The above `docker-compose` commands start a postgres container that allows connections
-without verifying passwords. Do not use this in production or on a shared system!
+With [Objectiv Bach](https://www.objectiv.io/docs/modeling/) (our data modeling library), you can use familiar Pandas-like DataFrame operations on your full SQL dataset. It is designed for effective feature creation for datasets that embrace [the open taxonomy of analytics](https://objectiv.io/docs/taxonomy).
```
jansentom, line 11: `It is designed for effective feature creation for datasets that embrace` 
If you mean the Objectiv extensions to Bach, yes. Else we should leave the last part out or make it more specific.

jansentom (approved):
Nice improvement, ship it (with maybe a minor fix)

README.md
jansenbob: agree that we should only do this in main readme. that's the only place it is used, so ack.

README.md
jansenbob: Agree. will fix

README.md
jansenbob: `It includes extensions for effective feature creation for datasets that embrace the open taxonomy of analytics`

How about that?

PR #292, opened 2021-11-18
WIP: Create better demo docker compose

borft:
Creates stack with:
- website container (actual objectiv.io website, instrumented with the tracker)
- collector
- postgres, to store / query events
- notebook with product-analytics


docker-compose.yaml
```diff
@@ -42,12 +57,8 @@ services:
   objectiv_postgres:
     container_name: objectiv_postgres
     image: postgres:latest
-    ports:
-      - 127.0.0.1:5432:5432
```
borft, line 46: removed default port for postgres. This means you have to docker exec into the container to access the db directly.

It also means less problems with ports not being available, because there's already something listening on 5432

docker-compose.yaml
```diff
@@ -42,12 +57,8 @@ services:
...
     networks:
       - obj
-    volumes:
```
borft, line 49: no need to persist data, it gets populated at startup anyways. It also means every start-up starts with a fresh db

docker-compose-dev.yaml
```diff
@@ -0,0 +1,88 @@
+# Start the Objectiv Pipeline.
```
borft, line 1: this is the "old" docker-compose file, more suited for dev work, as there's no app (so no new data), and the default db is empty.

analysis/docker/Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM python:3.9
```
borft, line 1: new dockerfile to build a jupyter container, it only installs what we need, so it's about half the size of the "standard" notebook container. (2GB in stead of 4GB)

analysis/docker/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+    apt -q update && \
+    apt -qy upgrade && \
+    apt -qy install postgresql-client && \
+    apt clean && \
+    python3 -m venv /services/venv && \
+    . /services/venv/bin/activate && \
+    python -m pip install --upgrade pip && \
+    useradd --user-group --system --create-home --no-log-init notebook
+
+COPY analysis/docker/*.sql  /services
```
borft, line 17: copies sql files into the container. This is where you need to put the most recent dump, in order for it to be packaged

analysis/docker/entrypoint.sh
```diff
@@ -0,0 +1,18 @@
...
+# wait a bit to give PG a chance to start up
+sleep 10
+for sql in /services/*.sql
+do
+  echo "Loading data from $sql into $POSTGRES_HOSTNAME/$POSTGRES_DB"
+  cat $sql | psql -U $POSTGRES_USER -h $POSTGRES_HOSTNAME $POSTGRES_DB
+done
+
+# init db
+export DSN="postgresql://${POSTGRES_USER}:${POSTGRES_PASSWORD}@${POSTGRES_HOSTNAME}/${POSTGRES_DB}"
```
borft, line 15: The nb reads this variable from the environment, so it knows how to connect. 1 var is a little less noise/boilerplate than 5

docker-compose.yaml
```diff
@@ -16,16 +16,31 @@ x-db_env_variables: &db_env_variables
...
+      interval: 30s
+      timeout: 2s
+      retries: 3
+
   objectiv_collector:
     container_name: objectiv_collector
     depends_on:
       - objectiv_postgres
     # default image: eu.gcr.io/objectiv-production/backend:latest
-    image: ${OBJECTIV_CONTAINER_URL-eu.gcr.io/objectiv-production}/backend:${OBJECTIV_CONTAINER_TAG-latest}
```
borft, line 24: Change default URL to use dockerhub in stead of GCP

.dockerignore
```diff
@@ -1,10 +1,11 @@
 docker-compose.yaml  
 metabase-experiment
-tracker-experiment
+tracker/
+
 backend/jsons/
 backend/venv/
-rod/node_modules/
-rod/build/
```
borft, line 7: this doesn't live in this repo anymore

Makefile
```diff
@@ -25,19 +25,21 @@ push-images: push-image-backend push-image-notebook
 push-image-%:
 	$(eval MODULE = $(subst push-image-,,$@))
 	$(eval URL=$(CONTAINER_REPO)/$(MODULE))
-	docker tag objectiv/$(MODULE):$(TAG) $(URL):latest
-	docker push $(URL)
-	gcloud container images add-tag --quiet $(URL):latest $(URL):$(REVISION)
+	docker tag objectiv/$(MODULE):$(TAG) $(URL):test
```
borft, line 28: change URL to simpler docker.io / /dockerhub

analysis/bach_open_taxonomy/bach_open_taxonomy/sankey_dash.py
```diff
@@ -4,7 +4,7 @@
 import ast
 
 
-def get_app(Dash, feature_frame, url_base_pathname='/'):
+def get_app(Dash, feature_frame, url_base_pathname='/', dash_options=None):
```
borft, line 7: @thijs-obj added optional options ;)

docker-compose.yaml
```diff
@@ -61,13 +72,18 @@ services:
   objectiv_notebook:
     container_name: objectiv_notebook
     # default image: eu.gcr.io/objectiv-production/notebook:latest
-    image: ${OBJECTIV_CONTAINER_URL-eu.gcr.io/objectiv-production}/notebook:${OBJECTIV_CONTAINER_TAG-latest}
+    image: ${OBJECTIV_CONTAINER_URL-objectiv}/notebook:${OBJECTIV_CONTAINER_TAG-latest}
     ports:
-      - 127.0.0.1:8888:8888
+      - "8888:8888"
+      - "8053:8053"
```
thijs-obj, line 78:
```suggestion
      - "127.0.0.1:8888:8888"
      - "127.0.0.1:8053:8053"
```
the token is hardcoded, so we don't want to expose the notebook to the outside world

analysis/bach_open_taxonomy/bach_open_taxonomy/sankey_dash.py
thijs-obj: Good :+1: 

Just mentioning this because I am not sure if you are aware of the possibility, but if you do `**dash_options` here then `dash_options` is automatically a dictionary, but then you'd have to change the call in product_analytics.ipynb to `app = get_app(Dash, feature_frame, server_url='http://localhost:8053')`

analysis/docker/entrypoint.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+# load virtualenv
+source /services/venv/bin/activate
+
+# wait a bit to give PG a chance to start up
+sleep 10
```
thijs-obj, line 7: In the tests of our previous project (objectiv-core) we used to do something like the following to wait for the postgres container to be ready:
```bash
next_wait_time=0
until psql -U $POSTGRES_USER -h $POSTGRES_HOSTNAME $POSTGRES_DB -c "\l" || [ $next_wait_time -eq 6 ]; do
   echo "Connecting to database failed, sleeping and then retrying"
   sleep $(( next_wait_time++ ))
done
```

Might be nice to do that here too in a future PR, as it could improve the boot time

thijs-obj (commented):
looks good to me

PR #293, opened 2021-11-18
Bach: Windowing rework

jansentom:
- use `df.groupby(cols).window()` instead `df.window(cols)`, mainly to make `rolling()` and `expanding()` more pandas-like.
    - Make window function `window` argument Optional
    - Window functions now unset `group_by` on the Series, because they can be evaluated without it.
    - in `agg()`, update the index of the new df by looking at the returned Series.
    - Update some docs
    - Use new windowing syntax in `product_analytics.ipynb` and `series_objectiv.py`

This allows for syntax like
```
# preferred way 
df['col'] = df.groupby('colA').rolling(2).sum()
# or 
rolling = df.groupby('colA').rolling(2)
df['col'] = df.colA.sum(rolling)

# and deprecates
df['col'] = df.rolling('colA', 2).sum()
```

`window()` has been marked as expert in the docs, and we point users to `rolling()` and `expanding()` which should do most tricks.

PR #294, opened 2021-11-19
Bach: More consistent use of groupby index

jansentom:
Slightly change how we deal with `group_by` and `index` and Series and DataFrame.
Made possible by the fact that WindowFunctionExpressions are now their own type, so based on https://github.com/objectiv/objectiv-analytics/pull/293

From the added docs:
```
    # When a Series is used as an index, it should be free from any pending aggregation (and thus `group_by`
    # should be None, and its `Series.index` should be `{}`.
    #
    # `DataFrame.group_by` should always match the `Series.group_by` for all Series in the `data` dict.
    # Series in the `index`, should have `Series.index` == `{}`.
    #
    # The rule here: If a series needs a `group_by` to be evaluated, then and only then it should carry that
    # `group_by`. This implies that index Series coming from a GroupBy, do not carry that `group_by`. Only
    # the data columns do.
```

Main changes:
- Remove group_by from Series that are an index and add consistency checks around that to constructors.
- Add some documentation around how `Series.index`, `Series.group_by` and `DataFrame.group_by` interact.
- Make `DataFrame.__getitem__` more flexible, and improve #docs on how that all works
- Extend tests to deal with new situations.

### DIFF
--- a/bach/bach/dataframe.py
+++ b/bach/bach/dataframe.py
@@ -113,20 +113,21 @@ class DataFrame:
 
     # A DataFrame holds the state of a set of operations on it's base node
     #
-    # The main component in this are the dicts of Series that it keeps: `index` and `data`
+    # The main components in this are the dicts of Series that it keeps: `index` and `data`
     # The `data` Series represent all data columns, possibly waiting for aggregation.
     # The `index` Series are used as an index in key lookups, but serve no other purpose but for nice
     # visualisation when converting to pandas Dataframes.
     #
-    # When a Series is used as an index, it should be free from any pending aggregation (and thus `group_by`
-    # should be None, and its `Series.index` should be `{}`.
+    # When a Series is used as an index, it should be free from any pending aggregation (and thus
+    # `Series.group_by` should be None, and its `Series.index` should be `{}`.
     #
     # `DataFrame.group_by` should always match the `Series.group_by` for all Series in the `data` dict.
-    # Series in the `index`, should have `Series.index` == `{}`.
+    #  (and `Series.index` should match `Series.group_by.index`, but that's checked in `Series.__init__`)
     #
+    # To illustrate (copied verbatim from Series docs):
     # The rule here: If a series needs a `group_by` to be evaluated, then and only then it should carry that
-    # `group_by`. This implies that index Series coming from a GroupBy, do not carry that `group_by`. Only
-    # the data columns do.
+    # `group_by`. This implies that index Series coming from `GroupBy.index`, do not carry that `group_by`.
+    # Only the data Series that actually need the aggregation to happen do.
     #
     # Order is also tracked in `order_by`. It can either be None or a list of SortColumns. Ordering is mostly
     # kept throughout operations, but for example materialization resets the sort order.
--- a/bach/bach/dataframe.py
+++ b/bach/bach/dataframe.py
@@ -110,6 +110,26 @@ class DataFrame:
     # todo note on ordering for slices?
     # todo 'meaning that they originate from the same data source' ok, by approximation
     # todo _get_dtypes also queries the database
+
+    # A DataFrame holds the state of a set of operations on it's base node
+    #
+    # The main component in this is the dists of Series that it keeps: `index` and `data`
+    # The `data` Series represent all data columns, possibly waiting for aggregation.
+    # The `index` Series are used as an index in key lookups, but serve no other purpose but for nice
+    # visualisation when conversion to pandas Dataframes.
+    #
+    # When a Series is used as an index, it should be free from any pending aggregation (and thus `group_by`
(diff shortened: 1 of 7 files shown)


bach/bach/dataframe.py
```diff
@@ -688,37 +717,46 @@ def __getitem__(self,
...
+                if key.expression.has_windowed_aggregate_function:
+                    raise ValueError('Cannot apply a Boolean series containing a window function to '
+                                     'DataFrame. Hint: materialize() the DataFrame before creating the '
+                                     'Boolean series')
+
+                # If the key has no group_by but the df has, this is a filter before aggregation. This is
+                # supported. It can change the aggregated results though, but that makes sense.
+                # A common case here is a filter on the columns in the group_by e.g. the index of this df.
+                # We don't support using aggregated series to filter on a non-aggregated df though:
+                if key.group_by and not self._group_by:
```
jansentom, line 737: This is the only functional change. See changes to [tests](https://github.com/objectiv/objectiv-analytics/pull/294/files#diff-d4bec6523bc2fe35200577a49afd6ec67368b0084c238467864e62de320eecd5) below.

bach/bach/expression.py
```diff
@@ -270,7 +270,7 @@ def is_independent_subquery(self) -> bool:
         # If this Expression is wrapped around a IndependentSubqueryExpression, most likely, there will be
         # just one in here, but let's make sure.
         all_isq = [d.is_independent_subquery for d in self._data if isinstance(d, Expression)]
-        return len(all_isq) and all(all_isq)
+        return len(all_isq) > 0 and all(all_isq)
```
jansentom, line 273: mypy happiness

bach/tests/functional/bach/test_df_getitem.py
```diff
@@ -113,15 +113,57 @@ def test_get_item_materialize():
         ]
     )
 
+def test_get_item_mixed_groupby():
+    bt = get_bt_with_test_data(full_data_set=True)[['municipality', 'inhabitants', 'founding']]
 
-def test_get_item_slice():
```
jansentom, line 117: The test that was removed here is completely covered by `test_positional_slicing` in the same file

bach/tests/unit/bach/test_series.py
```diff
@@ -32,36 +32,47 @@ def test_equals():
 
     int_type = get_series_type_from_dtype('int64')
     float_type = get_series_type_from_dtype('float64')
-    sleft = int_type(engine=None, base_node=None, index={}, name='test', expression='test', group_by=None)
-    sright = int_type(engine=None, base_node=None, index={}, name='test', expression='test', group_by=None)
+
+    expr_test = Expression.construct('test')
```
jansentom, line 36: This was "broken", Series don't accept strings as expressions anymore

bach/tests/functional/bach/test_df_getitem.py
```diff
@@ -113,15 +113,57 @@ def test_get_item_materialize():
...
+        grouped_all_sum[grouped_sum > 50000],
+        order_by='inhabitants_sum',
+        expected_columns=['municipality', '_index_skating_order_sum', 'inhabitants_sum', 'founding_sum'],
+        expected_data=[
+            ['Súdwest-Fryslân', 31, 52965, 7864],
+            ['Leeuwarden', 1, 93485, 1285]
+        ]
+    )
+
+    # Apply a filter on a column of the pre-aggregation df
```
thijs-obj, line 137: Pandas doesn't allow this, and I am not sure that we should. This seems like a feature that is not often needed, but can be confusing when used because the actions on the data (filtering and then aggregating) happen in a different order in the code (we invoke the filter after having done the grouping and summing).

thijs-obj: Maybe we should have a bit clearer error for this case also:
```python
grouped[bt.founding < 1300]
```

thijs-obj (approved):
Love the comments. Don't really see the use-case for the more flexible filtering on aggregated stuff

bach/tests/functional/bach/test_df_getitem.py
jansentom: I agree with your arguments. Let's quickly discuss tomorrow what to do with both these cases. No rush to merge this anyway.

jansentom: > Maybe we should have a bit clearer error for this case also:
> ```python
> grouped[bt.founding < 1300]
> ```
```
  raise ValueError(f'The df has groupby set, but contains Series that have no aggregation '
                   f'function yet. Please make sure to remove these from the frame, or setup'
                   f'aggregation through agg(), or on all individual series, first.'
                   f'Unaggregated series: {not_aggregated}')
```

PR #301, opened 2021-11-22
add meta description to mdx files

borft:
Basically, this looks for the first paragraph (p) in the body (inside <main />), and takes the text from it, limited to 500 chars, and sets that in the head/meta tags for docusaurus to use.

### DIFF
--- a/bach/docs/generate.py
+++ b/bach/docs/generate.py
@@ -285,7 +285,13 @@ for url in urls:
 
     if len(description_elements) > 0:
         description_element = description_elements[0].text
-        description = description_element[:500].replace('\n', ' ')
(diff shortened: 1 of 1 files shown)


jansentom:
Looks okay to me, but shouldn't we want to take complete sentences, or at least complete words?

jansentom (commented):
LGTM, one generic comment.

jansentom (approved):
LGTM, one comment, posted twice ;)

borft:
> Looks okay to me, but shouldn't we want to take complete sentences, or at least complete words?

Fair point. Now using whole words as a quick fix.

PR #311, opened 2021-11-30
updating conversion feature + duration between events

vincenthoogsteder:
- Conversion example now looks at both ways users can go into GH
- Conversion funnel now uses the same df as feature creation, otherwise it contains docs as well and becomes even less readable
- Duration between events metric now looks at contributor features, as the conversion feature did not give a duration


hendrik-obj:
lgtm!

hendrik-obj (approved):

PR #316, opened 2021-12-02
Sqlmodel materialization improvements

thijs-obj:
Changes:
 * add `to_sql_materialized_nodes()` which gives multiple sql statements, one for each materialized node
 * Add support for different kinds of materialization
 * `SqlModel`: optional parameter 'materialization_name` that can be used to specify the name of the view/table to create
* Add option to `find_nodes` to give the longest path for found nodes. We leverage this in the sql generation
 * tests

FYI @jansentom


bach/sql_models/graph_operations.py
```diff
@@ -106,36 +106,56 @@ def get_node(start_node: SqlModel, reference_path: RefPath) -> SqlModel:
...
 
     This function uses a breadth first approach, and the returned FoundNodes are in the order they were
-    found. If a node is encountered multiple times, then only the first occurrence will be in the result.
+    found. If a node is encountered multiple times, then only the first or last occurrence will be in the
+    result depending on the value of use_last_found_instance.
 
     :param start_node: start node
     :param function: Function that should return either True or False for a given SqlModel
-    :return: A list of tuples. Each tuple contains the found SqlModel and a reference path to that model
+    :param use_last_found_instance: If set to true, if a node is encountered multiple times then the
```
jansentom, line 124: would `first_instance` be more logical as a name as it's the default as well. This makes for much easier explanation below as well.

bach/sql_models/model.py
```diff
@@ -499,6 +577,27 @@ def set_materialization(self,
...
         return replace_node_in_graph(self, reference_path, replacement_model)
 
+    def set_materialization_name(self,
+                                 reference_path: RefPath,
+                                 materialization_name: Optional[str]) -> 'SqlModel[T]':
+        """
+        Create a (partial) copy of the graph that can be reached from self, with the materialization_name of
+        the referenced node updated.
+
+        The node identified by the reference_path is copied and updated, as are all nodes that
```
jansentom, line 587: I don't understand why that's necessary. Can't we just update the node only?

bach/sql_models/sql_generator.py
```diff
@@ -14,21 +15,90 @@ def to_sql(model: SqlModel) -> str:
...
 
     # case: len(result) > 1
     sql = 'with '
     sql += ',\n'.join(f'{row.quoted_cte_name} as ({row.sql})' for row in queries[:-1])
     sql += '\n' + queries[-1].sql
-    return sql
+    return _materialize(sql, model)
+
+
+def _materialize(sql_query: str, model: SqlModel) -> str:
```
jansentom, line 78: Nice and simple. Very easy to override in case different DB dialects want this differently.

bach/sql_models/sql_generator.py
```diff
@@ -77,30 +160,41 @@ def _to_cte_sql(compiler_cache: Dict[str, List[SemiCompiledTuple]],
...
+        _single_model_to_sql(compiler_cache=compiler_cache, model=model, reference_names=reference_names)
+    )
 
     compiler_cache[model.hash] = result
     return result
 
 
-def model_to_quoted_cte_name(model):
-    """ Get the name for the cte that will be generated from this model, quoted and escaped. """
+def model_to_name(model: SqlModel):
```
jansentom, line 181: :+1:

bach/tests/unit/sql_models/util.py
```diff
@@ -12,13 +12,15 @@ def assert_roughly_equal_sql(sql_a: str, sql_b: str):
...
+    if a_stripped != b_stripped:
+        print(f'\n\n{a_stripped}\n\n != \n\n{b_stripped}\n\n')
+        raise AssertionError(f'\n\n{a_stripped}\n\n != \n\n{b_stripped}\n\n')
 
 
 class ValueModel(SqlModelBuilder):
     @property
     def sql(self) -> str:
-        return 'select {key} as key, {val} as value'
+        return "select '{key}' as key, {val} as value"
```
jansentom, line 23: would be nice to use the proper escaping methods here. More cross dialect proof

jansentom (commented):
Nice work, and impressive tests. The only general comment I have is that I find the `materialization_name` confusing. That's what we called `savepoint_name` in our discussions, right?

bach/sql_models/model.py
```diff
@@ -331,6 +387,7 @@ def _calculate_hash(self) -> str:
             3. properties
             4. materialization
```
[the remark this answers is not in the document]
thijs-obj: https://github.com/objectiv/objectiv-analytics/pull/317 changes this to how you suggested it

bach/sql_models/model.py
thijs-obj: Immutability was part of the original design. Changed in https://github.com/objectiv/objectiv-analytics/pull/317

bach/sql_models/graph_operations.py
thijs-obj: Changed in e45a8ca

bach/tests/unit/sql_models/util.py
thijs-obj: That is actually not completely trivial here. We'd have to do that by overriding the `properties_to_sql` function, that is quite some work tho. We should probably make that function configurable: either escape all string or don't. Now it never escapes anything

thijs-obj:
> Nice work, and impressive tests. The only general comment I have is that I find the `materialization_name` confusing. That's what we called `savepoint_name` in our discussions, right?

yes. But I figured it makes more sense to call it `materialization_name` here, as it relates directly to the `materialization` property that we already have

jansentom:
> > Nice work, and impressive tests. The only general comment I have is that I find the `materialization_name` confusing. That's what we called `savepoint_name` in our discussions, right?
> 
> yes. But I figured it makes more sense to call it `materialization_name` here, as it relates directly to the `materialization` property that we already have

Okay, cool to keep like this.

PR #317, opened 2021-12-02
WIP sqlmodels: Make materialization mutable

thijs-obj:
This is a WIP concept, that is very close to being reviewable. I just need to do a self-review and some pondering on this, might go to review mostly unchanged.

This is a change that's used by #332, it makes a fundamental change to the graphs/trees that we use in sql-model. Previously graphs (or rather trees) were completely immutable. With this change the edge structure of a graph is still immutable, but the materialization information in each node is mutable.

Obviously this has quite some implications for what we can do with these SqlModel trees:
+ Sharing materialization changes in a graph with all objects that have a references to a (sub) graph (e.g. DataFrames) becomes trivial. Otherwise this would be very challenging
- The inherit complexity of shared mutable data structures. e.g. caching and reusing objects is not trivial anymore
- Some of the things that become more complex in trees specifically: 1) Node hashes are no longer unique. 2) Comparing two trees is not trivial anymore. Even if the hash and materialization is the same, it might be different further down the tree.

# Scenario examples of workflows immutable vs mutable
Scenario's from the end user perspective, assumes the changes (or similar) that are in the follow up #332 

## Mutability
```python
sps = Savepoints()
df = DataFrame.from_table('original_table')
df['municipality'] = df['municipality'].str[:10] 
df2 = df.copy()
df_g = df2[['municipality', 'founding']]\
    .groupby('municipality')\
    .min()\
    .sort_values('founding_min')

sql1 = df_g.view_sql()

# Now set the state of df as savepoint 'start', and materialize that as a table
# This also means that df2 and df_g do not query the 'original_table' table anymore but the 'start' table
df.set_savepoint(sps, 'start', Materialization.TABLE)

# The savepoint changed the sql for df_g
sql2 = df_g.view_sql()
assert sql1 != sql2

#  add a 'grouped' save point, doesn't change anything about any of the DataFrames
df_g.set_savepoint(sps, 'grouped', Materialization.VIEW)

# Now create the table and view
sps.execute()
# Now load the savepointed DataFrame as 'df'. This changes nothing
df = sps.get_df('start')
# load df_g2 from the savepoint. df_g2 will be exactly the same as df_g
df_g2 = sps.get_df('grouped')

# below assert is true, as they do the same query
assert df_g.to_pandas().values.to_list() == df_g2.to_pandas().values.to_list()
# this will always hold. notice: EQUALS
assert df_g == df_g2

sql2 = df_g2.view_sql()
assert sql1 != sql2
```

## Immutability
```python
sps = Savepoints()
df = DataFrame.from_table('original_table')
df['municipality'] = df['municipality'].str[:10] 
df2 = df.copy()
df_g = df2[['municipality', 'founding']]\
    .groupby('municipality')\
    .min()\
    .sort_values('founding_min')

# Now set the state of df as savepoint 'start', and materialize that as a table
# df, df2 and df_g are unchanged, they still query the 'original_table'
df.set_savepoint(sps, 'start', Materialization.TABLE)

# The savepoint dod not change the sql for df_g
sql2 = df_g.view_sql()
assert sql1 == sql2

#  add a 'grouped' save point, doesn't change anything about any of the DataFrames
df_g.set_savepoint(sps, 'grouped', Materialization.VIEW)

# Now create the table and view
sps.execute()
# Now load a different DataFrame as 'df'. NOW it will query the 'start' table and not the 'original_table'
df = sps.get_df('start')
# load df_g2 from the savepoint. df_g2 will query the 'grouped' view, which is based on the 'start' table
df_g2 = sps.get_df('grouped')

# below assert is true as long as the 'original_table' doesn't change. If it does change then that change
# is propagated to df_g as that is still based on the original frame, but not to df_g2 as that queries the
# 'start' table which will not be automatically updated.
assert df_g.to_pandas().values.to_list() == df_g2.to_pandas().values.to_list() 
# this will always hold. notice: NOT equals
assert df_g != df_g2

```

let's discuss this next week @jansentom @hendrik-obj


bach/bach/dataframe.py
```diff
@@ -186,77 +184,6 @@ def __init__(
             raise ValueError(f"The names of the index series and data series should not intersect. "
                              f"Index series: {sorted(index.keys())} data series: {sorted(series.keys())}")
 
-    def copy_override(
```
thijs-obj, line 189: just moved this function to a place a bit lower, so it's grouped more logically with other copy and instantiation functions. No changes to this function

bach/bach/dataframe.py
```diff
@@ -556,20 +483,119 @@ def get_instance(
...
+                    engine=args['engine'], base_node=args['base_node'],
+                    index=args['index'],
+                    name=key, expression=expression_class.column_reference(key),
+                    group_by=args['group_by']
+                )
+                args['series'] = new_series
+
+        return self.__class__(**args, **kwargs)
+
+    def copy_override_base_node(self, base_node: SqlModel) -> 'DataFrame':
```
thijs-obj, line 557: From here on it's new code. The above code is the part that was moved

thijs-obj: This function would be useful without the mutability change too

thijs-obj: resolving this conversation. Earlier in this PR there was a function moved, and this comment indicated where the moved code stopped and the new code started. The code moving is already in main, so that is not visible here anymore.

PR #318, opened 2021-12-06
Bach sample changes

thijs-obj:
Change: Do not insert an additional node when calling `get_sample()`. This ensures that the DataFrame returned by `get_unsampled()` is compatible with the original DF.

To make this possible there is a new requirement to users calling `get_sample()` tho: The dataframe must be in a materialized state.

Other changes:
* Moved some code. Might be easiest to review this code per commit. As all code moving is in 59513106b7547d17d97913fade56695a91c459b2 (=first commit), and the actual changes are in ff4a359cddadca97af591580c529b10471931462 and c18aef8694d271706cc811439af5568748cc99ce
* Add some type annotations

### DIFF
--- a/bach/bach/dataframe.py
+++ b/bach/bach/dataframe.py
@@ -363,7 +363,7 @@ class DataFrame:
             order by ordinal_position;
         """
         with engine.connect() as conn:
-            sql = _escape_parameter_characters(conn, sql)
+            sql = escape_parameter_characters(conn, sql)
             res = conn.execute(sql)
         return {x[0]: get_dtype_from_db_dtype(x[1]) for x in res.fetchall()}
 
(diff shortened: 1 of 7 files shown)


bach/bach/dataframe.py
```diff
@@ -328,14 +328,34 @@ def dtypes(self) -> Dict[str, str]:
         return {column: data.dtype for column, data in self.data.items()}
 
     @property
-    def group_by(self):
+    def group_by(self) -> Optional['GroupBy']:
```
thijs-obj, line 331: Just added this cause I saw it was missing. But this cause some small ripples in other places to satisfy mypy (which was right to complain)

bach/bach/merge.py
```diff
@@ -220,26 +220,7 @@ def merge(
...
-    :param how: supported values: {‘left’, ‘right’, ‘outer’, ‘inner’, ‘cross’}
-    :param on: optional, column(s) to join left and right on.
-    :param left_on: optional, column(s) from the left df to join on
-    :param right_on: optional, column(s) from the right df/series to join on
-    :param left_index: If true uses the index of the left df as columns to join on
-    :param right_index: If true uses the index of the right df/series as columns to join on
-    :param suffixes: Tuple of two strings. Will be used to suffix duplicate column names. Must make column
-        names unique
-    :return: A new Dataframe. The original frames are not modified.
+    See :py:meth:`bach.DataFrame.merge` for more information.
```
thijs-obj, line 223: Unrelated change: Moved the comments to the DataFrame class. As now the comments there say 'see bach.merge.merge', but that's not included in the generated html it seems: https://objectiv.io/docs/modeling/dataframe/bach.DataFrame.merge

bach/bach/sample.py
```diff
@@ -0,0 +1,104 @@
...
+    See :py:meth:`bach.DataFrame.get_sample` for more information.
+    """
+    if sample_percentage is None and filter is None:
+        raise ValueError('Either sample_percentage or filter must be set')
+
+    if not df.is_materialized:
+        raise ValueError("Cannot call get_sample on a non-materialized dataframe. "
+                         "Call materialize() first.")
+
+    original_node = df.base_node
```
thijs-obj, line 32: This is basically the whole change

bach/bach/merge.py
jansentom: good one

bach/bach/sample.py
```diff
@@ -25,7 +25,11 @@ def get_sample(df: DataFrame,
     if sample_percentage is None and filter is None:
         raise ValueError('Either sample_percentage or filter must be set')
 
-    original_node = df.get_current_node(name='before_sampling')
+    if not df.is_materialized:
```
jansentom, line 28: there is a groupby check [below](https://github.com/objectiv/objectiv-analytics/pull/318/commits/ff4a359cddadca97af591580c529b10471931462#R41) that can go because of this.

jansentom (approved):
small comment, rest looks nice!

bach/bach/sample.py
thijs-obj: good point. will clean up

bach/bach/series/series.py
```diff
@@ -862,6 +862,8 @@ def _check_unwrap_groupby(self,
         else:
             if isinstance(wrapped, DataFrame):
                 unwrapped = wrapped.group_by
+                if unwrapped is None:
+                    unwrapped = GroupBy([])
```
thijs-obj, line 866: mypy found this `unwrapper is None` case that was not handled. I guessed from the docstring that this `unwrapped = GroupBy([])` would make sense

bach/tests/unit/bach/test_expression.py
```diff
@@ -144,7 +144,7 @@ def test_is_constant():
     assert not WindowFunctionExpression.construct('agg({}, {}, {})', const1, const2, const3).is_constant
 
 
-def test_is_constant():
+def test_is_single_value():
```
thijs-obj, line 147: unrelated change that I had stashed somewhere. There is already a function called `test_is_constant` above, so this one overwrote that one

jansentom: the name was retarded anyway. Mea culpa

bach/bach/dataframe.py
```diff
@@ -328,14 +328,34 @@ def dtypes(self) -> Dict[str, str]:
...
         return copy(self._group_by)
 
+    @property
+    def is_materialized(self) -> bool:
+        """
+        Return true if this DataFrame is in a materialized state, i.e. all information about the
+        DataFrame's values is encoded in self.base_node.
+
+        A DataFrame that's freshly constructed with :py:meth:`from_table`,
+        :py:meth:`from_model`, or :py:meth:`from_pandas` will be in a materialized state. Operations on such
```
hendrik-obj, line 346: if a table or model contain json type columns it is not in the materialized state as those get cast as jsonb.

hendrik-obj:
good stuff!

bach/bach/dataframe.py
jansentom: hmm, in that case we should delegate the is_materialized check partially to the series. Most of them will be able to use Thijs' implementation, but the json one should have special treatment in that case

thijs-obj: I added comments here and in the docstring of materialize(). And added this as a bug in our tracking document

thijs-obj: I'd rather merge this now tho and fix that in a later branch, so other branches can keep moving.

PR #319, opened 2021-12-07
Remove roadmap references

ivarpruijn:
We're moving to another solution, so let's remove the links to the roadmap in GitHub for now.


jansenbob (approved):
LGTM

PR #332, opened 2021-12-14
WIP: Bach export workflows

thijs-obj:
Ready for a preliminary review.

Most interesting things I think:
* The api we offer to the end user, and whether that integrates well into regular workflows.
  * naming in this api, e.g. `savepoints`, `set_savepoint()`, `execute()`
  * contrast with initial idea document (proposal.md)
* The logic and api for immediately creating a table when needed, and drop/creating later if something changed. Building a table might be very expensive, but not creating it immediately might be an annoying workflow?
  *  `df.set_savepoint()` creates tables directly if the materialization is set to table
  * `savepoints.update_materialization()` doesn't directly create each table. But rather waits for `execute()` to create everything in one batch.
  *  robustness of this is not good yet. But can be improved.

This PR builds on #317, but I am not sure yet I am a fan of the change there.


bach/proposals.md
```diff
@@ -0,0 +1,48 @@
+# Ideas on what workflows to support
```
thijs-obj, line 1: this file is outdated. Will remove later

thijs-obj:
Closing this, as we merged https://github.com/objectiv/objectiv-analytics/pull/347

PR #310, opened 2021-11-30
Bach Open Taxonomy: get started improvements

hendrik-obj:

PR #333, opened 2021-12-15
Fix function name after conflicting merges

thijs-obj:
This fixes a broken import/function name in the bach_open_taxonomy.

This broke because the function name was changed in https://github.com/objectiv/objectiv-analytics/pull/318, while the code using the old function name was added in https://github.com/objectiv/objectiv-analytics/pull/310.

#310 got merged 6 days ago, just after #318 was merged, and since then this was broken on main.


thijs-obj:
Please have a look @jansentom

jansentom (approved):

PR #334, opened 2021-12-15
WIP: Add Coveralls action and README badges

sdirosa:
Low prio: just experimenting with Coveralls here, determine if we want to add this workflow inside of the test one or as a separate one and how to setup the monorepo. 

If I can get it to work with the monorepo I'll update the READMEs with the badges

### DIFF
--- a/.github/workflows/tracker-tests.yml
+++ b/.github/workflows/tracker-tests.yml
@@ -14,6 +14,17 @@ on:
 jobs:
   run-tracker-tests:
     runs-on: ubuntu-latest
+    permissions:
+      actions: none
+      checks: none
+      contents: read
+      deployments: none
+      issues: none
+      packages: none
+      pull-requests: read
+      repository-projects: read
+      security-events: none
+      statuses: read
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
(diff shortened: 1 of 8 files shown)


.github/workflows/tracker-tests.yml
```diff
@@ -7,30 +7,69 @@ on:
...
+      actions: none
+      checks: none
+      contents: read
+      deployments: none
+      issues: none
+      packages: none
+      pull-requests: read
+      repository-projects: read
+      security-events: none
+      statuses: read
```
sdirosa, line 27: I would expect this to fail, because some of the jobs need to write artifacts, but it's unclear if those are regarded as content or not, so just testing it up :)

tracker/Makefile
```diff
@@ -8,3 +8,12 @@ build: clean
...
 	yarn publish:verdaccio
+
+merge-coverage:
+	rm -rf .nyc_output
+	mkdir .nyc_output
+	cp `find core plugins trackers -type f -name coverage-final.json -not -path '*node_modules*'` --backup=numbered .nyc_output
+	nyc merge .nyc_output
+	nyc report --reporter=lcov
+	rm coverage.json
+	rm -rf .nyc_output
```
sdirosa, line 19: Some script to merge coverage reports together, this is actually quite nice for having an overview of all coverage reports across the tracker monorepo. I'll probably pull it out in its own branch.

sdirosa:
Closing this as it doesn't work right out of the box for us due to having a monorepo in a subfolder of another repository in a different language. This service will group all the packages under a single objectiv-analytics overview, which is not what we want.

Next steps:
I plan to look into codcov.io. It looks like it may be able to support our structure (and it's a more complex and mature service overall). I will do that in another PR.

@jansentom Can you revoke the permissions we granted to Coveralls? I'll shoot you another request for Codecov when I've some time to work on this again. It's low prio, so could be a while :)

PR #335, opened 2021-12-16
Move Tracker Transports and QueueStores in their own NPM packages

sdirosa:
All of these Transports and Queues were already used in multiple Trackers and currently duplicated.

The monorepo has two new folders, transports and queues, containing standalone Transport and QueueStore packages. This will make them easy to reuse and maintain, or fork.

No functional changes.


sdirosa:
Re-tested the monorepo bulk commands, no issues. Merging this up.

PR #336, opened 2021-12-16
Bach is_materialized fix

thijs-obj:
Changes:
* Introduce `BachSqlModel`. A subclass of `SqlModel` that has a property `columns` with the names of the columns that the query renders.
* Use `BachSqlModel` everywhere in Bach
* Change to `is_materialized()`:  check whether the current columns in the `DataFrame` match the columns produced by the `base_node`. **<---** This is the reason for the other changes

Alternative solutions could be:
1) Tracking the columns as part of the DataFrame instead of in a SqlModel subclass. This seemed a bit more dirty to me
2) Adding the `columns` to the SqlModel class, instead of adding it in a subclass like in this PR. This might be really nice actually, just started down a different path in this PR, but we could do this in the future
3) Not fixing `df.is_materialized`. This is used for the savepoints, and users might rely on it working. Seems like a bad alternative


bach/tests/functional/bach/test_df_materialize.py
```diff
@@ -149,3 +149,9 @@ def test_is_materialized():
     df['city'] = df['x']
     del df['x']
     assert not df.is_materialized
+
+    # removing columns
+    df = df_original.copy()
+    assert df.is_materialized
+    del df['municipality']
+    assert not df.is_materialized
```
thijs-obj, line 157: This test fails on main. The other changes in this PR fix this

jansentom (approved):
Solid stuff!

hendrik-obj (approved):

analysis/notebooks/open-taxonomy-how-to.ipynb
```diff
@@ -118,11 +118,12 @@
     "* `day`: the day of the session as a date.\n",
     "* `moment`: the exact moment of the event.\n",
     "* `user_id`: the unique identifyer of the user based on the cookie.\n",
-    "* `session_id`: a unique incremented integer id for each session. Starts at 0 for the selected data in the Objectiv Frame.\n",
     "* `global_contexts`: a json-like data column that stores additional information on the event that is logged. This includes data like device data, application data, and cookie information. [See below](#global_contexts) for more detailed explanation. \n",
     "* `location_stack`: a json-like data column that stores information on the exact location where the event is triggered in the product's UI. [See below](#location_stack) for more detailed explanation.\n",
     "* `event_type`: the type of event that is logged.\n",
-    "* `stack_event_types`: the parents of the event_type."
+    "* `stack_event_types`: the parents of the event_type.\n",
```
thijs-obj, line 124: Nice :ok_hand: Thanks @hendrik-obj

PR #339, opened 2021-12-16
Remove automatic tracking for visibility events of regular sections

sdirosa:
As discussed, visibility tracking is a complex topic and we decided to only track those events automatically when we are 100% sure we are doing it correctly. 

This PR switches visibility automatic tracking off exception made for Overlays and Expandable contexts.

### DIFF
--- a/tracker/core/tracker/tests/TrackerTransport.test.ts
+++ b/tracker/core/tracker/tests/TrackerTransport.test.ts
@@ -366,7 +366,6 @@ describe('TrackerTransportRetry', () => {
   });
 
   it('should stop retrying if we reached maxRetryMs', async () => {
-    jest.useRealTimers();
     const slowFailingTransport = {
       transportName: 'SlowFailingTransport',
       handle: async () => new Promise((_, reject) => setTimeout(() => reject(new TransportSendError()), 100)),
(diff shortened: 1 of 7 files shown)


tracker/core/tracker/tests/TrackerTransport.test.ts
```diff
@@ -387,6 +386,8 @@ describe('TrackerTransportRetry', () => {
       expect.arrayContaining([new Error('maxRetryMs reached')])
     );
 
+    jest.runAllTimers();
+
```
sdirosa, line 390: Unrelated fix. Jest documentation suggests to use fake timers whenever possible as they are faster and more reliable than real ones. I've seen this task failing intermittently and I believe it's due to the real times, which were not needed anyway. Let's see if this fixes the issue.

tracker/trackers/browser/src/definitions/LocationContext.ts
```diff
@@ -231,3 +231,10 @@ export type AnyActionContext = Infer<typeof AnyActionContext>;
 export const AnyClickableContext = union([AnyActionContext, ExpandableSectionContext]);
 
 export type AnyClickableContext = Infer<typeof AnyClickableContext>;
+
+/**
+ * Struct union to match any Showable Context, that is Overlays and ExpandableSectionContext
+ */
+export const AnyShowableContext = union([OverlayContext, ExpandableSectionContext]);
+
+export type AnyShowableContext = Infer<typeof AnyShowableContext>;
```
sdirosa, line 240: ExpandableSectionContext may go from here, but we need to have a discussion about it first. It's a complex element that may be interpreted subjectively.

PR #341, opened 2021-12-16
Speed up Tracker test workflow

sdirosa:
Just an experiment stemmed from stumbling upon this: https://github.com/actions/setup-node#caching-packages-dependencies

As we add more packages the current workflow is becoming a bit too slow. I've added a few optimizations in here to reduce redundant checks and speed up bulk commands we run on every push and pull_request.

Some changes also improve the speed of scripts runtimes in dev mode. Like test, prettify, tsc, etc.


sdirosa:
This shaves off about 1 minute from the overall execution

.github/workflows/tracker-tests.yml
```diff
@@ -17,20 +17,19 @@ jobs:
...
           check-latest: true
-      - name: Install monorepo dependencies
-        run: |
-          cd tracker
-          yarn install
+          cache: 'yarn'
+          cache-dependency-path: tracker/yarn.lock
+      - name: Install dependencies
+        working-directory: tracker
+        run: yarn install
```
sdirosa, line 24: This speeds up the monorepo root, but to cache all of the packages we need to add a job for each. It's a bit of work, so perhaps for another time. The improvements coming from the optimizations already present in this branch are pretty decent anyway.

tracker/CONTRIBUTING.md
```diff
@@ -155,6 +155,9 @@ Runs prettier for `core/schema/src/*`, `core/tracker/src/ContextFactories.ts` an
 ### `yarn tsc`
 Runs the TypeScript compiler for all typed packages.
 
+### `yarn tsc:generated`
+Runs the TypeScript compiler for `core/schema` and `core/tracker`.
+
```
sdirosa, line 160: Added this as we generate only schema and core factories, there's no need of running TSC for all other packages in the monorepo as nothing is yet generated for those.

sdirosa:
Merging this. There are other improvements to be made, but it's already quite a nice improvement, especially in dev where we can take advantage of parallel runs

PR #342, opened 2021-12-16
Bach sample improvements

thijs-obj:
Changes:
* Don't require the `DataFrame` to be materialized when calling `get_sample()`
* Don't materialize in `get_unsampled()`. Previously we did that if `df._group_by` was set
* Also support sampling grouped DataFrames
* The `is_materialized` function is not used anymore now. But I do require it in another PR, so haven't removed it here.

I think somebody (hendrik?) asked me last time when I implemented #318 whether we couldn't skip the materialization requirement. I think I claimed then that that was not possible, no idea why :man_shrugging: I think this should work. Ran into this while working on the is_materialized stuff

### DIFF
--- a/bach/bach/dataframe.py
+++ b/bach/bach/dataframe.py
@@ -554,6 +554,28 @@ class DataFrame:
 
         return self.__class__(**args, **kwargs)
 
+    def copy_override_base_node(self, base_node: SqlModel) -> 'DataFrame':
+        """
+        INTERNAL
+
+        Create a copy of self, with the base_node overridden in both the returned DataFrame and the Series
+        that are part of that DataFrame. If self.group_by is not None, then it's base_node is updated as
+        well.
+        This is different from :py:meth:`copy_override()`, which when provided with a new base_node only
+        overrides the base_node of the DataFrame and not of the Series that make up the DataFrame nor of
+        the GroupBy.
+        """
+        index = {name: series.copy_override(base_node=base_node) for name, series in self.index.items()}
+
+        group_by = self.group_by
+        if group_by is not None:
+            group_by = group_by.copy_override_base_node(base_node=base_node)
+
+        series = {name: series.copy_override(base_node=base_node, group_by=[group_by], index=index)
+                  for name, series in self.data.items()}
+
+        return self.copy_override(base_node=base_node, index=index, series=series, group_by=[group_by])
+
     def copy(self):
         """
         Return a copy of this DataFrame.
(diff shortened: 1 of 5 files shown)


bach/bach/sample.py
```diff
@@ -49,7 +46,7 @@ def get_sample(df: DataFrame,
 
             sql = f'''
                 create temporary table tmp_table_name on commit drop as
-                ({df.view_sql()});
+                ({to_sql(df.base_node)});
```
thijs-obj, line 49: main change 1/2

bach/bach/sample.py
```diff
@@ -62,23 +59,13 @@ def get_sample(df: DataFrame,
...
     new_base_node = SampleSqlModel(table_name=table_name, previous=original_node)
-
-    return DataFrame.get_instance(
-        engine=df.engine,
-        base_node=new_base_node,
-        index_dtypes=df.index_dtypes,
-        dtypes=df.dtypes,
-        group_by=None
-    )
+    return df.copy_override_base_node(base_node=new_base_node)
```
thijs-obj, line 62: main change 2/2

bach/bach/dataframe.py
```diff
@@ -554,6 +554,28 @@ def copy_override(
 
         return self.__class__(**args, **kwargs)
 
+    def copy_override_base_node(self, base_node: SqlModel) -> 'DataFrame':
```
jansentom, line 557: should this logic not be in `copy_override`? supporting base_node overrides there makes no sense

bach/bach/partitioning.py
```diff
@@ -58,9 +61,12 @@ def __init__(self, group_by_columns: List[Series]):
         for col in group_by_columns:
             if not isinstance(col, Series):
                 raise ValueError(f'Unsupported argument type: {type(col)}')
-            if col.expression.is_constant and not isinstance(col, SeriesInt64):
-                # Dictated by PG
-                raise ValueError('Non-integer constant group by series not supported')
+            if col.expression.is_constant:
```
jansentom, line 64: smart

jansentom (approved):

bach/bach/dataframe.py
thijs-obj: Maybe. Didn't go that route as `copy_override` is already quite a complicated function. And it has a lot of parameters that would interact with this logic. e.g. what to do if both `base_node` and `group_by` are set, or `index`, or `series`, or `index_dyptes` or `series_dtyptes`

PR #347, opened 2021-12-21
Bach export workflow alternative

thijs-obj:
For final review

Based on discussion with Hendrik: https://docs.google.com/document/d/1Zi4LPNz2vpRC3OnWyBk5YbdKVg22dcrXM4LQetxZSuw/edit (private link, not externally visible)

Some notable points:
* We track savepoints in DataFrame
* When merging two dataframes we merge the savepoints objects
* We don't track savepoints in indivdual series

We really need to think on whether we want to track all this state in the DataFrame by default. But let's think about that very well later


bach/tests/functional/bach/test_df_savepoints.py
```diff
@@ -0,0 +1,43 @@
...
+Copyright 2021 Objectiv B.V.
+"""
+import pytest
+
+from sql_models.model import Materialization
+from tests.functional.bach.test_data_and_utils import get_bt_with_test_data, assert_equals_data
+from tests.functional.bach.test_savepoints import remove_created_db_objects
+
+
+def test_savepoint_materialization():
```
thijs-obj, line 11: This is the main usage example: Operations on the savepoints can be done through `df.savepoints`, with a shortcut for saving the current state as a savepoint with `df.set_savepoint()`

bach/tests/functional/bach/test_df_savepoints.py
```diff
@@ -0,0 +1,43 @@
...
+    df = df[['city', 'municipality']]
+    df.set_savepoint('savepoint3', Materialization.VIEW)
+
+    expected_columns = ['_index_skating_order', 'city', 'municipality']
+    expected_data = [[2, 'Snits', 'Súdwest-Fryslân']]
+
+    assert_equals_data(df, expected_columns=expected_columns, expected_data=expected_data)
+
+    df_mat = df.savepoints.get_materialized_df('savepoint3', engine)
+    with pytest.raises(Exception, match='relation "savepoint3" does not exist'):
```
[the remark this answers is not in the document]
thijs-obj: True, not sure if there is a simple way around that tho

bach/bach/savepoints.py
```diff
@@ -0,0 +1,289 @@
...
+            raise ValueError(f'Another savepoint has the same base_node, not allowed. Other savepont: '
+                             f'{existing_basenodes[df.base_node]}')
+
+        self._entries[name] = SavepointEntry(
+            name=name,
+            df_original=df.copy(),
+            materialization=materialization
+        )
+
+    def set_materialization(self, name: str, materialization: Union[Materialization, str]):
```
thijs-obj, line 104: renamed this function. This can be used to set the materialization. By default the materialization when using `DataFrame.set_savepoint` is CTE, i.e. nothing is materialized. So the user needs to explicitly set that.

thijs-obj: Change based on comments in private chat

bach/tests/functional/bach/test_df_savepoints.py
```diff
@@ -0,0 +1,43 @@
...
+        # The df_mat DataFrame queries directly from the 'savepoint3' view in the database.
+        # This will fail because that view has not been created yet.
+        df_mat.to_pandas()
+
+    # Create tables and views, then try again
+    created_result = df.savepoints.write_to_db(engine)
```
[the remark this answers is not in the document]
thijs-obj: It now uses the df's engine by default

PR #348, opened 2021-12-22
Fixbach  docs case

borft:
This PR does the following:
- rename the subdirs from dataframe/series/objectiv to DataFrame/Series/Objectiv, to reflect the .rst's they originate from

This means we remove the dataframe <->DataFrame issues, that break Windows builds, and make for confusing URL's

### DIFF
--- a/bach/docs/source/Objectiv.rst
+++ b/bach/docs/source/Objectiv.rst
@@ -16,4 +16,4 @@ Reference
     ObjectivStack
     SeriesGlobalContexts
     SeriesLocationStack
-    FeatureFrame
+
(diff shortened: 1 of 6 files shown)


bach/docs/source/Objectiv.rst
```diff
@@ -11,9 +11,9 @@ Reference
 ---------
 .. autosummary::
     :template: autosummary/class_short.rst
-    :toctree: objectiv
+    :toctree: Objectiv
 
     ObjectivStack
     SeriesGlobalContexts
     SeriesLocationStack
-    FeatureFrame
```
borft, line 19: @jansentom this was renamed to `ObjectivFrame`, haven't added that yet, as there are some errors/warnings in the docstrings there. We could consider picking that up here as well.

jansentom (approved):
I'm okay with this. As discussed: please check notebook links to docs to make sure they're not broken, as well as deeplinks from www to docs, thnx.

PR #353, opened 2021-12-23
Sqlmodels: small bug fix

thijs-obj:


bach/tests/unit/sql_models/test_graph_operations.py
```diff
@@ -173,6 +173,18 @@ def test_find_nodes():
     ]
 
 
+def test_find_nodes_duplicates():
```
thijs-obj, line 176: This test failed without the changes

jansentom (approved):
Impressive find. took me 5 reads to understand what was happening

PR #356, opened 2021-12-28
Variable support

thijs-obj:
# End User Changes:
Variable support. Variables are created as Series which can be used similar to constants, except that their value can later be changed. Variables are identified by a combination of name and dtype.

Specifically the following new functions are available:
  * `DataFrame.create_variable()`: adds the variable and its initial value to `self.variables` and returns a `Series` that can be used as a placeholder whose value can later be changed. Wherever the returned `Series` object is used, the value of the variable is set when the state of the DataFrame is translated to a sql query
  * `DataFrame.set_variable()`. Set a value in `self.variables`
  * `DataFrame.get_all_variable_usage()`: Give an overview of all variables that are used. This can help with diagnosing variable problems.

# Code Changes:
* Supporting change: Introduce two `BachSqlModel` subclasses: `CurrentNodeSqlModel` and `MergeSqlModel`.
  * The `__init__` of these classes generate the raw sql.
  * Models have no properties that are `Expressions` or `Lists[Expression]`. All Expressions are translated to sql in the Model `__init__()`
    *  `Expression.to_sql()` now assumes it will become raw Model.sql and will have format() called on in twice
  *  Got rid of `BachSqlModelBuilder`
* Add ExpressionToken `VariableToken` that can be used as a placeholder for values
* `DataFrame` has a mapping `variables` with values for variables
* `df.variables` are used and set as properties when creating instances of `CurrentNodeSqlModel` and `MergeSqlModel`.
* `df.variables` are used to replace existing properties in the whole tree of `self.base_node`

PR #357, opened 2021-12-28
Sqlmodels get and update properties in the graph

thijs-obj:
Adds two operations to sqlmodels:
1. `get_all_properties()`: Get all properties in the graph
2. `update_properties_in_graph`: update all occurrences of properties in the graph

This will be used for adding variables to bach. See https://github.com/objectiv/objectiv-analytics/pull/356 (still WIP at this time)


bach/sql_models/graph_operations.py
```diff
@@ -194,6 +194,57 @@ def replace_node_in_graph(start_node: SqlModel,
...
+        node_keys = set(node.properties.keys())
+        return bool(find_keys.intersection(node_keys))
+
+    found_nodes = find_nodes(start_node, function=filter)
+    for found_node in found_nodes:
+        node_keys = set(found_node.model.properties.keys())
+        matching_keys = find_keys.intersection(node_keys)
+        dict_to_update = {
+            key: property_values[key]
+            for key in matching_keys if found_node.model.properties[key] != property_values[key]
```
jansentom, line 240: Is this compare necessary or is it just nice? I don't see any side-effects when we just overwrite or am I missing something?

jansentom (approved):
LGTM, one minor question

bach/sql_models/graph_operations.py
thijs-obj: aah good one, this was not needed.

I changed the code a little bit now, so that we do use this filtering. If `dict_to_update` is empty we don't update anything. See 3203c72633ba1d6ebf797ffd62857040722952c4

thijs-obj:
merging as there was a LGTM. changes after that were minor

PR #358, opened 2022-01-03
WIP: Tracker copyright year update

sdirosa:
This updates the copyright everywhere in the Tracker monorepo. 

Waiting for a confirmation on the format to use. Year vs PreviousYear-Year.

### DIFF
--- a/.github/workflows/tracker-tests.yml
+++ b/.github/workflows/tracker-tests.yml
@@ -15,7 +15,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
         with:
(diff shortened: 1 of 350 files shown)


.github/workflows/tracker-tests.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
         with:
-          node-version: '14'
+          node-version: '16'
```
sdirosa, line 18: Unrelated update, node 16 is faster and it's what I actually use as daily on dev

PR #360, opened 2022-01-03
Finalize deprecation of old taxonomy

sdirosa:
This Pr finalizes the latest version of the monorepo packages to 0.0.5. Version 0.0.4 was a mix of legacy packages and new ones. While version 0.0.5 is our new clean slate. This includes also internal packages that were left behind until now.


sdirosa:
Published on npm, merging up.

PR #361, opened 2022-01-05
Tracker React: finalize bundling

sdirosa:
A few changes to finalize the bundler for React modules. Some fixes and some improvements.

This also bumps versions to v0.0.6.

Next PRs:
- Refactor TrackerQueue to immediately send events and batch them only when sent in bulk, as opposed to a timer approach as it is now
- Split out and finalize the LocationTree module into a proper validation module


tracker/core/react/src/common/LocationTree.ts
```diff
@@ -6,6 +6,10 @@ import { AbstractLocationContext } from '@objectiv/schema';
 import { getLocationPath } from '@objectiv/tracker-core';
 import { LocationContext } from '../types';
 
+/**
+ * TODO: this module is highly experimental and currently not enabled
+ */
+
```
sdirosa, line 12: I had planned to test this on staging with a number of cases, but we decided to move faster and I don't think this module is production read.

We also want to split this out in its own package/service as it would be nice to be able to validate individual Events or sessions without having to install the Tracker packages or the Collector. 

I will finalize it when we get there.

tracker/core/react/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
-    "jsx": "react-jsx"
+    "jsx": "react"
```
sdirosa, line 4: I've decided to switch away from the new jsx transforms as they were introducing some unexpected errors in older versions of React and I'd like, for the time being, to keep compatibility a bit higher. This forced our users to hgave React 17 and up.

tracker/package.json
```diff
@@ -12,11 +12,11 @@
   ],
   "scripts": {
     "clear": "yarn exec make clean && yarn test --clearCache && yarn cache clean",
-    "build-core": "yarn workspaces foreach -p --include @objectiv/tracker-core --verbose run build",
+    "build-core": "yarn workspaces foreach --include @objectiv/tracker-core --include @objectiv/tracker-core-react --verbose run build",
```
sdirosa, line 15: Here we build the tracker-core-react package as well, not in parallel because they cores depend on each other.

tracker/package.json
```diff
@@ -12,11 +12,11 @@
...
     "build-plugins": "yarn workspaces foreach -p --include '@objectiv/plugin**' --verbose run build",
     "build-queues": "yarn workspaces foreach -p --include '@objectiv/queue**' --verbose run build",
     "build-transports": "yarn workspaces foreach -p --include '@objectiv/transport**' --verbose run build",
-    "build-trackers": "yarn workspace @objectiv/tracker-browser build && yarn workspace @objectiv/tracker-angular build",
+    "build-trackers": "yarn workspace @objectiv/tracker-react build && yarn workspace @objectiv/tracker-browser build && yarn workspace @objectiv/tracker-angular build",
```
sdirosa, line 19: Build the react tracker web package as well

tracker/trackers/react/rollup.config.js
```diff
@@ -19,6 +20,6 @@ export default [
         sourcemap: true,
       },
     ],
-    plugins: [nodeResolve(), commonjs(), ts(), terser(), sizes(), filesize()],
+    plugins: [nodeResolve({ browser: true }), peerDepsExternal(), commonjs(), ts(), terser(), sizes(), filesize()],
```
sdirosa, line 23: We need to add React as external peer dependency, otherwise Rollup will automatically embed React 17 in the build and that will cause mayhem.

tracker/core/react/src/locationWrappers/LocationContextWrapper.tsx
```diff
@@ -33,8 +31,8 @@ export type LocationContextWrapperProps = {
 export const LocationContextWrapper = ({ children, locationContext }: LocationContextWrapperProps) => {
   const tracker = useTracker();
 
-  // Add new LocationEntry to LocationTree as well
-  LocationTree.add(locationContext, useParentLocationContext());
+  // TODO: Add new LocationEntry to LocationTree as well (LocationTree is not ready for production yet)
+  //LocationTree.add(locationContext, useParentLocationContext());
```
sdirosa, line 35: Here I've disabled the LocationTree module, until it's ready to be used

sdirosa:
Merging this, tested on Website, as I move onto the TrackerQueue optimizations and queue waiting tweaks for external links

PR #363, opened 2022-01-05
TrackerQueue runner mechanic refactoring + Bugfixes

sdirosa:
In this PR:

- BUGFIX: Got rid of `es6-error` dependency in Core Tracker; turns out [it doesn't work after uglification ](https://github.com/bjyoungblood/es6-error/issues/31#issuecomment-301128220). This fixes our Retry mechanism.
- Removed `startRunner` from TrackerQueue interface. 
- TrackerQueue `push` executes `run` right after adding the received Events to the store. This makes the Queue eager and will send events immediately if idle. As opposed to be timed.
- TrackerQueue `run` recursively invokes itself until all work is done.
- TrackerQueue `run` will keep track of its last execution and delay the next run by `batchDelayMs`. No more `setInterval`. This fixes also performance and memory issues on slower machines and a warning from Google Chrome related to setInterval being too slow on heavy payloads.
- BUGFIX: TrackerQueue `run` will always return a valued Promise. There was a case returning `undefined`.


sdirosa:
These changes have been tested in: https://github.com/objectiv/objectiv.io/pull/216

PR #365, opened 2022-01-10
Fix arm64 docker images

borft:
We use our own metabase container, as the ones provided by metabase only support amd64. 
See https://github.com/objectiv/objectiv-demo/pull/45/files for the `Dockerfile`.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -103,7 +103,7 @@ services:
 
   objectiv_metabase:
     container_name: objectiv_metabase
-    image: objectiv/metabase
+    image: ${OBJECTIV_CONTAINER_URL-objectiv}/metabase:${OBJECTIV_CONTAINER_TAG-latest}
     ports:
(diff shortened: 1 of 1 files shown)


thijs-obj:
This change seems technical OK to me. But I do have a question about this in combination with https://github.com/objectiv/objectiv-demo/pull/45/files.

MetaBase is published under the AGPL. Do we need to explicitly make our own Dockerfile AGPL licensed, or maybe even that whole repo? Have we looked into this?
I think we can proceed and merge this PR, as I don't think we'd mind making that Dockerfile AGPL if that is required. But we should double check whether we are currently in line with the license.

CC: @ivarpruijn

borft:
Good point. Alternatively, we could build the image, without the embedded metabase `.jar`, and download it at runtime (260MiB). That way we are not bundeling it in our image. Downside, is slower start-up, dependency on being online when starting the demo, and also relying on the download working at start-up time.

PR #369, opened 2022-01-10
level support in reset_index

KathiaBarahona:


bach/tests/functional/bach/test_df_reset_index.py
```diff
@@ -34,6 +34,24 @@ def test_reset_index_to_empty():
...
+    # level + drop
+    lbt = bt_cp.reset_index(level='city', drop=True)
+    assert 'city' not in lbt.data
+    assert list(lbt.index.keys()) == final_index
+
+    # dropping invalid level
+    invalid_level = 'random'
+    with pytest.raises(ValueError, match=fr"'{invalid_level}' level not found"):
+        bt_cp.reset_index(level=['city', invalid_level])
+
```
thijs-obj, line 54: Can I suggest one more test case:
``` python
def test_reset_index_no_change():
    bt = get_bt_with_test_data()
    bt = bt.set_index(['skating_order', 'city'], append=True)
    lbt = bt.reset_index(level=[])
    assert list(lbt.index.keys()) == list(bt.index.keys())
```

This doesn't currently work, but I think we should make it pass. It works with Pandas

bach/bach/dataframe.py
```diff
@@ -978,13 +978,19 @@ def rename(self, mapper: Union[Dict[str, str], Callable[[str], str]] = None,
         df._data = new_data
         return None if inplace else df
 
-    def reset_index(self, drop: bool = False, inplace: bool = False) -> Optional['DataFrame']:
+    def reset_index(
+        self,
+        level: Optional[Union[str, Sequence[str]]] = None,
```
thijs-obj, line 983: I think this is good.

@hendrik-obj do you agree, or do we need to support integers here for the columns too?

thijs-obj (changes requested):
One small change needed I think

bach/tests/functional/bach/test_df_reset_index.py
KathiaBarahona: thanks, nice catch!

thijs-obj (commented):
Looks good to me. But let's wait for @hendrik-obj to reply too

hendrik-obj (approved):
lgtm!

thijs-obj (approved):

PR #370, opened 2022-01-10
do series operarations on different base nodes and group bys

hendrik-obj:
With this branch it is always possible to set series to dataframes if the index is of the same dtype (only for single level index). Index name can be different. 
Base nodes can be different.
Group by can be different.
Similarly operations on series are now possible using the index.


bach/docs/source/DataFrame.rst
```diff
@@ -35,9 +35,12 @@ which the condition is true.
 Moving Series around
 ~~~~~~~~~~~~~~~~~~~~
 Values, Series or DataFrames can be set to another DataFrame. Setting Series or DataFrames to another
-DataFrame is possible if they share the same base node. This means that they originate from the same data
-source. In most cases this means that the series that is to be set to the DataFrame is a result of
-operations on the DataFrame that is started with.
+DataFrame is possible if they share the same base node or index dtype. DataFrames and Series share the
+same base node if that they originate from the same data source. In most cases this means that the series
```
thijs-obj, line 39:
```suggestion
same base node if they originate from the same data source. In most cases this means that the series
```

thijs-obj (approved):
A lot of tiny comments, nothing major

bach/bach/series/series.py
```diff
@@ -355,21 +355,26 @@ def get_column_expression(self, table_alias: str = None) -> Expression:
...
             if self.base_node != other.base_node or self.group_by != other.group_by:
                 if other.expression.is_single_value:
                     other = self.as_independent_subquery(other)
                 else:
-                    # TODO improve error
-                    raise ValueError("rhs has a different base_node or group_by, but contains more than one "
-                                     "value. This is not supported.")
+                    # align base nodes and group by
+                    df = self.to_frame()
+                    df._index_merge(key='__other', value=other, how='outer')
```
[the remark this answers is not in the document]
hendrik-obj: added comments about this method (also in DataFrame)

thijs-obj (approved):

bach/tests/functional/bach/test_df_setitem.py
```diff
@@ -233,6 +234,38 @@ def test_set_different_base_node():
     )
 
 
+def test_set_different_group_by():
+    bt = get_bt_with_test_data(full_data_set=True)
+    mt = get_bt_with_railway_data()
+    bt_g = bt.groupby('city')[['inhabitants','founding']]
```
KathiaBarahona, line 240:
```suggestion
    bt_g = bt.groupby('city')[['inhabitants', 'founding']]
```

bach/bach/dataframe.py
```diff
@@ -908,6 +882,61 @@ def __setitem__(self,
         else:
             raise ValueError(f'Key should be either a string or a list of strings, value: {key}')
 
+    def _index_merge(self,
+                     key: str,
+                     value: 'Series',
+                     how: str = 'left'):
```
KathiaBarahona, line 888: **question**: When do you expect `how == 'outer'`?

bach/bach/dataframe.py
```diff
@@ -908,6 +882,61 @@ def __setitem__(self,
...
+                        right_index=True,
+                        how=how,
+                        suffixes=('', '__remove'))
+
+        if how == 'outer':
+            new_index = df.index[index_name].fillna(df[value_index_name + '__index'])
+        elif how == 'left':
+            new_index = df.index[index_name]
+        else:
+            raise NotImplementedError(f'how "{how}" not supported')
```
KathiaBarahona, line 924: I think it's not necessary to raise `NotImplementedError`. This should be handled by merge operation.
```suggestion
         new_index = df.index[index_name]
        if how == 'outer':
            new_index = new_index.fillna(df[value_index_name + '__index'])
```

bach/bach/dataframe.py
```diff
@@ -908,6 +882,61 @@ def __setitem__(self,
...
+                             ' index')
+        index_name = self.index_columns[0]
+        value_index_name = list(value.index.keys())[0]
+        if self.index[index_name].dtype != value.index[value_index_name].dtype:
+            raise ValueError('dtypes of indexes should be the same')
+        renamed_value = value.copy_override(name=key)
+
+        if how == 'outer':
+            renamed_value = renamed_value.to_frame().materialize()
+            renamed_value[value_index_name + '__index'] = renamed_value.index[value_index_name]
```
KathiaBarahona, line 911: *question*: Is this a hack for the outer join bug?

bach/bach/dataframe.py
KathiaBarahona: nvm, already saw it's used by series.

bach/tests/functional/bach/test_df_setitem.py
```diff
@@ -233,6 +234,38 @@ def test_set_different_base_node():
...
+    mt_g = mt.groupby('town').station_id.count()
+
+    with pytest.raises(ValueError, match="Setting new columns to grouped DataFrame is only supported if the"
+                                         " DataFrame has aggregated columns"):
```
KathiaBarahona, line 244: **tip:** You can set the match as a regex if you don't want to copy the whole text :)  [source](https://docs.pytest.org/en/latest/how-to/assert.html#:~:text=with%20pytest.raises(valueerror%2C%20match%3Dr%22.*%20123%20.*%22)%3A)
```suggestion
    with pytest.raises(ValueError, match=r"Setting new columns to grouped DataFrame is only supported if the"):
```

hendrik-obj: Yes, I did it on purpose actually to match the entire error message. Bit of overdoing it, but just to be sure...

bach/bach/dataframe.py
hendrik-obj: this function is specifically only to be used with outer and left, and only for `__setitem__` and `_get_supported` from series.py. I didn't test other situations. Now if you would do 'right' for example with you suggested change, it would give only one index from the 'left' df (and the data from the index from the right would be missing). 
Therefore suggest to leave it as it is, but also move the functionality to merge asap, so this function can be removed.

bach/bach/dataframe.py
hendrik-obj: yes, and it does some renaming etc for left merge. so behaviour is slightly different from merge for left merge on indexes as well.

bach/tests/functional/bach/test_df_setitem.py
hendrik-obj: ah yes, the tests don't get tested for this..

KathiaBarahona (approved):
lgtm

PR #372, opened 2022-01-12
fix order of applied series

KathiaBarahona:


bach/bach/dataframe.py
```diff
@@ -1630,12 +1630,15 @@ def merge(
...
-                              axis: int = 1,
-                              numeric_only: bool = False,
-                              exclude_non_applied: bool = False,
-                              *args, **kwargs) -> List['Series']:
+    def _apply_func_to_series(
+        self,
+        func: Union[ColumnFunction, Dict[str, ColumnFunction]],
+        axis: int = 1,
+        numeric_only: bool = False,
+        exclude_non_applied: bool = False,
```
thijs-obj, line 1638: I think this function is only ever called from `DataFrame.agg()` and that sets this `exclude_non_applied` to `True`. So I think we could just delete this parameter, and remove the special handling of the `non_applied_series` below.

But @jansentom is more knowledgeable about this than me, so his opinion might differ

thijs-obj (commented):
One comment, otherwise looks good with a solid test.

bach/bach/dataframe.py
jansentom: Let's leave it for now. There is some pandas `apply()` functionality that we probably want to support sometime that needs this, and this implementation is quite clean now. No need to remove it imho.

KathiaBarahona: alrighty

PR #373, opened 2022-01-12
Update README.md to align with docs & site

jansenbob:
**Note: This should only be merged when objectiv-stack.svg has been uploaded to production in the Objectiv Docs.**


ivarpruijn (approved):
LGTM, the image is also on production now

PR #374, opened 2022-01-12
df describer

KathiaBarahona:
Implemented `describe` in DataFrames. Will open another PR for `series.describe()`.

Will create other issues to support:
- top for categorical columns.
- percentiles [done]
- support of datetime_is_numeric  <- we cannot convert datetime series to numeric. Only supports max and min agg.


bach/bach/describe.py
```diff
@@ -0,0 +1,162 @@
...
+    )
+
+
+_SUPPORTED_STATS_X_STMT = {
+    SupportedStats.COUNT: 'count({col})',
+    SupportedStats.MEAN: 'cast(avg({col}) as double precision)',
+    SupportedStats.STD: 'cast(stddev({col}) as double precision)',
+    SupportedStats.MIN: 'min({col})',
+    SupportedStats.MAX: 'max({col})',
+    SupportedStats.UNIQUE: 'count( distinct {col})',
```
KathiaBarahona, line 47: @thijs-obj Not sure if we already have smth similar to this somewhere.

bach/bach/dataframe.py
```diff
@@ -1911,6 +1911,23 @@ def nunique(self, axis=1, skipna=True, **kwargs):
...
+        """
+        from bach.series import SeriesAbstractNumeric
+
+        df = self
+        for col in df.data.values():
+            if not isinstance(col, SeriesAbstractNumeric):
+                continue
+            df._data[col.name] = col.round(decimals)
+
+        return df
```
KathiaBarahona, line 1929: should had opened another PR for this, but kinda noticed I needed it when I was finishing

bach/bach/series/series.py
```diff
@@ -919,8 +922,12 @@ def _derived_agg_func(self,
                              f'`{self.name}` Try calling materialize() on the DataFrame'
                              f' this Series belongs to first.')
 
+        derived_dtype = self.dtype if dtype is None else dtype
         if isinstance(expression, str):
-            expression = AggregateFunctionExpression.construct(f'{expression}({{}})', self)
+            base_expression = f'{expression}({{}})'
```
jansentom, line 927: very clean, nice!

bach/bach/describe.py
```diff
@@ -0,0 +1,183 @@
...
+    SupportedStats.MIN: DataFrame.min,
+    SupportedStats.MAX: DataFrame.max,
+    SupportedStats.UNIQUE: DataFrame.nunique,
+}
+
+FilteringDType = Union[Optional[Union[str, Sequence[str]]], List[Type]]
+
+
+@dataclass
+class DataFrameDescriber:
```
jansentom, line 62: I did not review this class in detail yet, leaving that to @thijs-obj for Monday ;)

jansentom (commented):
Impressive that you managed to get this working in your first week! 
A few general remarks:
- Very nice approach, and most of it quite easy to review.
- You mix  "column" or "columns" with "series" to name the individual data series. We use series to stick to Pandas jargon, although columns is not wrong, it's nice to have some consistency there. 
- I did not fully finish the review, leaving some for @thijs-obj on Monday ;)

bach/bach/describe.py
```diff
@@ -0,0 +1,183 @@
...
+        if dtypes_series[SupportedStats.CATEGORICAL]:
+            return SupportedStats.CATEGORICAL
+
+        raise ValueError('DataFrame or Series has no supported dtype to describe.')
+
+    def _get_stats(self, columns_to_describe: List[str]) -> DataFrame:
+        final_df: DataFrame
+        for pos, stat in enumerate(self.main_stat.value):
+            stat_df = self.df.copy_override()[columns_to_describe]
+            stat_df = stat_df.to_frame() if isinstance(stat_df, Series) else stat_df
```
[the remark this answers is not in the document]
KathiaBarahona: not really, mypy was complaining :)

bach/bach/dataframe.py
```diff
@@ -1911,6 +1911,23 @@ def nunique(self, axis=1, skipna=True, **kwargs):
...
+        :param decimals: number of decimal places to round each numeric column
+        :returns: a new DataFrame with rounded numerical columns
+        """
+        from bach.series import SeriesAbstractNumeric
+
+        df = self
+        for col in df.data.values():
+            if not isinstance(col, SeriesAbstractNumeric):
+                continue
+            df._data[col.name] = col.round(decimals)
```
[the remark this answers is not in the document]
KathiaBarahona: I tried doing `_apply_func_to_series` but got a bunch of problems with the renaming of columns, so that's why I decided this approach instead of doing some weird hack. Eg: `column_A_round`. Also I wouldn't consider `round` as an aggregation, it's more like a `transformation` operation, so that's why I also didn't wanted to mix both concepts

bach/bach/describe.py
```diff
@@ -0,0 +1,183 @@
...
+from typing import Optional, Union, Sequence, List, Type, Dict, Callable
+
+from bach import (
+    DataFrame, Series, SeriesString, SeriesAbstractNumeric, DataFrameOrSeries, SeriesBoolean,
+    get_series_type_from_dtype,
+)
+from bach.sql_model import BachSqlModelBuilder
+
+
+def describe_frame(
```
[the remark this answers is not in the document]
KathiaBarahona: I plan to use it for `Series.describe`, so I'm just avoiding future code duplication

bach/bach/series/series_numeric.py
```diff
@@ -101,7 +101,9 @@ def std(self, partition: WrappedPartition = None, skipna: bool = True, ddof: int
             where N represents the number of elements
         """
         self._ddof_unsupported(ddof)
-        return self._derived_agg_func(partition, 'stddev_samp', skipna=skipna)
+        return self._derived_agg_func(
+            partition, 'stddev_samp', 'double precision', skipna=skipna, cast_db_type=True,
```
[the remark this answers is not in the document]
KathiaBarahona: I started to get errors since db `Numeric` type isn't supported. I didn't wanted to add it to the type register, since I was not really sure how it might affect other functionalities.

bach/bach/series/series_numeric.py
```diff
@@ -120,8 +122,12 @@ def mean(self, partition: WrappedPartition = None, skipna: bool = True) -> 'Seri
         :param partition: The partition or window to apply
         :param skipna: Exclude NA/NULL values
         """
-        return cast('SeriesFloat64',  # for the mypies
-                    self._derived_agg_func(partition, 'avg', 'double precision', skipna=skipna))
+        return cast(
+            'SeriesFloat64',  # for the mypies
+            self._derived_agg_func(
+                partition, 'avg', 'double precision', skipna=skipna, cast_db_type=True,
```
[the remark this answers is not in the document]
thijs-obj: I realise you din't type this. But I think the `'double precision'` here is sort of a bug. `'double precision'` is one of the `dtype_aliases` of `'float64'`, so this works. Most `dtype_aliases` exist for Pandas compatibility (e.g. in pandas you can say `s.astype('f8')` which is the same as `s.astype('float64')`). However, the reason that 'double precision' is an alias is more for historic reasons in our code base than for pandas compatibility.

Anyway long story short, the dtype is `float64`, let's write that here instead of `double precision` as that only obscures things imho

bach/bach/series/series.py
```diff
@@ -919,8 +922,12 @@ def _derived_agg_func(self,
...
+            if cast_db_type and is_supported_db_dtype(derived_dtype):
```
thijs-obj, line 928: I think this is a bit tricky. if `cast_db_type` is set, but `is_supported_db_dtype(derived_dtype)` is False then we'll just not do anything here. But shouldn't we raise an error in that case?

It happens to be that we call this function with `dtype='double precision'`, and `'double precision'` happens to be the `db_dtype` of a `float64` in Postgres, so this works. But if we called this with `dtype='float64', cast_db_type=True` then it would just not do anything

bach/bach/types.py
```diff
@@ -32,6 +32,10 @@ def get_dtype_from_db_dtype(db_dtype: str) -> str:
     return get_series_type_from_db_dtype(db_dtype).dtype  # type: ignore
 
 
+def is_supported_db_dtype(db_dtype: str) -> bool:
+    return db_dtype in _registry.db_dtype_to_series
```
thijs-obj, line 36: There is some weird stuff in `TypeRegistry` because of cyclic imports. As a result if you just use `_registry`, then `_real_init()` might not have run and this function doesn't work yet. Better to add a bit of indirection here and call a function inside `_registry`, that makes sure that `_real_init()` has been called.


For example, I was just curious what this function would yield, so I typed this in ipython:
```
In [1]: from bach.types import is_supported_db_dtype, _registry

In [2]: is_supported_db_dtype('double precision')
Out[2]: False
```

thijs-obj: Or if you know a more elegant solution to the loading of `TypeRegistry` then that could be tackled as a separate thing

bach/bach/dataframe.py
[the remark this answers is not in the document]
thijs-obj: I quite like this approach. Here I only have to look at the code here to understand what happens, otherwise I'll also have to look at `_apply_funct_to_series()`. Do agree with Tom on not mutating the current DataFrame, Pandas doesn't do that either. Think that's fixed now tho :+1:

bach/tests/functional/bach/test_df_describe.py
```diff
@@ -0,0 +1,66 @@
+import numpy as np
+import pandas as pd
+
+from bach import DataFrame
+from bach.describe import DataFrameDescriber, SupportedStats
+from tests.functional.bach.test_data_and_utils import get_bt_with_test_data, assert_equals_data
+
+
+def test_df_categorical_describe() -> None:
```
thijs-obj, line 9: I think we should have some tests here for the actual user interface, i.e. a test that runs `df.describe()`.

Now we only have tests for the code that `df.describe()` calls, but I think in the end the main thing we want is to be sure that we don't have regressions in the end-user's interface

bach/bach/describe.py
```diff
@@ -0,0 +1,184 @@
...
+            stat_df[self.STAT_COLUMN_NAME] = stat
+            stat_df[f'{self.STAT_COLUMN_NAME}_position'] = pos
+            final_df = stat_df if pos == 0 else self._append_stats(final_df, stat_df)
+
+        final_df = final_df.sort_values(by=f'{self.STAT_COLUMN_NAME}_position')
+        final_df = final_df.round(decimals=self.RESULT_DECIMALS)
+        final_df = final_df[[self.STAT_COLUMN_NAME] + columns_to_describe]  # type: ignore
+        return final_df.set_index(self.STAT_COLUMN_NAME)  # type: ignore
+
+    def _append_stats(self, stats_1_df: DataFrame, stats_2_df: DataFrame) -> DataFrame:
```
thijs-obj, line 169: I think this works, but this generates very long sql statement. Would it be possible to `union` all the statistics in one big sql statement?. Something like:

```
select * from {{stat1}}
union
select * from {{stat2}}
union
select * from {{stat3}}
union
select * from {{stat4}}
```

bach/bach/describe.py
```diff
@@ -0,0 +1,184 @@
...
+    datetime_is_numeric: bool
+    percentiles: Optional[Sequence[float]]
+
+    main_stat: SupportedStats = field(init=False)
+
+    STAT_COLUMN_NAME = 'stat'
+    RESULT_DECIMALS = 2
+
+    def __post_init__(self) -> None:
+        self._process_params()
```
thijs-obj, line 75: small comment, but can we not just copy everything that is in `_process_params()` and put it here? Now in reviewing I have to go to another function to see what is actually happening here. Putting it in another function doesn't seem to add anything

bach/bach/series/series_numeric.py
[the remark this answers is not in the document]
KathiaBarahona: sure, the reason I wrote "double precision" is because I needed to cast the column as it, that's why I'm passing `cast_db_type=true`. I know this is a hacky solution, but I had no other way to avoid the `numeric` db type error. If you have a better solution, let me know

bach/bach/describe.py
KathiaBarahona: I rather have a separate function where we dump all sort of validations or "re processing" stuff for now. As I mentioned, we can avoid such things using [pydantic's validators](https://pydantic-docs.helpmanual.io/usage/validators/). We can discuss about it later.

bach/bach/describe.py
```diff
@@ -0,0 +1,184 @@
...
+    SupportedStats.MIN: DataFrame.min,
+    SupportedStats.MAX: DataFrame.max,
+    SupportedStats.UNIQUE: DataFrame.nunique,
+}
+
+FilteringDType = Union[Optional[Union[str, Sequence[str]]], List[Type]]
+
+
+@dataclass
+class DataFrameDescriber:
```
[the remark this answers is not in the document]
KathiaBarahona: 1. Yes, I'd just changed `described_columns` to `series_to_describe`. This is now a property for the class. Also changed `_get_stat_type_based_on_all_series` to `_set_stat_type_based_on_all_series`, so know it's a bit more clear where we actually set both properties.
2. I think it won't be that efficient, since `describe` generates UNIONs, therefore the final query for the dataframe will be a bunch of subqueries containing Unions which we will later need to join, so it actually sounds way more complex than expected.

bach/bach/dataframe.py
```diff
@@ -1940,6 +1940,23 @@ def nunique(self, axis=1, skipna=True, **kwargs):
         return self._aggregate_func('nunique', axis=axis,
                                     level=None, numeric_only=False, skipna=skipna, **kwargs)
 
+    def round(self, decimals: int = 0):
```
thijs-obj, line 1943: Perhaps it would be an idea to put this in another PR with functional tests etc. and merge that PR first? Then this PR becomes a bit smaller

KathiaBarahona: agree!

KathiaBarahona: https://github.com/objectiv/objectiv-analytics/pull/385

bach/bach/describe.py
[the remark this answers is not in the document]
KathiaBarahona: done! Now we call `DescribeOperation` in both classes

bach/bach/describe.py
KathiaBarahona: Done!

bach/bach/describe.py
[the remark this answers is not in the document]
KathiaBarahona: This changed, I followed a similar convention as in `ConcatOperation`. Now we don't set any attribute in any function

bach/bach/describe.py
KathiaBarahona: ended up moving this to `__init__`

bach/bach/describe.py
```diff
@@ -0,0 +1,263 @@
...
+_SUPPORTED_STATS_X_AGG: Dict[SupportedStats, Callable[[DataFrame], DataFrame]] = {
+    SupportedStats.COUNT: DataFrame.count,
+    SupportedStats.MEAN: DataFrame.mean,
+    SupportedStats.STD: DataFrame.std,
+    SupportedStats.MIN: DataFrame.min,
+    SupportedStats.MAX: DataFrame.max,
+    SupportedStats.FIRST: DataFrame.min,
+    SupportedStats.LAST: DataFrame.max,
+    SupportedStats.UNIQUE: DataFrame.nunique,
+    SupportedStats.FREQ: DataFrame.mode,
```
KathiaBarahona, line 61: The result of  `freq` is different than to Pandas. We are returning the most frequent value, meanwhile Pandas returns the frequency number of the most frequent value.

KathiaBarahona: To be honest, I find our result more significant, but that's my personal opinion. @hendrik-obj  what do you think about this? should we stick to Pandas?

jansentom (changes requested):
Hi Kathia, 

Very cool that you got this to work. I do find the "how" quite complex, mainly due to all the dtype magic.

All stats functions in `DataFrame` (e.g. `df.sum()`) do their own series selection based on type if required. Could we not use those directly, and only use include/exclude to filter the series we want to process beforehand?


PS It's really nice that everything is typed.

bach/bach/describe.py
KathiaBarahona: removed!

bach/bach/describe.py
```diff
@@ -0,0 +1,264 @@
...
+    MIN = 'min'
+    MAX = 'max'
+    FIRST = 'first'
+    LAST = 'last'
+    UNIQUE = 'unique'
+    TOP = 'top'
+    FREQ = 'freq'
+
+
+_SERIES_TYPE_X_SUPPORTED_STAT: Dict[Type, Tuple[SupportedStats, ...]] = {
```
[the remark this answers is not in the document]
KathiaBarahona: yup, as agreed I removed this dictionary and evaluate all SupportedStats for supported dtypes.

bach/bach/dataframe.py
```diff
@@ -2190,11 +2190,37 @@ def var(self, axis=1, skipna=True, level=None, ddof: int = 1, numeric_only=False
...
+        include: Optional[Union[str, Sequence[str]]] = None,
+        exclude: Optional[Union[str, Sequence[str]]] = None,
+        datetime_is_numeric: bool = False,
+    ) -> 'DataFrame':
+        """
+        Returns descriptive statistics, it will vary based on what is provided
+
+        :param percentiles: list of percentiles to be calculated. Values must be between 0 and 1.
+        :param include: dtypes to be included, if not provided calculations will be based on numerical columns
+        :param exclude: dtypes to be excluded
```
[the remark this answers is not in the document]
KathiaBarahona: what do you mean? include + exclude?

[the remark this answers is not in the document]
jansentom: ignore that. Not interesting enough.

bach/bach/series/series_numeric.py
jansentom, line 105:
```suggestion
            partition, 'stddev_samp', 'float64', skipna=skipna, cast_db_type=True,
```

bach/bach/series/series_numeric.py
jansentom, line 128:
```suggestion
                partition, 'avg', 'float64', skipna=skipna, cast_db_type=True,
```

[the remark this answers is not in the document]
jansentom: the cast thingy is fine, it's just the "old bug" in there. See suggestion below.

jansentom (commented):
This is starting to look nice! I did a review mostly on structure and details, and left some comments. One to address together tomorrow after standup, the rest small details / nitpicks. Will do a functional review again once we fix the last structural things. See you tmr!

bach/bach/describe.py
```diff
@@ -0,0 +1,213 @@
...
+        stat_df[self.STAT_SERIES_NAME] = stat.value
+        stat_df[f'{self.STAT_SERIES_NAME}_position'] = stat_position
+        return stat_df
+
+    def _calculate_percentiles(self) -> Optional[DataFrame]:
+        """
+        Returns dataframe containing percentiles per each numerical series.
+        """
+        series_to_aggregate = self._get_series_to_aggregate()
+        numerical_series_names = [
```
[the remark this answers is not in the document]
KathiaBarahona: tried to do it as in `_calculate_stat`, but now that I remember we had to define `quantile` method in `Series`, soo it won't work. I'll add a comment explaining this

bach/bach/describe.py
```diff
@@ -0,0 +1,213 @@
...
+    COUNT = 'count'
+    MEAN = 'mean'
+    STD = 'std'
+    MIN = 'min'
+    MAX = 'max'
+    NUNIQUE = 'nunique'
+    MODE = 'mode'
+
+
+_SUPPORTED_SERIES_TYPES: Tuple[Type, ...] = (
```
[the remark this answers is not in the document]
KathiaBarahona: added comment and issue https://github.com/objectiv/objectiv-analytics/issues/430

bach/bach/types.py
KathiaBarahona: removed, we no longer need this

bach/bach/series/series_numeric.py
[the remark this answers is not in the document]
KathiaBarahona: casting hack was removed, seems everything is ok since all the type casting is done by `DataFrameConcatOperation`.

jansentom (approved):
yeeehaa :cowboy_hat_face:

bach/bach/describe.py
```diff
@@ -0,0 +1,214 @@
...
+    if filtering_dtypes == 'all':
+        return _SUPPORTED_SERIES_TYPES
+
+    filtering_dtypes = filtering_dtypes if not isinstance(filtering_dtypes, str) else [filtering_dtypes]
+    return tuple(get_series_type_from_dtype(dtype) for dtype in filtering_dtypes)
+
+
+class DescribeOperation:
+    """
+    In order to implement this class you should provide the following params:
```
thijs-obj, line 47:
```suggestion
    In order to instantiate this class you should provide the following params:
```

bach/bach/describe.py
```diff
@@ -0,0 +1,214 @@
...
+    obj: a DataFrame or Series to be described. If a series is give, it will be transformed into a DataFrame
+    include: A dtype or list of dtypes to be described. If nothing is provided,
+    only numerical series will be considered
+    exclude: A dtype or list of dtype to be excluded from analysis.
+    datetime_is_numeric: A boolean specifying if datetime series should be treated as numeric columns
```
thijs-obj, line 52: Would be good to mention here too that this parameter is not supported.

bach/bach/dataframe.py
```diff
@@ -2212,11 +2212,37 @@ def var(self, axis=1, skipna=True, level=None, ddof: int = 1, numeric_only=False
...
 
+    def describe(
+        self,
+        percentiles: Optional[Sequence[float]] = None,
+        include: Optional[Union[str, Sequence[str]]] = None,
+        exclude: Optional[Union[str, Sequence[str]]] = None,
+        datetime_is_numeric: bool = False,
+    ) -> 'DataFrame':
+        """
+        Returns descriptive statistics, it will vary based on what is provided
```
thijs-obj, line 2223: Can we not be more specific than `it will vary based on what is provided`? The reader doesn't really gain much by reading that.

thijs-obj (approved):
Have some small things on comments. But code wise looks good to me. Don't think this would need another round of review

bach/bach/describe.py
```diff
@@ -0,0 +1,214 @@
...
+            stat_df = self._calculate_stat(stat=stat, stat_position=pos)
+            if stat_df:
+                all_stats_df.append(stat_df)
+
+        percentiles_df = self._calculate_percentiles()
+        if percentiles_df:
+            all_stats_df.append(percentiles_df)
+
+        describe_df = DataFrameConcatOperation(objects=all_stats_df)()
+        describe_df = describe_df.sort_values(by=f'{self.STAT_SERIES_NAME}_position')
```
thijs-obj, line 104: almost forgot. Instead of `f'{self.STAT_SERIES_NAME}_position'` can we not name this column something like `__stat_position`. We could then say that any column starting with double underscores is a reserved name, and users shouldn't use that.

Not really a big deal now, as we have very few users, but it will by quite confusing for users if they happen to have a column that's called `stat_position`. That will give weird results, even tho the users cannot know that we use that column for sorting purposes as it doesn't show up in the results.

PR #375, opened 2022-01-13
model hub improvements

hendrik-obj:
- Different structure of calling methods (map, aggregate, filter)
- New models
  - map
    - is_new_user
    - conversions
    - funnel
  - aggregate
    - session_duration
    - frequency
- Repo structure adjusted
- Notebook updated with new examples

note: metabase class still in series_objectiv.py


analysis/bach_open_taxonomy/bach_open_taxonomy/series/series_objectiv.py
```diff
@@ -11,10 +11,8 @@
 from bach.expression import Expression, quote_string, quote_identifier
 from bach import DataFrame
 from bach.types import register_dtype
-from typing import TYPE_CHECKING
 
-if TYPE_CHECKING:
-    from bach.series import Series
+from bach.series import Series
```
borft, line 15: @thijs-obj changed this, as `bach.Series` is needed by Metabase(), can you check?

thijs-obj: If it works this is fine.

The `if TYPE_CHECKING` is only added as a tool to prevent cyclic imports, but if there are none then it is not needed

PR #376, opened 2022-01-13
React Tracker: Fix and enable validation

sdirosa:
In this PR I've refactored the Validation module (LocationTree) to correctly function in React, also when using SSR or pre-rendered components like Docusaurus does.

Main changes:

- The internal state holding the LocationNode entries of the LocationTree is now a linked list. Each node has a parent link. 
- There are several new helper methods, such as `children`, `rootNode`, `error`.
- Errors are cached by instance id, not by location path, this ensures they are up to date across re-renders.
- LocationContextWrapper is now tracking id changes. These occur whenever a wrapper re-renders.
- New validation: we validate root uniqueness. Before it was possible to have an array of roots in the LocationTree.
- LocationTree can infer Tracker Plugins location mutations. This ensures the validation is aware of what Plugins do.
- BUGFIX: Location removal now cleans up also the errorCache. This cache ensures we don't over-report errors.
- Improvement: Location removal now clears up orphaned nodes. This keeps the location list in sync with the UI.
- BUGIFX: TrackerProvider initializes LocationTree instead of ObjectivProvider. This allows using multiple trackers.
- Improvement: The relevant Providers and Wrappers auto detect the environment and enable validation accordingly. 
- Improvement: We do not throw errors anymore if we can't infer id and title from Pressables, instead we console.error.
- Improvement: Whenever we cannot infer id or title we help pinpoint where to look by hinting the LocationStack.
- Several small fixes and improvements to existing tests. Mostly related to using ReactTracker instead of Core Tracker.

To Do:

- [x] Manual testing in Website
- [x] Manual testing in ROD

### DIFF
--- a/tracker/trackers/browser/src/BrowserTracker.ts
+++ b/tracker/trackers/browser/src/BrowserTracker.ts
@@ -55,7 +55,7 @@ export class BrowserTracker extends Tracker {
     }
 
     // If node is in `development` mode and console has not been configured, automatically use the browser's console
-    if (!config.console && process.env.NODE_ENV?.startsWith('dev')) {
+    if (config.console === undefined && process.env.NODE_ENV?.startsWith('dev')) {
       config.console = console;
     }
 
--- a/tracker/trackers/react/src/common/LocationTree.ts
+++ b/tracker/trackers/react/src/common/LocationTree.ts
@@ -7,55 +7,21 @@ import { getLocationPath } from '@objectiv/tracker-core';
 import { LocationContext } from '../types';
 
 /**
- * TODO: this module is highly experimental and currently not enabled
- */
-
-/**
- * LocationTree nodes have the same shape of LocationContext but they can have children LocationNodes themselves.
+ * LocationTree nodes have the same shape of LocationContext, but they can have children LocationNodes themselves.
  */
 export type LocationNode = LocationContext<AbstractLocationContext> & {
   /**
-   * An array of LocationNode objects, which may contain more children themselves.
+   * The parent LocationNode identifier.
    */
-  children: LocationNode[];
+  parentLocationId: string | null;
 };
 
 /**
- * LocationTree root node has `id`, `_type` and `__location_id` hardcoded to null. It's just a container.
- */
-export type LocationRootNode = {
-  /**
-   * LocationRootNode id can only be null.
-   */
-  id: null;
-
-  /**
-   * LocationRootNode _type can only be null.
-   */
-  _type: null;
-
-  /**
-   * LocationRootNode __location_id can only be null.
-   */
-  __location_id: null;
-
-  /**
-   * An array of LocationNode objects, which may contain more children themselves.
-   */
-  children: LocationNode[];
-};
-
-/**
- * Internal state to hold the complete known LocationTree.
+ * Internal state to hold a complete list of all known LocationNodes.
  * Each node, exception made for the root one, is a uniquely identifiable Location Context.
- * All nodes may have children LocationNodes.
+ * All nodes, except the root ine, have a parent LocationNode.
  */
-const locationTree: LocationRootNode = {
-  id: null,
-  _type: null,
-  __location_id: null,
-  children: [],
-};
+export let locationNodes: LocationNode[] = [];
 
 /**
  * Internal state to keep track of which identifiers are already known for a certain issue. This is used to prevent
@@ -63,10 +29,10 @@ const locationTree: LocationRootNode = {
  *
  * NOTE: Currently we support only `collision` issues. As more checks are implemented this Map may change.
  */
-const errorCache = new Map<string, 'collision'>();
+export const errorCache = new Map<string, 'collision'>();
 
 /**
- * LocationTree is a global object providing a few utility methods to interact with the `locationTree` global state.
+ * LocationTree is a global object providing a few utility methods to interact with the `locationNodes` global state.
  * LocationContextWrapper makes sure to add new LocationNodes to the tree whenever a Location Wrapper is used.
  */
 export const LocationTree = {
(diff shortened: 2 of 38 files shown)


tracker/trackers/browser/src/BrowserTracker.ts
```diff
@@ -55,7 +55,7 @@ export class BrowserTracker extends Tracker {
     }
 
     // If node is in `development` mode and console has not been configured, automatically use the browser's console
-    if (!config.console && process.env.NODE_ENV?.startsWith('dev')) {
+    if (config.console === undefined && process.env.NODE_ENV?.startsWith('dev')) {
       config.console = console;
     }
```
sdirosa, line 60: This change allows developers to disable console in dev, by setting it to a `falsy` value for example

tracker/trackers/react/src/common/LocationTree.ts
```diff
@@ -3,170 +3,224 @@
...
+        locationContext: LocationContext<AbstractLocationContext>
+      ) => {
+        LocationTree.add(locationContext, parentLocationContext);
+        // Update the rootId to the last LocationContext we add this way. This effectively updates the root node.
+        rootId = locationContext.__location_id;
+        return locationContext;
+      },
+      null
+    );
   },
```
sdirosa, line 117: I plan to get rid of all of this initialization as part of the Debuggability epic. Currently Plugins decorate the LocationStack at runtime only when the event gets triggered, but that's an issue for whenever we'd like to determine the location stack of an element before triggering it.

tracker/trackers/react/src/common/LocationTree.ts
```diff
@@ -3,170 +3,224 @@
...
+        accumulator.push(locationNode);
+      }
+      return accumulator;
+    }, [] as LocationNode[]);
+
+    // Keep running until the cleaned up tree stop changing in size
+    if (sizeBeforeCleanup !== locationNodes.length) {
+      LocationTree.remove(locationContext);
+    }
+  },
```
sdirosa, line 225: This removal and cleanup ensures that whenever a root element gets removed and re-added, also all of its previous errors and children get re-processed by the LocationTree. In an SPA that equals to catching routing and UI changes that involve whole branches.

tracker/trackers/react/src/common/factories/makeIdFromString.ts
```diff
@@ -26,12 +26,12 @@ export const makeIdFromString = (sourceString: string): string => {
...
+      .slice(0, 64);
   }
 
-  // Throw if we did not manage to generate a valid id
-  if (!id) {
-    throw new Error('Could not generated a valid id. Please provide one manually.');
+  // Catch empty strings and return null
+  if (!id || !id.length) {
+    return null;
   }
```
sdirosa, line 35: Throwing is always dangerous in these Trackers. If we make a mistake in not catching any of these errors we may end up crashing the target application. I've switched to console logging and I've added a fallback in the using components to proxy the UI without changes whenever an error occurs, instead of stopping the rendering.

Basically if we can't determine an Id or Title now the original component gets rendered as is, along with an error in console. This ensures the App is unaffected while developers fix the issue.

tracker/trackers/react/src/common/isDevMode.ts
```diff
@@ -0,0 +1,9 @@
+/*
+ * Copyright 2021-2022 Objectiv B.V.
+ */
+
+/**
+ * Helper function to determine if we are in development mode.
+ * Determines whether Node environment is development and if we are in a browser by checking the window object.
+ */
+export const isDevMode = () => process.env.NODE_ENV?.startsWith('dev') && typeof window !== 'undefined';
```
sdirosa, line 9: This is a good candidate to be moved to the Core Tracker, but I'll do that when I get to refactor Browser Tracker to use it.

tracker/trackers/react/src/common/providers/ObjectivProvider.tsx
```diff
@@ -59,7 +58,6 @@ export const ObjectivProvider = ({ children, tracker, options }: ObjectivProvide
     if (trackApplicationLoaded) {
       trackApplicationLoadedEvent({ tracker });
     }
-    LocationTree.clear();
   });
```
sdirosa, line 61: This is now responsibility of TrackerProvider.

tracker/trackers/react/src/locationWrappers/LocationContextWrapper.tsx
```diff
@@ -30,9 +36,26 @@ export type LocationContextWrapperProps = {
...
+      LocationTree.remove(locationContext);
+    }
+  });
+
+  useOnChange<LocationContext<AbstractLocationContext>>(locationContext, (previousLocationContext) => {
+    if (isDevMode()) {
+      LocationTree.remove(previousLocationContext);
+      LocationTree.add(locationContext, parentLocationContext);
+    }
+  });
```
sdirosa, line 58: Since Root locations are also based on LocationContextWrapper, when they change the LocationTree gets correctly updated. Basically all of the previous LocationNodes are unmounted and garbage collected, and a new tree gets asynchronously created from scratch. 

This ensures the LocationTree is always in sync with the UI.

tracker/trackers/react/src/trackedContexts/TrackedLinkContext.tsx
```diff
@@ -62,6 +57,21 @@ export const TrackedLinkContext = React.forwardRef<HTMLElement, TrackedLinkConte
     ...(forwardHref ? { href } : {}),
   };
 
+  if (!linkId) {
+    const locationPath = getLocationPath(useLocationStack());
+    console.error(
+      `｢objectiv｣ Could not generate a valid id for LinkContext @ ${locationPath}. Please provide either the \`title\` or the \`id\` property manually.`
+    );
+    return React.createElement(Component, componentProps);
+  }
```
sdirosa, line 66: As mentioned above, we now return a clone of the original element (no onClick handler added) whenever we cannot generate an Id automatically. Same for Pressable.

tracker/trackers/react/tests/TrackedLinkContext.test.tsx
```diff
@@ -79,8 +85,30 @@ describe('TrackedLinkContext', () => {
...
+          </TrackedDiv>
+        </TrackedRootLocationContext>
+      </ObjectivProvider>
+    );
+
+    expect(console.error).toHaveBeenCalledTimes(1);
+    expect(console.error).toHaveBeenCalledWith(
+      '｢objectiv｣ Could not generate a valid id for LinkContext @ RootLocation:root / Content:content. Please provide either the `title` or the `id` property manually.'
+    );
+  });
```
sdirosa, line 108: This test showcases a new debuggability improvement. If we can't generate an id we tell the user where to look for the issue by adding the known location of the problematic Tracked element.

sdirosa:
Tested thoroughly in the website, also triggering several of the exceptions we now catch. This is ready for testing on staging.

PR #377, opened 2022-01-14
Add release.yml for auto-generation of changelogs

ivarpruijn:
This configuration helps to automatically generate changelogs based on labels of PRs. See [docs here](https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes).

Configuration options are fairly limited for our monorepo, but this is a nice first step to help with generating changelogs.

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,28 @@
+changelog:
+  exclude:
+    labels:
+      - "Release: Ignore for Changelog"
+  categories:
+    - title: Breaking Changes 🛠
+      labels:
+        - "Type: Breaking Change"
+    - title: New Features 🎉
+      labels:
+        - "Type: Epic"
+        - "Type: Story"
+        - "Type: Enhancement"
+    - title: Fixes
+      labels:
+        - "Type: Bug"
+    - title: Docs
+      labels:
+        - "Area: Docs"
+    - title: Demo notebooks
+      labels:
+        - "Area: Demo"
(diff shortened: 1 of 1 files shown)


sdirosa:
This seems to be the way to go, or indeed a starting point, but I think we have quite some work to do.

1. I never use labels myself and often work on epics that do not reflect with the issues in timeline. Is it something I should start worrying about?

2. Pretty much everyone at some point created a branch for a purpose and then ended up doing a lot more in it than expected (example: The Taxonomy update we released recently). I think we have to be sure branches are about one change and one only if we want to create meaningful changelogs from them. Or is it okay to tweak those cases by hand?

3. Regarding branches themselves, it doesn't look like we have any sort of standard naming for our branches so the log may be confusing at times. Do we want to adopt some standard for branch names and commits, such as semantic releases, etc?

That said I'm also fine in merging this as is and tweak stuff manually until the automation suits us nicely.

Perhaps related: when merging I often receive these failures: https://github.com/objectiv/objectiv-analytics/runs/4816031730?check_suite_focus=true

I'm not sure if this influences automatic labeling, I thought I'd mention it just in case.

ivarpruijn:
@sdirosa Did a quick check of some things we discussed recently, but will take more time, so leaving it for now. I think we can merge this as-is, will make changelogs a bit easier to generate.

sdirosa (approved):

PR #385, opened 2022-01-19
round implementation

KathiaBarahona:


bach/tests/functional/bach/test_df.py
```diff
@@ -115,3 +117,17 @@ def test_boolean_indexing_same_node():
             [11, 11, 'Dokkum', 'Noardeast-Fryslân', 12675, 1298]
         ]
     )
+
+
+def test_round():
```
thijs-obj, line 122: I like that we check here against what pandas does, that's a nice check.

I do think we should test more cases:
1. a column that shouldn't round at all (e.g. strings)
2. It's seems there is a bug in `SeriesAbstractNumeric.round()` that it changes the value of integers into floats. This bug is only obvious when looking at the actual data output, the dtype still says 'int64'. Would be great if we would catch that in a test and fix it (to reproduce, create dataframe called `bt` with integers, and check: `np.testing.assert_equal(bt.round().to_pandas().dtypes.values, bt.to_pandas().dtypes.values)`)

bach/tests/functional/bach/test_df.py
```diff
@@ -115,3 +117,17 @@ def test_boolean_indexing_same_node():
...
+    pdf = pd.DataFrame(
+        data={
+            'a': [1.9, 3.0, 4.123, 6.425124, 2.00000000001, 2.1, np.nan, 7.],
+            'b': [11.9, 32.0, 43.123, 64.425124, 25.00000000001, 6, np.nan, 78.],
+        },
+    )
+    bt = get_from_df('test_round', pdf)
+
+    for i in 0, 3, 5, 9:
+        np.testing.assert_equal(pdf.round(i).values, bt.round(i).values)
```
thijs-obj, line 132: There is a subtle bug here that won't always occur:  `bt.round(i).values` is not sorted, depending on the order in which Postgres returns the rows this test might fail

bach/tests/functional/bach/test_df.py
```diff
@@ -115,3 +117,17 @@ def test_boolean_indexing_same_node():
...
+        np.testing.assert_equal(pdf.round(decimals=i).values, bt.round(decimals=i).values)
```
thijs-obj, line 133: I created an issue to migrate from `values` to `to_numpy()` in the future: https://github.com/objectiv/objectiv-analytics/issues/387

Not directly related to this PR, but running this in a debugger was the trigger to create the issue.

FYI @jansentom @hendrik-obj mentioning you guys too, as we don't really have  a weekly 'go over the backlog session'

bach/tests/functional/bach/test_df.py
KathiaBarahona: I'd changed `np.testing` with `pd.testing`. I'd been experiencing some weird trouble when comparing nan values.

bach/tests/functional/bach/test_df.py
KathiaBarahona: 1. Added a string column to test
2. Yup, the int64 series shouldn't have any effect when doing round(). Fixed it on the series

Issue #395, opened 2022-01-20
Feature: implement DataFrame.sort_index()


thijs-obj:
Implement most basic functionality, inline with Series.sort_index().
We can probably leverage the data that `DataFrame.sort_values()` already stores for sorting

See https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.sort_index.html

We aim to support `level`, `inplace`, `ascending` parameters

PR #398, opened 2022-01-21
df.sort_index implementation

KathiaBarahona:
https://github.com/objectiv/objectiv-analytics/issues/395

### DIFF
--- a/bach/bach/dataframe.py
+++ b/bach/bach/dataframe.py
@@ -1556,7 +1556,7 @@ class DataFrame:
             if isinstance(idx_l, str) and idx_l not in self.index_columns:
                 raise ValueError(f'dataframe has no {idx_l} index level.')
 
-            level_name = idx_l if isinstance(idx_l, str) else self.index_columns[idx_l]  # type: ignore
+            level_name = idx_l if isinstance(idx_l, str) else self.index_columns[idx_l]
             index_columns.append(level_name)
 
         return index_columns
--- a/bach/bach/dataframe.py
+++ b/bach/bach/dataframe.py
@@ -1536,7 +1536,7 @@ class DataFrame:
         :returns: a new DataFrame with the specified ordering if ``inplace=False``,
          otherwise it updates the original and returns None.
         """
-        sort_by = self.index_columns if not level else self._get_index_columns_by_level(level)
+        sort_by = self.index_columns if not level else self._get_indexes_by_level(level)
         df = self.sort_values(by=sort_by, ascending=ascending)
 
         if inplace:
@@ -1545,8 +1545,8 @@ class DataFrame:
 
         return df
 
-    def _get_index_columns_by_level(self, level: Level) -> List[str]:
-        index_columns = []
+    def _get_indexes_by_level(self, level: Level) -> List[str]:
+        selected_indexes = []
         nlevels = len(self.index_columns)
 
         for idx_l in level if isinstance(level, list) else [level]:
@@ -1556,10 +1556,10 @@ class DataFrame:
             if isinstance(idx_l, str) and idx_l not in self.index_columns:
                 raise ValueError(f'dataframe has no {idx_l} index level.')
 
-            level_name = idx_l if isinstance(idx_l, str) else self.index_columns[idx_l]
-            index_columns.append(level_name)
+            level_name = idx_l if isinstance(idx_l, str) else self.index_columns[idx_l]  # type: ignore
+            selected_indexes.append(level_name)
 
-        return index_columns
+        return selected_indexes
 
     def to_pandas(self, limit: Union[int, slice] = None) -> pandas.DataFrame:
         """
--- a/bach/bach/dataframe.py
+++ b/bach/bach/dataframe.py
@@ -1556,7 +1556,7 @@ class DataFrame:
             if isinstance(idx_l, str) and idx_l not in self.index_columns:
                 raise ValueError(f'dataframe has no {idx_l} index level.')
 
-            level_name = idx_l if isinstance(idx_l, str) else self.index_columns[idx_l]  # type: ignore
+            level_name = idx_l if isinstance(idx_l, str) else self.index_columns[idx_l]
             selected_indexes.append(level_name)
 
         return selected_indexes
--- a/bach/bach/dataframe.py
+++ b/bach/bach/dataframe.py
@@ -35,6 +35,8 @@ ColumnFunction = Union[str, Callable, List[Union[str, Callable]]]
 #     - list of functions and/or function names, e.g. [SeriesInt64.sum, 'mean']
 #     - dict of axis labels -> functions, function names or list of such.
 
+Level = Union[int, List[int], str, List[str]]
+
 
 class SortColumn(NamedTuple):
     expression: Expression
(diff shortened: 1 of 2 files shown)


bach/bach/dataframe.py
```diff
@@ -1518,6 +1520,48 @@ def sort_values(
...
+        df = self.sort_values(by=sort_by, ascending=ascending)
+
+        if inplace:
+            self._update_self_from_df(df)
+            return None
+
+        return df
+
+    def _get_index_columns_by_level(self, level: Level) -> List[str]:
+        index_columns = []
```
jansentom, line 1549: I find it a bit confusing that `index_columns` has the same name as `self.index_columns`. Perhaps `sort_series` would be better?

bach/bach/dataframe.py
```diff
@@ -1518,6 +1520,48 @@ def sort_values(
...
+    def sort_index(
+        self,
+        level: Optional[Level] = None,
+        ascending: Union[bool, List[bool]] = True,
+        inplace: bool = False,
+    ) -> Optional['DataFrame']:
+        """
+        Sort dataframe by index levels.
+        :param level: int or level name or list of ints or level names.
+         If not specified, all index columns are used
```
jansentom, line 1532: We try to abstract away from the database here. Please use `series` instead of `columns`.  (Applies in a few other places as well)

bach/tests/functional/bach/test_df_sort_index.py
```diff
@@ -0,0 +1,86 @@
...
+        btr.sort_index(level=[0, 'random'])
+
+    result = btr.sort_index(level=['platforms', 0, 'platforms'])
+    assert_equals_data(
+        result,
+        expected_columns=['town', 'platforms', 'station_id', 'station'],
+        expected_data=[
+            ['Drylts', 1, 1, 'IJlst'],
+            ['It Hearrenfean', 1, 2, 'Heerenveen'],
+            ['Ljouwert', 1, 5, 'Camminghaburen'],
```
jansentom, line 43: I kind of like that you're using these unpronounceable fixtures too :)

bach/bach/dataframe.py
```diff
@@ -1518,6 +1520,48 @@ def sort_values(
...
+        nlevels = len(self.index_columns)
+
+        for idx_l in level if isinstance(level, list) else [level]:
+            if isinstance(idx_l, int) and nlevels < idx_l:
+                raise ValueError(f'dataframe has only {nlevels} levels.')
+
+            if isinstance(idx_l, str) and idx_l not in self.index_columns:
+                raise ValueError(f'dataframe has no {idx_l} index level.')
+
+            level_name = idx_l if isinstance(idx_l, str) else self.index_columns[idx_l]  # type: ignore
```
jansentom, line 1559: We should probably fix the reason the type ignore is required instead.

bach/bach/dataframe.py
```diff
@@ -1518,6 +1520,48 @@ def sort_values(
...
+            if level_name not in index_columns:
```
jansentom, line 1560: I don't think this is correct as the length of the returned list now has a different length than the length of the ascending list in the calling method? (see comment in tests below too)

bach/tests/functional/bach/test_df_sort_index.py
```diff
@@ -0,0 +1,86 @@
...
+    with pytest.raises(ValueError, match=r'dataframe has only 2 levels'):
+        btr.sort_index(level=3)
+
+    with pytest.raises(ValueError, match=r'dataframe has no random index level'):
+        btr.sort_index(level='random')
+
+    with pytest.raises(ValueError, match=r'dataframe has no random index level'):
+        btr.sort_index(level=[0, 'random'])
+
+    result = btr.sort_index(level=['platforms', 0, 'platforms'])
```
jansentom, line 36: This breaks if you give `ascending=[True,True,True]`. We should probably test for that and fix it.

jansentom (changes requested):
Very nice already. Some small fixes required, but almost ready to ship I think! Should be doable before drinks :beer:

bach/bach/dataframe.py
KathiaBarahona: mmm IMHO it's better `index_columns` just in case other future methods might need to filter out the indexes by level. I don't want it to be sticked just for sorting implementations :)

bach/bach/dataframe.py
KathiaBarahona: good catch, I'll probably remove this condition and see if that causes any error

bach/bach/dataframe.py
KathiaBarahona: actually, there was no need. I thought so because pycharm was complaining

bach/bach/dataframe.py
jansentom: What about `selected_columns` then? :) If you disagree, @thijs-obj can decide what he likes better ;)

bach/bach/dataframe.py
```diff
@@ -1518,6 +1520,47 @@ def sort_values(
...
+        Sort dataframe by index levels.
+        :param level: int or level name or list of ints or level names.
+         If not specified, all index series are used
+        :param ascending: Whether to sort ascending (True) or descending (False). If this is a list, then the
+            `level` must also be a list and ``len(ascending) == len(level)``.
+        :param inplace: Perform operation on self if ``inplace=True``, or create a copy.
+        :returns: a new DataFrame with the specified ordering if ``inplace=False``,
+         otherwise it updates the original and returns None.
+        """
+        sort_by = self.index_columns if not level else self._get_index_columns_by_level(level)
```
jansentom, line 1539: I realise that you are not responsible for the naming of `self.index_columns`, but that it'd be better to name that `index_series`. If you and Thijs agree, we can add that to the backlog

jansentom (approved):
LGMT! Some stuff to discuss with @thijs-obj , but no blockers from my side to merge.

bach/bach/dataframe.py
KathiaBarahona: I agree with that name :+1:

bach/bach/dataframe.py
KathiaBarahona: If we want to avoid using the term `column`, let's rename it :)

jansentom: Yeah, I'd like that for all stuff that's not directly SQL. It's probably a bit big to include in this PR. Can you make an issue so we can include it on Monday?

bach/tests/functional/bach/test_df_sort_index.py
```diff
@@ -0,0 +1,86 @@
...
+    assert_equals_data(
+        result,
+        expected_columns=['town', 'platforms', 'station_id', 'station'],
+        expected_data=[
+            ['Drylts', 1, 1, 'IJlst'],
+            ['It Hearrenfean', 1, 2, 'Heerenveen'],
+            ['Ljouwert', 1, 5, 'Camminghaburen'],
+            ['It Hearrenfean', 2, 3, 'Heerenveen IJsstadion'],
+            ['Snits', 2, 7, 'Sneek Noord'],
+            ['Snits', 2, 6, 'Sneek'],
```
thijs-obj, line 46: This test is not deterministic. If we sort on `['platforms', 'town', 'platforms']` then these two rows might be returned as either
```
            ['Snits', 2, 7, 'Sneek Noord'],
            ['Snits', 2, 6, 'Sneek'],
```
or
```
            ['Snits', 2, 6, 'Sneek'],
            ['Snits', 2, 7, 'Sneek Noord'],
```
Both are correct.


This might be related to Tom's comment about this breaking when `ascending=[True, True, True]` ?

bach/bach/dataframe.py
thijs-obj: I agree that consistent naming is very important.

We have used columns in quite a few places already, but I think Tom is right that we probably should use 'Series' instead of 'columns' in most places.

Generally I am not sure that we should automatically always stick with the Pandas nomenclature tho? Some of the Pandas naming is really vague I think, and not very accessible for people from outside pandas. E.g. 'level' for indices, took me quite a while to figure out that that meant columns :) Perhaps we could use all the pandas naming, but also use 'sql' terminology like "columns" in the docstrings? Let's discuss this in the Data Science channel, or a meeting.

thijs-obj (commented):
One small comment about a test, otherwise looks good to me

bach/tests/functional/bach/test_df_sort_index.py
KathiaBarahona: mm nope, the issue with the `ascending=[True, True, True]` is that I was just considering one time each level. So if user provides `['platforms', 'town', 'platforms']` with a list of ascending values, it will raise an error since `len(level) != len(ascending)`. 

Tho, your comment is valid. I will fix this test :+1:

bach/bach/dataframe.py
thijs-obj: but now it's back

PR #402, opened 2022-01-24
abstraction of concat operation, SeriesConcatOperation

KathiaBarahona:
https://github.com/objectiv/objectiv-analytics/issues/403


bach/bach/concat.py
```diff
@@ -1,39 +1,115 @@
...
+def get_merged_series_dtype(dtypes: Set[str]) -> str:
+    if len(dtypes) == 1:
+        return dtypes.pop()
+    elif all(
+        issubclass(get_series_type_from_dtype(dtype), SeriesAbstractNumeric)
+        for dtype in dtypes
+    ):
+        return'float64'
+
+    return DEFAULT_CONCAT_SERIES_DTYPE
```
KathiaBarahona, line 28: Not sure if this clear or the right place to define it

bach/bach/dataframe.py
```diff
@@ -2248,16 +2248,17 @@ def append(
...
 
         other_dfs = other if isinstance(other, list) else [other]
-        return ConcatOperation(
-            dfs=[self] + other_dfs,
+        concatenated_df = DataFrameConcatOperation(
+            objects=[self] + other_dfs,
             ignore_index=ignore_index,
             sort=sort,
         )()
+        return concatenated_df if isinstance(concatenated_df, DataFrame) else concatenated_df.to_frame()
```
KathiaBarahona, line 2261: always expecting a dataframe, mypy raises typing error otherwise

bach/bach/concat.py
thijs-obj: Might be nice to have a comment on why 'string' is the default. Is that what pandas does, or is that just because that's the only type we can always cast to? I honestly don't know and was wondering while reading

bach/bach/concat.py
```diff
@@ -1,39 +1,120 @@
...
+
+
 @dataclass
-class ConcatOperation:
-    dfs: List[DataFrame]
+class ConcatOperation(ABC):
```
thijs-obj, line 32: If we inherit from `ABC` here, then it would be nice to annotate the abstract methods with `@abstractmethod`. That way it will give earlier errors if a subclass doesn't implement all methods.

bach/bach/concat.py
```diff
@@ -1,39 +1,120 @@
...
+    objects: Sequence[DataFrameOrSeries]
```
thijs-obj, line 33: I think the fact that objects can be a `Sequence[DataFrameOrSeries]` is a potential source of errors, and makes the code less nice. If I pass something like `objects=[data_frame, series, another_data_frame]` then mypy will not complain.

Some ideas:
1) Just have two separate classes that don't share a common base class. There is not a lot of shared code anyway, and what is shared could be factored out like already happened with `get_merged_series_dtype` and `ConcatSqlModel`. Nor does having a base class hugely improves some other aspect of the code, IMHO.


2) Alternative: Can we not use Generics here? Something like 
```python
T = TypeVar("T")

@dataclass
class ConcatOperation(Generic[T]):
    objects: Sequence[T]
    ...

class DataFrameConcatOperation(ConcatOperation[DataFrame]):
    ....


class SeriesConcatOperation(ConcatOperation[Series]):
    ...
```
Note: didn't test this, just wondering whether this is possible

bach/bach/concat.py
```diff
@@ -1,39 +1,120 @@
...
-from bach.sql_model import BachSqlModel, construct_references, get_variable_values_sql, filter_variables
+from bach.sql_model import BachSqlModel, construct_references
 from bach.utils import ResultSeries, get_result_series_dtype_mapping
 from sql_models.model import CustomSqlModelBuilder, Materialization
 
 
+DEFAULT_CONCAT_SERIES_DTYPE = 'string'
+
+
+def get_merged_series_dtype(dtypes: Set[str]) -> str:
```
thijs-obj, line 19: If this is a public function (i.e. no `_` in front of it) then we should expect python code from outside this file to use this function too. But for that we would need some docstring here I think. e.g. the fact that it modifies `dtypes` might be quite surprising

bach/bach/series/series.py
```diff
@@ -1215,6 +1215,32 @@ def window_nth_value(self, n: int, window: WrappedWindow = None):
...
+        Append rows of other series to the caller series.
+
+        :param other: objects to be added
+        :param ignore_index: if true, drops indexes of all objects to be appended
+
+        :return: a new series with all rows from appended other.
+        """
+        from bach.concat import SeriesConcatOperation
+        if not other:
+            raise ValueError('no series to append.')
```
thijs-obj, line 1233:
```suggestion
            return self
```
Seems that this is what pandas does, and I think it is a bit more logical too. saves our end users an edge-case that they have to handle

The docstring should then be something like
`:return: a new series with all rows from appended other or self if other is empty`

bach/bach/concat.py
KathiaBarahona: wow! thanks a lot. Now I can remove some dumb validations, never used Generic typing before.

bach/bach/concat.py
KathiaBarahona: Pandas actually cast it as "object", but in our case we don't have a way to have multiple object types in a column, so what it seems reasonable for me was just to cast them as "text", unless we actually want an actual SQL error to be raised.

bach/tests/functional/bach/test_series_append.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+@patch.object(SeriesConcatOperation, '__call__')
+def test_series_append_errors(mocked_concat_call: MagicMock) -> None:
+    bt = get_bt_with_test_data(full_data_set=False)
+    mocked_concat_call.return_value = bt
+    with pytest.raises(ValueError, match='no series to append'):
+        bt.city.append(other=[])
+
+    with pytest.raises(Exception, match='concatenated series should result in new series.'):
```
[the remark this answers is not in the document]
KathiaBarahona: will remove this test, I had some mypy issues because of the return type (DataFrameOrSeries).

bach/tests/functional/bach/test_series_append.py
```diff
@@ -0,0 +1,114 @@
+from unittest.mock import patch, MagicMock
+
+import pytest
+
+from bach import Series
+from bach.concat import SeriesConcatOperation
+from tests.functional.bach.test_data_and_utils import get_bt_with_test_data, assert_equals_data
+
+
+@patch.object(SeriesConcatOperation, '__call__')
```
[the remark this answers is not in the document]
KathiaBarahona: I use `@patch` just because  `monkeypatch` is not my cup of tea, but since we want consistency I'll change it :+1:

bach/bach/concat.py
```diff
@@ -1,39 +1,133 @@
...
+        self.ignore_index = ignore_index
+        self.sort = sort
+
+    def __call__(self, *args, **kwargs) -> TDataFrameOrSeries:
+        if not len(self.objects):
+            raise ValueError('no objects to concatenate.')
+
+        if len(self.objects) == 1:
+            index = {} if self.ignore_index else self.objects[0].index
+            return self.objects[0].copy_override(index=index)  # type: ignore
```
thijs-obj, line 52: `# type: ignore` is needed here, but the other three places in this file it can be removed.

Also think we should strive to get rid of this one, but it looks like fixing that is not completely trivial as there is some buggy code relying on the return type of `Series.copy_override()` not being defined. So for now let's only remove the other three instances

bach/bach/concat.py
```diff
@@ -1,39 +1,133 @@
...
+            series_name: ResultSeries(
+                name=series_name,
+                expression=Expression.identifier(series_name),
+                dtype=_get_merged_series_dtype(series_dtypes[series_name]),
+            )
+            for series_name in series_names
+        }
+
+    @abstractmethod
+    def _get_series(self) -> Dict[str, ResultSeries]:
```
thijs-obj, line 94: Both `_get_series()`, `_join_series_expressions()`, and `_get_model()` are never called by the Abstract class.

I don't think it makes sense to define them here as abstract functions. If the subclasses don't use those functions but still give the correct result for `_get_concatenated_object()`  then this parent class shouldn't care about how that's achieved.

bach/bach/concat.py
```diff
@@ -1,39 +1,133 @@
...
+        raise NotImplementedError()
+
+    @abstractmethod
+    def _join_series_expressions(self, obj: TDataFrameOrSeries) -> Expression:
+        raise NotImplementedError()
+
+    @abstractmethod
+    def _get_concatenated_object(self) -> TDataFrameOrSeries:
```
thijs-obj, line 102: We should have some docstring here explaining what this parent class expects from its child classes with regard to this function. What should this return?

Doesn't need to be long, but not having a specification of what a subclass should do means there is no real abstraction here. The implementer/reviewer/debugger of the subclass will still have to read the parent class to understand what this function should do.

bach/tests/functional/bach/test_series_append.py
```diff
@@ -0,0 +1,98 @@
+from bach import Series
+from tests.functional.bach.test_data_and_utils import get_bt_with_test_data, assert_equals_data
+
+
+def test_series_append_same_dtype() -> None:
```
thijs-obj, line 5: nice tests :+1:

bach/tests/functional/bach/test_series_append.py
```diff
@@ -0,0 +1,98 @@
...
+        expected_columns=['city_skating_order'],
+        expected_data=[
+            ['1'],
+            ['2'],
+            ['3'],
+            ['Drylts'],
+            ['Ljouwert'],
+            ['Snits'],
+        ],
+    )
```
thijs-obj, line 98: One interesting test that's missing is the case where two series have the same index name, but different types.
E.g. something like this (quick copy-paste-edit, not claiming this is perfect:) :
```python
def test_series_append_same_dtype_different_index() -> None:
    bt = get_bt_with_test_data(full_data_set=False)[['city', 'skating_order']]
    bt_other_index = bt.reset_index(drop=False)
    bt_other_index['_index_skating_order'] = bt_other_index['_index_skating_order'].astype('string')
    bt_other_index = bt_other_index.set_index('_index_skating_order')
    bt.skating_order = bt.skating_order.astype(str)

    result = bt_other_index.city.append(bt.skating_order)
    assert isinstance(result, Series)

    assert_equals_data(
        result.sort_values(),
        expected_columns=['_index_skating_order', 'city_skating_order'],
        expected_data=[
            ['1', '1'],
            ['2', '2'],
            ['3', '3'],
            ['3', 'Drylts'],
            ['1', 'Ljouwert'],
            ['2', 'Snits'],
        ],
    )
```

I'd expect this either to work, or to get a clear error, but now we get a database error

thijs-obj (commented):
Left some comments. One real bug I think (the sql-error when index dtypes are not matching) but throwing a better exception earlier doesn't seem hard, and the other things are also quite small I think.

**Some comments missing**
However, I think this code deserves more comments. If I want to append series I have to create an `SeriesConcatOperation` object and then execute the instance as a function. Only way to find that out is by looking at the code, that doesn't really abstract the complexity of this 300 line file away for the programmer that simply wants to call a function to append dataframes (that programmer could be me, a future contributor, or somebody else). If there was some comment that explained how to use that, then there would be no need to look at the actual code for such a person.

Additionally I think  there could be some comments about what the parent class expects the child classes to do, and how they work together to implement the functionality. This doesn't need to be a lot, but currently there is nothing really.

Generally I think we should optimize more for reviewers, people reading the code, future contributors, and future me (because I'll forget what I read today pretty soon).

bach/tests/functional/bach/test_series_append.py
KathiaBarahona: will raise an error

bach/bach/concat.py
KathiaBarahona: removed them.

bach/bach/concat.py
KathiaBarahona: added some docstrings about how to implement each subclass, what the parent class should expect in this method

thijs-obj (approved):
lgtm

PR #404, opened 2022-01-25
quantile implementation

KathiaBarahona:
https://github.com/objectiv/objectiv-analytics/issues/410

Depends on series.append https://github.com/objectiv/objectiv-analytics/pull/402, since we need to concat calculations of multiple quantiles into a new series.

### DIFF
--- a/bach/bach/dataframe.py
+++ b/bach/bach/dataframe.py
@@ -1,4 +1,5 @@
 from copy import copy
+from functools import reduce
 from typing import List, Set, Union, Dict, Any, Optional, Tuple, cast, NamedTuple, \
     TYPE_CHECKING, Callable, Hashable, Sequence
 from uuid import UUID
@@ -2031,14 +2032,26 @@ class DataFrame:
         :param numeric_only: whether to aggregate numeric series only, or attempt all.
         :returns: a new DataFrame with the aggregation applied to all selected columns.
         """
-        return self._aggregate_func(
+        df = self
+        if df.group_by is None:
+            df = df.groupby()
+
+        new_series = df._apply_func_to_series(
             func='quantile',
             axis=axis,
-            numeric_only=numeric_only,
-            level=level,
+            numeric_only=True,
+            exclude_non_applied=True,
+            partition=df.group_by,
             q=q,
             **kwargs,
         )
+        # TODO: check all have the same index
+        series_to_df = [series.to_frame() for series in new_series]
+        df = reduce(
+            lambda left, right: left.merge(right, left_index=True, right_index=True),
+            series_to_df
+        )
+        return self
 
     def nunique(self, axis=1, skipna=True, **kwargs):
         """
--- a/bach/bach/dataframe.py
+++ b/bach/bach/dataframe.py
@@ -2296,7 +2296,7 @@ class DataFrame:
             ignore_index=ignore_index,
             sort=sort,
         )()
-        return concatenated_df if isinstance(concatenated_df, DataFrame) else concatenated_df.to_frame()
+        return concatenated_df
 
 
 def dict_name_series_equals(a: Dict[str, 'Series'], b: Dict[str, 'Series']):
--- a/bach/bach/dataframe.py
+++ b/bach/bach/dataframe.py
@@ -19,7 +19,7 @@ from sql_models.sql_generator import to_sql
 if TYPE_CHECKING:
     from bach.partitioning import Window, GroupBy
     from bach.savepoints import Savepoints
-    from bach.series import Series, SeriesBoolean, SeriesAbstractNumeric
+    from bach.series import Series, SeriesBoolean
 
 # TODO exclude from docs
 DataFrameOrSeries = Union['DataFrame', 'Series']
@@ -2020,7 +2020,6 @@ class DataFrame:
         self,
         q: Union[float, List[float]] = 0.5,
         axis=1,
-        numeric_only=True,
         **kwargs,
     ):
         """
@@ -2028,10 +2027,13 @@ class DataFrame:
 
         :param q: value or list of values between 0 and 1.
         :param axis: only ``axis=1`` is supported. This means columns are aggregated.
-        :param numeric_only: whether to aggregate numeric series only, or attempt all.
         :returns: a new DataFrame with the aggregation applied to all selected columns.
         """
+        from bach.series.series_numeric import SeriesAbstractNumeric
         df = self
+        if all(not isinstance(s, SeriesAbstractNumeric) for s in df.all_series.values()):
+            raise ValueError('Cannot calculate quantiles for dataframe with no numeric columns.')
+
         if df.group_by is None:
             df = df.groupby()
 
@@ -2042,7 +2044,7 @@ class DataFrame:
             new_series = df._apply_func_to_series(
                 func='quantile',
                 axis=axis,
-                numeric_only=numeric_only,
+                numeric_only=True,
                 exclude_non_applied=True,
                 partition=df.group_by,
                 q=qt,
--- a/bach/bach/dataframe.py
+++ b/bach/bach/dataframe.py
@@ -8,7 +8,7 @@ import pandas
 from sqlalchemy.engine import Engine
 from sqlalchemy.future import Connection
 
-from bach.expression import Expression, SingleValueExpression, VariableToken
+from bach.expression import Expression, SingleValueExpression, VariableToken, AggregateFunctionExpression
 from bach.sql_model import BachSqlModel, CurrentNodeSqlModel, get_variable_values_sql
 from bach.types import get_series_type_from_dtype, get_dtype_from_db_dtype
 from sql_models.graph_operations import update_placeholders_in_graph, get_all_placeholders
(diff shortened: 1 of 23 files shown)


bach/bach/series/series.py
```diff
@@ -1069,6 +1069,34 @@ def mode(self, partition: WrappedPartition = None, skipna: bool = True):
...
+                expression=AggregateFunctionExpression.construct(
+                    f'percentile_cont({qt}) within group (order by {{}})',
+                    self,
+                ),
+            )
+
+            quantile_df = agg_result.to_frame().materialize()
+            # maps the resultant quantile
+            quantile_df['q'] = qt
+            quantile_df = quantile_df.set_index('q')
```
KathiaBarahona, line 1090: **Context**
When calculating multiple quantiles for a series, the method should return a Series object where each row has the calculated quantile plus it's result. 
Example:
|      | result |
|:----:|:------:|
| 0.25 |    1   |
|  0.5 |    2   |
| 0.75 |    3   |

Since `quantile` is an aggregated function, setting a constant expression as its index requires multiple changes on more complex methods, therefore I decided to convert the result into a dataframe and materialize the aggregation subquery, this way we can add the constant value as an index (was the least hacky solution that I managed to do).

bach/bach/dataframe.py
```diff
@@ -2015,6 +2016,50 @@ def mode(self, axis=1, skipna=True, level=None, numeric_only=False, **kwargs):
...
+        quantile_df = quantile_df.sum() if unique_quantile else quantile_df.groupby(by='q').sum()
+
+        quantile_df.rename(
+            columns={
+                name: name.replace('_sum', '') for name in quantile_df.data_columns
+            },
+            inplace=True
+        )
+        return quantile_df
+
```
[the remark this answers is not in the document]
KathiaBarahona: Also tried using `DataFrame.merge` instead of the concat operation. But was a bit of a headache when the series have empty indexes.

bach/bach/dataframe.py
```diff
@@ -2015,6 +2016,42 @@ def mode(self, axis=1, skipna=True, level=None, numeric_only=False, **kwargs):
...
+            q=q,
+            **kwargs,
+        )
+        if isinstance(q, float) or len(q) == 1:
+            return df.copy_override(base_node=new_series[0].base_node, series={s.name: s for s in new_series})
+
+        # merge all quantiles based on 'q' index
+        return reduce(
+            lambda left, right: left.merge(right, on='q'), [series.to_frame() for series in new_series]
+        )
```
KathiaBarahona, line 2053: we should consider merging with multiple dataframes, this would  make this query significantly smaller

bach/bach/dataframe.py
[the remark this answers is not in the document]
KathiaBarahona: ignore ^ that comment

bach/bach/dataframe.py
jansentom: would be cool to support, but quite complex maybe. Could you create an issue for something like "allow merge on multiple dataframes on single "on"?

KathiaBarahona: There it goes. https://github.com/objectiv/objectiv-analytics/issues/414
Anyways, I ended up using concat operation, the query got a bit smaller

bach/bach/dataframe.py
```diff
@@ -2015,6 +2016,63 @@ def mode(self, axis=1, skipna=True, level=None, numeric_only=False, **kwargs):
...
+                return quantile_df
+
+            # a hack in order to avoid calling quantile_df.materialized().
+            # Currently doing quantile['q'] = qt
+            # will raise some errors since the expression is not an instance of AggregateFunctionExpression
+            quantile_df['q'] = initial_series.copy_override(
+                dtype='float64',
+                expression=AggregateFunctionExpression.construct(fmt=f'{qt}'),
+            )
+            quantile_df.view_sql()
```
KathiaBarahona, line 2068: whoops, removing it

bach/bach/dataframe.py
```diff
@@ -2015,6 +2016,62 @@ def mode(self, axis=1, skipna=True, level=None, numeric_only=False, **kwargs):
...
+        Returns the quantile of all numeric columns.
+
+        :param q: value or list of values between 0 and 1.
+        :param axis: only ``axis=1`` is supported. This means columns are aggregated.
+        :param numeric_only: whether to aggregate numeric series only, or attempt all.
+        :returns: a new DataFrame with the aggregation applied to all selected columns.
+        """
+        df = self
+        if df.group_by is None:
+            df = df.groupby()
```
thijs-obj, line 2036: question: is groupby needed? Can we not use `percentile_cont` in sql without grouping?

bach/bach/dataframe.py
```diff
@@ -2015,6 +2016,62 @@ def mode(self, axis=1, skipna=True, level=None, numeric_only=False, **kwargs):
...
+            new_series = df._apply_func_to_series(
+                func='quantile',
+                axis=axis,
+                numeric_only=numeric_only,
+                exclude_non_applied=True,
+                partition=df.group_by,
+                q=qt,
+                **kwargs,
+            )
+            initial_series = new_series[0]
```
thijs-obj, line 2051: question: do we not need to handle the two special cases where `len(q) == 0` and `len(self.data) == 0` ?

Or are we satisfied with raising an exception like we do now?

bach/bach/dataframe.py
```diff
@@ -2015,6 +2016,62 @@ def mode(self, axis=1, skipna=True, level=None, numeric_only=False, **kwargs):
...
+                series={s.name: s for s in new_series},
+                index={},
+                group_by=[initial_series.group_by],
+            )
+            if len(quantiles) == 1:
+                return quantile_df
+
+            # a hack in order to avoid calling quantile_df.materialized().
+            # Currently doing quantile['q'] = qt
+            # will raise some errors since the expression is not an instance of AggregateFunctionExpression
```
thijs-obj, line 2063: thanks good explanation :ok_hand:

bach/bach/dataframe.py
```diff
@@ -2015,6 +2016,62 @@ def mode(self, axis=1, skipna=True, level=None, numeric_only=False, **kwargs):
...
 
+    def quantile(
+        self,
+        q: Union[float, List[float]] = 0.5,
+        axis=1,
+        numeric_only=True,
+        **kwargs,
+    ):
+        """
+        Returns the quantile of all numeric columns.
```
thijs-obj, line 2027: it says here `all numeric columns`. So what happens if I call this with `numeric_only=False` ?

bach/docs/source/DataFrame.rst
```diff
@@ -146,6 +146,7 @@ Reshaping, indexing, sorting & merging
     DataFrame.reset_index
     DataFrame.set_index
     DataFrame.merge
+    DataFrame.append
```
thijs-obj, line 149: nice :+1:

thijs-obj (commented):
I think the generated SQL is much nicer than before. Did leave a whole bunch of questions I have about how this works

bach/bach/dataframe.py
jansentom: we need it indeed, else we can not use AggregateFunctionExpression

bach/bach/dataframe.py
jansentom: Let's add the checks indeed.

bach/bach/dataframe.py
```diff
@@ -2015,6 +2016,62 @@ def mode(self, axis=1, skipna=True, level=None, numeric_only=False, **kwargs):
...
+            )
+            if len(quantiles) == 1:
+                return quantile_df
+
+            # a hack in order to avoid calling quantile_df.materialized().
+            # Currently doing quantile['q'] = qt
+            # will raise some errors since the expression is not an instance of AggregateFunctionExpression
+            quantile_df['q'] = initial_series.copy_override(
+                dtype='float64',
+                expression=AggregateFunctionExpression.construct(fmt=f'{qt}'),
```
jansentom, line 2066: You can slightly decrease the hackiness by not using the literal `qt` but use `SeriesFloat64.supported_literal_to_expression()`

bach/bach/series/series.py
```diff
@@ -1069,6 +1069,47 @@ def mode(self, partition: WrappedPartition = None, skipna: bool = True):
...
+                return agg_result
+
+            quantile_df = agg_result.to_frame()
+            # maps the resultant quantile
+            # a hack in order to avoid calling quantile_df.materialized().
+            # Currently doing quantile['q'] = qt
+            # will raise some errors since the expression is not an instance of AggregateFunctionExpression
+            quantile_df['q'] = agg_result.copy_override(
+                dtype='float64',
+                expression=AggregateFunctionExpression.construct(fmt=f'{qt}'),
```
jansentom, line 1102: Same comment as in df.quantile()

bach/bach/series/series.py
```diff
@@ -1069,6 +1069,47 @@ def mode(self, partition: WrappedPartition = None, skipna: bool = True):
             skipna=skipna
         )
 
+    def quantile(self, partition: WrappedPartition = None, q: Union[float, List[float]] = 0.5):
+        """
+        When q is a float or len(q) == 1, the resultant series index will be empty
```
jansentom, line 1074: that's not entirely true: the resulting series will have the index matching the requested group_by as set in `partition`.

jansentom (approved):
Some very small remarks/changes, but neeeeeearly done! It's come out very nice (thanks for the clear comments around the hack)

bach/bach/dataframe.py
```diff
@@ -2015,6 +2016,50 @@ def mode(self, axis=1, skipna=True, level=None, numeric_only=False, **kwargs):
...
+            numeric_only=numeric_only,
+            exclude_non_applied=True,
+            partition=df.group_by,
+            q=q,
+            **kwargs,
+        )
+        unique_quantile = isinstance(q, float) or len(q) == 1
+
+        from bach.concat import DataFrameConcatOperation
+        quantile_df = DataFrameConcatOperation([series.to_frame() for series in new_series])()
```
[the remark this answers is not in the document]
KathiaBarahona: Yes, the current problem is that we will have a bunch of nested joins. Created https://github.com/objectiv/objectiv-analytics/issues/414 to do merge with multiple dataframes at once

bach/bach/series/series.py
```diff
@@ -1069,6 +1069,34 @@ def mode(self, partition: WrappedPartition = None, skipna: bool = True):
             skipna=skipna
         )
 
+    def quantile(self, partition: WrappedPartition = None, q: Union[float, List[float]] = 0.5):
```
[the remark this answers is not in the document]
KathiaBarahona: Will add it, but we should also define the returning type to the other agg methods.

bach/bach/dataframe.py
KathiaBarahona: @jansentom we can raise an exception for now, Pandas just returns an empty  df for this cases, it just doesn't makes sense to me to try to calculate quantiles for a df that has no numeric series.
Also, I'll remove the `numeric_only` parameter

bach/bach/dataframe.py
KathiaBarahona: will remove that parameter, we shouldn't even consider trying with non numeric columns

bach/bach/dataframe.py
KathiaBarahona: Unfortunately not, all series must have an aggregated expression when group_by is present :(.

https://github.com/objectiv/objectiv-analytics/blob/7e492605da21d85be07c63971cda8d9abeaf45fe/bach/bach/dataframe.py#L1665-L1673

bach/bach/series/series.py
KathiaBarahona: huuh true true. good catch

bach/bach/dataframe.py
jansentom: Ah yes, but that's not what I meant. I meant `expression=AggregateFunctionExpression(SeriesFloat64.supported_literal_to_expression(qt))`

jansentom (approved):
Whoop! ship it!

bach/bach/dataframe.py
KathiaBarahona: yeeah, but actually that won't work. It'll be:
``` python
AggregateFunctionExpression(
    SeriesFloat64.supported_literal_to_expression(SeriesFloat64.supported_value_to_literal(qt))
)
```

Issue #383, opened 2022-01-18
Operations on a `BachSqlModel` may return an `SqlModel` leading to errors


thijs-obj:
**Issue:** Operations on a `BachSqlModel` might return an `SqlModel`, which might lead to errors.

Examples:
1. dataframe.py has a clear TODO  [see here](https://github.com/objectiv/objectiv-analytics/blob/bd16140e8d7d25467a03e36a6da1d2c924e27855/bach/bach/dataframe.py#L1658) on fixing typing.
2. sample.py sets the base_node to a SqlModel [here](https://github.com/objectiv/objectiv-analytics/blob/bd16140e8d7d25467a03e36a6da1d2c924e27855/bach/bach/sample.py#L88)

Currently mypy doesn't complain (or we tell it not to complain), but it could lead to bugs.

**Expected behavior**
Operations on `BachSqlModel`s should return `BachSqlModel`s.

Alternative: get rid of `BachSqlModel` and move the functionality of that class to the `SqlModel` class. Or do both.

thijs-obj:
This was fixed in https://github.com/objectiv/objectiv-analytics/pull/406

PR #406, opened 2022-01-25
SqlModels return correct subclass of SqlModel from copying functions

thijs-obj:
This addresses issue https://github.com/objectiv/objectiv-analytics/issues/383

This builds on some more trivial changes in https://github.com/objectiv/objectiv-analytics/pull/407

Changes:
1) Subclasses of `SqlModel` now return their own type when calling functions like `set`, `link`, `set_materialization`, and `set_materialization_name`. Before a `BachSqlModel.set_materialization()` call would return a `SqlModel`, now it returns a `BachSqlModel`.
2) Added new function `replace_non_start_node_in_graph` that's similar to `replace_node_in_graph` but guarantees to return the same type as the `start_node`
3) Changed Subclasses of `SqlModel`: The constructor now looks similar to `SqlModel`, and instead any custom sql wrangly happens in a `get_instance()` function. Having similar constructors makes it possible to instantiate the current class from `SqlMode.copy_override()`, which was needed for the first change.
4) Added tests

Please have a look @KathiaBarahona @jansentom


bach/tests/unit/bach/test_sql_model.py
```diff
@@ -0,0 +1,28 @@
+"""
+Copyright 2022 Objectiv B.V.
+"""
+from bach.sql_model import BachSqlModel
+from sql_models.model import CustomSqlModelBuilder, Materialization
+
+
+def test_bach_sql_model_copy():
```
thijs-obj, line 8: This test fails on the current main branch

bach/bach/dataframe.py
```diff
@@ -1713,8 +1713,7 @@ def view_sql(self, limit: Union[int, slice] = None) -> str:
         """
         model = self.get_current_node('view_sql', limit=limit)
         placeholder_values = get_variable_values_sql(variable_values=self.variables)
-        # TODO: fix typing here, or move all BachSqlModel logic to SqlModel, or both?
```
thijs-obj, line 1716: Here the wrong type that `update_placeholders_in_graph` returned was not directly a problem, as we only used the returned model to generate sql. But nice to get rid of the `# type: ignore` here too

bach/bach/sample.py
```diff
@@ -58,15 +58,15 @@ def get_sample(df: DataFrame,
...
+    new_base_node = SampleSqlModel.get_instance(
         table_name=table_name,
         previous=original_node,
         columns=original_node.columns
     )
     return df.copy_override_base_node(base_node=new_base_node)
 
 
-def get_unsampled(df) -> 'DataFrame':
+def get_unsampled(df: DataFrame) -> 'DataFrame':
```
thijs-obj, line 69: In the past mypy didn't complain about the type of `updated_graph` below, because it didn't know that `df` was a `DataFrame`. With the change on this line, mypy started complaining about the old code below.

The code below changed the `base_node` from a `BachSqlModel` to a `SqlModel`

bach/sql_models/graph_operations.py
```diff
@@ -251,13 +260,46 @@ def replace_node_in_graph(start_node: SqlModel,
...
 
+def replace_non_start_node_in_graph(
+        start_node: TSqlModel,
+        reference_path: RefPath,
+        replacement_model: SqlModel) -> TSqlModel:
+    """
+    Similar to :meth:`replace_node_in_graph()`, except in this case the reference_path cannot be empty, i.e.
+    the node to replace cannot be the start_node.
+
+    The limitation that reference_path cannot be empty has the advantage that this function guarantees to
```
thijs-obj, line 271: This is a bit hackish in a way, but it's very nice that this way mypy can correctly deduce the types and we don't need to do any casts

bach/sql_models/model.py
```diff
@@ -23,6 +23,17 @@
...
 
+class NotSet(Enum):
+    """
+    INTERNAL: Special token used as default value for parameters, to distinguish the default value from
+    None for Optional values.
+    """
+    token = 0
+
+
+not_set = NotSet.token
```
thijs-obj, line 34: This is a solution to the problem that a parameter to `copy_override()` can legit be `None`. I think this is a bit cleaner solution than what we currently have in DataFrame and Series where we wrap things in a list. We should probably use this solution there too, and move this Enum class to some shared code file.

bach/sql_models/model.py
```diff
@@ -551,13 +564,15 @@ def set(self,
         :return: an updated copy of this node
         """
         # import locally to prevent cyclic imports
-        from sql_models.graph_operations import get_node, replace_node_in_graph
+        from sql_models.graph_operations import get_node, replace_non_start_node_in_graph
+        if reference_path == tuple():
```
thijs-obj, line 568: It's a bit lame that we get different code paths here for the case that `reference == tuple()` and all other cases. But it does make it obvious to both mypy and any future programmers what the return type  is.

bach/tests/functional/bach/test_df_sample.py
```diff
@@ -204,3 +204,33 @@ def test_sample_grouped():
...
+
+    all_data_bt = bt_sample.get_unsampled()
+    all_data_bt['extra_ppl'] = all_data_bt.inhabitants + 5
+
+    assert_equals_data(
+        all_data_bt,
+        expected_columns=_EXPECTED_COLUMNS_OPERATIONS,
+        expected_data=_EXPECTED_DATA_OPERATIONS
+    )
+    assert not all_data_bt.is_materialized  # this used to raise an exception
```
thijs-obj, line 234: On the current main branch this fails

bach/sql_models/graph_operations.py
```diff
@@ -2,11 +2,15 @@
...
 """
 from collections import deque
-from typing import NamedTuple, List, Dict, Set, Tuple, Optional, Callable, Deque, Any, Hashable, Mapping
+from typing import NamedTuple, List, Dict, Set, Tuple, Optional, Callable, Deque, Hashable, Mapping, TypeVar
 
 from sql_models.model import SqlModel, RefPath
 
 
+TSqlModel = TypeVar('TSqlModel', bound='SqlModel')
+T2SqlModel = TypeVar('T2SqlModel', bound='SqlModel')
```
KathiaBarahona, line 11: When is it used?

bach/sql_models/graph_operations.py
```diff
@@ -251,13 +260,46 @@ def replace_node_in_graph(start_node: SqlModel,
     )
 
 
+def replace_non_start_node_in_graph(
```
KathiaBarahona, line 263: I'm a bit confused with this function, it's almost the same as `replace_node_in_graph` the difference is that it doesn't accepts an reference path. Why not reusing  `replace_node_in_graph` ?  this way we can avoid duplicated code.

bach/sql_models/graph_operations.py
```diff
@@ -222,20 +226,25 @@ def filter_function(node: SqlModel) -> bool:
     return start_node
 
 
-def replace_node_in_graph(start_node: SqlModel,
-                          reference_path: RefPath,
-                          replacement_model: SqlModel) -> SqlModel:
+def replace_node_in_graph(
+        start_node: SqlModel,
+        reference_path: RefPath,
+        replacement_model: SqlModel) -> SqlModel:
```
KathiaBarahona, line 232: why not `TSqlModel`?

bach/sql_models/model.py
KathiaBarahona: Interesting solution :+1:

bach/sql_models/model.py
```diff
@@ -467,7 +479,33 @@ def hash(self) -> str:
...
+            self.materialization_name if materialization_name is not_set else materialization_name
+        return self.__class__(
+            model_spec=self.model_spec if model_spec is None else model_spec,
+            placeholders=self.placeholders if placeholders is None else placeholders,
+            references=self.references if references is None else references,
+            materialization=self.materialization if materialization is None else materialization,
+            materialization_name=materialization_name_value
+        )
+
+    def copy_set(self: TSqlModel, new_placeholders: Mapping[str, Any]) -> TSqlModel:
```
KathiaBarahona, line 508: why do we have this method? isn't it the same to just call `SqlModel.copy_override(placeholders=placeholders)`? as a user I intuitively will just call `copy_override`.

bach/sql_models/model.py
```diff
@@ -477,16 +515,9 @@ def copy_set(self, new_placeholders: Mapping[str, Any]) -> 'SqlModel[T]':
...
-            references=self.references,
-            placeholders=placeholders,
-            materialization=self.materialization,
-            materialization_name=self.materialization_name
-        )
+        return self.copy_override(placeholders=placeholders)
 
-    def copy_link(self,
-                  new_references: Dict[str, 'SqlModel']) -> 'SqlModel[T]':
+    def copy_link(self: TSqlModel, new_references: Dict[str, 'SqlModel']) -> TSqlModel:
```
KathiaBarahona, line 520: same as in my previous comment.

bach/sql_models/model.py
thijs-obj: Fair question. I think we could replace all usage of these functions (very little) by `copy_override` indeed.

These functions do have slightly different functionality: they allow you to specify only part of the placeholders, and only those will be updated. But that's not actually used anywhere.

So I think there are two alternatives:
1) Make _copy_override() private and have these functions be the primary interface
2) Let copy_override() be the primary interface, and remove these functions.

What do you guys think?

bach/sql_models/graph_operations.py
thijs-obj: aah good one, I forgot about this!

Changed the return type of `replace_node_in_graph` using this type. This also allows for some simplification in sample.py. See commit https://github.com/objectiv/objectiv-analytics/pull/406/commits/45d9b7741c1744473a061e40c6394a103b26e224

bach/sql_models/graph_operations.py
thijs-obj: Changed in https://github.com/objectiv/objectiv-analytics/pull/406/commits/45d9b7741c1744473a061e40c6394a103b26e224, now we use `TSqlModel` and `T2SqlModel` here

bach/sql_models/graph_operations.py
thijs-obj: Does the docstring not explain the intended goal of this function. Or do you think that is clear, but that preventing duplicate code outweighs knowing the return type?

KathiaBarahona: The docstring is clear, but that means you have to read it every time to actually understand why we are doing almost the same as `replace_node_in_graph`. This might also mean that the function name is a bit misleading.

bach/sql_models/model.py
KathiaBarahona: I prefer option 2 since it involves less changes.

thijs-obj: I think option 1 is less work tho. As we have a lot of tests for the other functions, but none for copy_override. I didn't add any since we have so many tests for the functions that call that function. So I think if we go that route we should do it as a separate issue

thijs-obj: @jansentom you have any opinions on this?

bach/sql_models/model.py
thijs-obj: Created a low priority issue to adopt this in DataFrame and Series too: https://github.com/objectiv/objectiv-analytics/issues/411

bach/sql_models/model.py
thijs-obj: Ok, 2 votes for that option so it wins. Added deprecation comments: https://github.com/objectiv/objectiv-analytics/pull/406/commits/7f1f9df85749f4d4f43270e59596e272bdb4d95f

bach/sql_models/model.py
thijs-obj: https://github.com/objectiv/objectiv-analytics/pull/406#discussion_r792490497 see discussion there

bach/sql_models/graph_operations.py
thijs-obj: As discussed out-of-band: I am open to concrete suggestions. Let me know if you think of anything

jansentom (approved):
Nice to have strict types here. I do feel that it's a bit over-engineered now. Deed is done, so good to have this, but maybe a slightly more pragmatic approach would have been okay as well :nerd_face:

bach/sql_models/graph_operations.py
```diff
@@ -251,13 +261,46 @@ def replace_node_in_graph(start_node: SqlModel,
...
+
+    The original start node, and all nodes that it refers recursively are unchanged.
+    :param start_node: start node
+    :param reference_path: references to traverse to get to the node that has to be updated. Cannot be empty
+    :param replacement_model: model instance that will replace the reference node
+    :return: an updated copy of the start node
+    :raises ValueError: if reference_path is an empty tuple
+    """
+    if reference_path == tuple():
+        raise ValueError(f'reference path cannot be empty')
```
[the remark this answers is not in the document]
thijs-obj: Changed in https://github.com/objectiv/objectiv-analytics/pull/406/commits/af16e80d15cabf5ffe482b71d06318a055015be2
to 
```python
        raise ValueError(f'reference path cannot be empty, use replace_node_in_graph() instead.')
```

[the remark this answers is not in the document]
KathiaBarahona: I can live with that

KathiaBarahona (approved):
lgtm

PR #408, opened 2022-01-25
Downgrade AngularTracker compatibility to Angular 9

sdirosa:
The aim of this PR is to lower the minimum version of Angular required by our Angular Tracker package to v9. 

This means downgrading not only Angular from 12 to 9, but also TypeScript from 4.5 to 3.8.

Since Angular Tracker is a thin layer on top of Browser Tracker, the latter has to be downgraded to TypeScript 3.8 as well.

Preparation:
- Downgraded my global angular client from 12 to 9

Changes:

1. BrowserTracker: Downgraded TypeScript to 3.8.3
2. BrowserTracker: Superstruct dependency is now incompatible with lower TypeScript version. Removed it.
3. BrowserTracker: Replaced all type guards that we had implemented via Superstruct with our own.
4. BrowserTracker: Added and adjusted several unit tests which relied on Superstruct to work without it.
5. AngularTracker: Downgraded TypeScript to 3.8.3 and Angular common, core, compiler and cli to 9.
6. AngularTracker: Small code adjustments for compatibility with the new versions of TypeScript and Angular.
7. AngularTracker: Disabled ivy compiler in tsconfig, as recommended by Angular guidelines for Angular 9.
8. Bumped BrowserTracker version to 0.0.7. Dependent AngularTracker got bumped as well to 0.0.7.


sdirosa:
New packages are live on NPM; merging this up.

Issue #401, opened 2022-01-24
Improvement: don't materialize dfs in DataFrame.append()


thijs-obj:
Generally in `merge()` we don't materialize the dataframes being merged, in append() we always materialize.

* Investigate whether we can generate shorter and more readable SQL if we eliminate materialize() in `append()`
* Do some investigation on performance of the sql on Postgres and perhaps BigQuery?
* If this is an improvement: eliminate materialize() in append() when it's possible.

PR #413, opened 2022-01-27
remove materialization from DataFrameConcatOperation

KathiaBarahona:
**Issue**: https://github.com/objectiv/objectiv-analytics/issues/401

### Description
Previously `DataFrameConcatOperation` materialized each dataframe to be concatenated after calling `_fill_missing_series`. The reason of the materialization was that the class will use the result from `_get_series`, which returns the final series of the dataframe. The issue of this was that each expression made referenced to an existent column of the subquery.

In order to remove the materialization, I decided to remove `_fill_missing_series` method and instead of this, construct a new expression for the columns that don't exist in that base_node. This considerably made smaller the query.

### Checklist
- [X] Tests
- [X] Updated docstrings


thijs-obj (commented):
Question: in `SeriesConcatOperation` there is still a call to `materialize()`. Is that still needed?

thijs-obj:
The DataFrame part looks good to me. Question about the Series part

bach/bach/concat.py
```diff
@@ -143,35 +143,19 @@ def _get_overridden_objects(self) -> List[DataFrame]:
             if self.ignore_index:
```
jansentom, line 143: `df = obj.copy_override()` can be written as `df = obj.copy()`. That will do exactly the same, but shorter.

bach/tests/functional/bach/test_df_append.py
```diff
@@ -161,3 +161,34 @@ def test_append_w_different_dtypes() -> None:
...
+def test_append_w_non_materialized_df() -> None:
+    caller_pdf = pd.DataFrame({'b': [1, 2, 3, 4], 'c': [1, 2, 3, 4]})
+    other_pdf = pd.DataFrame({'b': [1.1, 2.2, 3.3, 4.4, 5.5], 'c': [5, 6, 7, 8, 9]})
+
+    caller_df = get_from_df('caller_df', caller_pdf)
+    other_df = get_from_df('other_df', other_pdf)
+
+    caller_df['d'] = caller_df['b'] + caller_df['c']
+
+    caller_df = caller_df.reset_index(drop=True)
```
jansentom, line 175: `reset_index` is not required here. `set_index()` below already discards the old index.

jansentom (approved):
Two small optimizations, but LGTM in general. Very nice cleanup of the generated SQL!

Issue #411, opened 2022-01-26
Improvement: use `not_set` in `DataFrame.copy_override()` and `Series.copy_override()`


thijs-obj:
Use `not_set` in `DataFrame.copy_override()` and `Series.copy_override()`, similar to what we have in `SqlModel`

See discussion on recently added `SqlModel.copy_override()`: https://github.com/objectiv/objectiv-analytics/pull/406#discussion_r792060095_

PR #426, opened 2022-01-28
implement NotSet in copy_override for DataFrame and Series, more typing

KathiaBarahona:
**Issue**:  https://github.com/objectiv/objectiv-analytics/issues/411

## Description
Following @thijs-obj  implementation in https://github.com/objectiv/objectiv-analytics/pull/406#discussion_r792060095, started using NotSet enum for `DataFrame.copy_override` and `Series.copy_override`. 

## Changes
* Moved NotSet enum to `constants.py` module
* Added typing for `Series.copy_override` params. (This led to further changes)
* Removed `List` typing for `group_by` param. This removes `group_by=[None]` hack.
* Started typing parameters with default None value as `Optional`. This is because [mypy's implicit optional behavior might change in a future](https://mypy.readthedocs.io/en/stable/kinds_of_types.html#:~:text=you%20can%20use%20the%20--no-implicit-optional%20command-line%20option%20to%20stop%20treating%20arguments%20with%20a%20none%20default%20value%20as%20having%20an%20implicit%20optional%5B...%5D%20type.%20it%E2%80%99s%20possible%20that%20this%20will%20become%20the%20default%20behavior%20in%20the%20future.)


bach/bach/dataframe.py
```diff
@@ -970,7 +974,7 @@ def _index_merge(self,
         value_index_name = list(value.index.keys())[0]
         if self.index[index_name].dtype != value.index[value_index_name].dtype:
             raise ValueError('dtypes of indexes should be the same')
-        renamed_value = value.copy_override(name=key)
+        renamed_value: DataFrameOrSeries = value.copy_override(name=key)
```
KathiaBarahona, line 977: I found a bit nasty we assign a DataFrame to a variable that initially was a Series. See line 980

bach/bach/series/series.py
```diff
@@ -378,15 +383,15 @@ def copy_override(self,
         if index and index_sorting is None:
             index_sorting = []
 
-        klass = self.__class__ if dtype is None else get_series_type_from_dtype(dtype)
+        klass: type = self.__class__ if dtype is None else get_series_type_from_dtype(dtype)
```
KathiaBarahona, line 386: Just followed this [hack](https://mypy.readthedocs.io/en/stable/kinds_of_types.html#:~:text=checkers%20and%20IDEs.-,The%20type%20of%20class,assume%20the%20following%20classes%3A,-class%20User%3A%0A%20%20%20%20%23%20Defines). If there's another way to type this, please let me know

jansentom (changes requested):
Nice improvement. Some remarks in the comments. Main ones:
- sql_models now depends on Bach, that's not preferred. (you could not have known)
- some forgotten docstring stuff

bach/bach/series/series.py
```diff
@@ -355,17 +359,18 @@ def copy(self):
...
-                      expression=None,
-                      group_by: Optional[List[Optional['GroupBy']]] = None,  # List so [None] != None
-                      sorted_ascending: Optional[List[Optional[bool]]] = None,   # List so [None] != None
-                      index_sorting: Optional[List[bool]] = None
-                      ):
+    def copy_override(
+        self: T,
+        dtype: Optional[str] = None,
+        engine: Optional[Engine] = None,
+        base_node: Optional[TSqlModel] = None,
```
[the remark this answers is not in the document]
KathiaBarahona: Unfortunately, mypy started complaining since sometimes it receives a BachSqlModel :(

bach/bach/constants.py
```diff
@@ -0,0 +1,12 @@
+from enum import Enum
```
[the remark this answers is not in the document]
KathiaBarahona: moved it to sql_models package, so now the dependency should be one way :grimacing:

KathiaBarahona:
> Nice improvement. Some remarks in the comments. Main ones:
> 
> * sql_models now depends on Bach, that's not preferred. (you could not have known)
> * some forgotten docstring stuff

Thanks for the docstring stuff, my brain automatically ignores them !

bach/bach/dataframe.py
```diff
@@ -571,11 +571,8 @@ def copy_override(
 
         Create a copy of self, with the given arguments overridden
 
-        Big fat warning: group_by can legally be None, but if you want to set that,
-        set the param in a list: [None], or [someitem]. If you set None, it will be left alone.
-
         There are three special parameters: index_dtypes, series_dtypes and single_value. These are used to
-        create new index and data series iff index and/or series are not given. `single_value` determines
```
jansentom, line 578: https://www.collinsdictionary.com/dictionary/english/iff :)

jansentom (approved):
LGTM!

bach/bach/dataframe.py
KathiaBarahona: the more you know haah

bach/bach/series/series.py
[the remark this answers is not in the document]
thijs-obj: I think this should be `BachSqlModel`, as that is what the `__init__` function expects.
```suggestion
        base_node: Optional[BachSqlModel] = None,
```

bach/bach/series/series.py
thijs-obj: Not something to change in this PR, but something for another issue. I think we should have a separate function `copy_override_dtype()` and remove the `dtype` parameter from this function. Then we would always know for sure what the return type is here, and mypy would for example be able to detect that `base_node` should be a `BachSqlModel` and not just a sub-class of `SqlModel` (see comment above)

bach/bach/series/series.py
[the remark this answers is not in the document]
thijs-obj: https://github.com/objectiv/objectiv-analytics/pull/463 fixes this

PR #431, opened 2022-02-02
Kathiabarahona/cut implementation

KathiaBarahona:
**Description**
Based on [Pandas.cut](https://pandas.pydata.org/docs/reference/api/pandas.cut.html). This feature is needed for `Series.value_counts`.
Used `width_bucket` sql function to achieve this. Considered the same calculations as in [Pandas implementation](https://github.com/pandas-dev/pandas/blob/bb1f651536508cdfef8550f93ace7849b00046ee/pandas/core/reshape/tile.py#L255-L273), so adjustments over each bin range are performed.

Moved concat.py to operations package, I think we should move all "operation" inteded classes to a same scope 
___
**Checklist**
- [X] Tests
- [x] docstrings
- [X] documentation


bach/bach/operations/cut.py
```diff
@@ -0,0 +1,195 @@
...
+
+from bach import SeriesAbstractNumeric, SeriesFloat64, Series, DataFrame
+from bach.expression import Expression
+
+
+class CutOperation:
+    series: SeriesAbstractNumeric
+    bins: int
+    right: bool
+    include_empty_bins: bool
```
KathiaBarahona, line 11: I'd just realized that Pandas `Series.value_counts` w/ bins actually shows 0s for those bins that have no value in the dataframe.

jansentom:
Whoa, this is a really nice PR! I just went through it, but need a fresh head (and coffee) and will check it out tomorrow.

bach/bach/operations/cut.py
```diff
@@ -0,0 +1,195 @@
...
+        """
+        Calculates upper and lower bound for each bucket.
+        """
+        bucket_properties_df = self._calculate_bucket_properties()
+
+        min_series = Series.as_independent_subquery(bucket_properties_df[f'{self.series.name}_min'])
+        step_series = Series.as_independent_subquery(bucket_properties_df[f'step'])
+
+        # self.series might not have data for all buckets, we need to actually generate the series
+        buckets = SeriesFloat64(
```
jansentom, line 150: should this not be `SeriesInt64`?

bach/tests/functional/bach/test_series_cut.py
```diff
@@ -0,0 +1,71 @@
+from decimal import Decimal
+
+import numpy as np
+import pandas as pd
+from psycopg2._range import NumericRange
```
jansentom, line 5: Is it possible to use [pandas.Interval](https://pandas.pydata.org/docs/reference/api/pandas.Interval.html) instead? Would be more generic  vs a very specific PG specifc data type

jansentom (commented):
Really nice PR to review. I have some small/medium remarks, but nothing big. Great piece of work that nicely hides a lot of complexity in simple flows. Very good use of Bach to solve the puzzle as well. Thank you!

bach/tests/functional/bach/test_series_cut.py
KathiaBarahona: Bach actually returns NumericRange instead :disappointed:. We might need to create an issue for this.

bach/bach/operations/cut.py
KathiaBarahona: it should be. nice catch!

bach/tests/functional/bach/test_series_cut.py
```diff
@@ -0,0 +1,71 @@
...
+            [14740, bin1_not_right],
+            [33520, bin2_not_right],
+            [93485, bin4_not_right],
+        ],
+    )
+
+    inhabitants_pdf = inhabitants.to_pandas()
+
+    to_assert = [
+        (pd.cut(inhabitants_pdf, bins=bins), result_right),
```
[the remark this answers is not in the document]
KathiaBarahona: done! sorted index after calling CutOperation and sorted pandas results

bach/bach/operations/cut.py
```diff
@@ -0,0 +1,195 @@
...
+        right: bool = True,
+        include_empty_bins: bool = False,
+    ) -> None:
+        self.series = series
+        self.bins = bins
+        self.right = right
+        self.include_empty_bins = include_empty_bins
+
+    def __call__(self) -> SeriesFloat64:
+        range_df = self._calculate_bucket_ranges()
```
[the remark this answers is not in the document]
KathiaBarahona: done! now we call once  `_calculate_bucket_properties` and pass the result to `_calculate_bucket_ranges` and `_calculate_buckets`

bach/tests/functional/bach/test_series_cut.py
KathiaBarahona: https://github.com/objectiv/objectiv-analytics/issues/447

bach/tests/unit/bach/test_cut.py
```diff
@@ -0,0 +1,167 @@
+import numpy as np
```
thijs-obj, line 1: This file should be in bach/tests/functional as the tests require database access

bach/bach/series/series_numeric.py
```diff
@@ -49,6 +49,15 @@ def round(self, decimals: int = 0) -> 'SeriesAbstractNumeric':
             expression=Expression.construct(f'round(cast({{}} as numeric), {decimals})', self)
         )
 
+    def cut(self, bins: int, right: bool = True) -> 'SeriesAbstractNumeric':
+        """
+        Segments values into bins.
+        :param bins: The amount of bins to segment data into
+        :param right: If true (by default), each bin will include the rightmost edge. (e.g (x,y]).
+        """
+        from bach.operations.cut import CutOperation
```
thijs-obj, line 58: Question: what happens if there are weird values in the data? e.g. `NaN`, `infinity`, `-infinity`, `null`

bach/bach/operations/cut.py
```diff
@@ -0,0 +1,195 @@
...
+class CutOperation:
+    series: SeriesAbstractNumeric
+    bins: int
+    right: bool
+    include_empty_bins: bool
+
+    RANGE_ADJUSTMENT = 0.001  # Pandas.cut currently uses 1%
+    RANGE_SERIES_NAME = 'range'
+
+    def __init__(
```
thijs-obj, line 16: We should have some docstring somewhere that describe what the parameters to __init__ mean

bach/bach/operations/cut.py
```diff
@@ -0,0 +1,195 @@
...
+            expression=Expression.construct(
+                case_stmt,
+                compare_with,
+                *[to_adjust] * 3,
+            )
+        )
+
+    def _calculate_bucket_ranges(self, bucket_properties_df: 'DataFrame') -> 'DataFrame':
+        """
+        Calculates upper and lower bound for each bucket.
```
thijs-obj, line 144:
```suggestion
        Calculates upper and lower bound for each bucket.
        Returns a DataFrame with a row for each bucket, and four columns:
         * bucket (integer 1 to N)
         * lower_bound (float)
         * upper_bound (float)
         * range (object containing both lower_bound and upper_bound)
```
Maybe it is just me, but I think having a bit more extensive comments would make reviewing (and later editting/debugging/etc.) of this code _much_ simpler. I am having a real hard time reading this at the moment

bach/bach/operations/cut.py
```diff
@@ -0,0 +1,195 @@
...
+
+        range_df[self.RANGE_SERIES_NAME] = range_df.bucket.copy_override(
+            expression=Expression.construct(
+                # casting is needed since numrange does not support float64
+                f'numrange(cast({{}} as numeric), cast({{}} as numeric), {self.bounds})',
+                range_df.all_series['lower_bound'],
+                range_df.all_series['upper_bound'],
+            ),
+        )
+        return range_df
```
thijs-obj, line 195: if we don't use the lower_bound and upper_bound, then we could also remove those columns here. That will make the generated SQL slightly simpler. The SQL we generate is quite complex, so any improvement on that would be great imho

bach/bach/operations/cut.py
```diff
@@ -0,0 +1,195 @@
...
+        bins: int,
+        right: bool = True,
+        include_empty_bins: bool = False,
+    ) -> None:
+        self.series = series
+        self.bins = bins
+        self.right = right
+        self.include_empty_bins = include_empty_bins
+
+    def __call__(self) -> SeriesFloat64:
```
thijs-obj, line 28: We should describe in a docstring what the result of `__call__()` looks like, because one would expect a simple SeriesFloat64, but it is actually a Series with Range objects

bach/bach/series/series_numeric.py
```diff
@@ -49,6 +49,15 @@ def round(self, decimals: int = 0) -> 'SeriesAbstractNumeric':
             expression=Expression.construct(f'round(cast({{}} as numeric), {decimals})', self)
         )
 
+    def cut(self, bins: int, right: bool = True) -> 'SeriesAbstractNumeric':
+        """
+        Segments values into bins.
+        :param bins: The amount of bins to segment data into
```
thijs-obj, line 55:
```suggestion
        Segments values into bins.

        :param bins: The amount of bins to segment data into
```

without the space, this doesn't render nicely in html.

To reproduce:
```bash
cd bach/docs  # this assumes the virtual environment is setup for sphinx
make html
xdg-open bach/docs/build/html/Series/bach.SeriesAbstractNumeric.cut.html
```

bach/bach/series/series_numeric.py
KathiaBarahona: In case for `NaN` and `null` will be ignored since `width_bucket` will return a null bucket and will be ignored when [merging](https://github.com/objectiv/objectiv-analytics/blob/2d70093ec24d7754216de5068a25c0cc90c6166c/bach/bach/operations/cut.py#L36-L40) with `range_df` (which actually contains all valid buckets).  

Regards `infinity` and `-infinity` probably will raise an error when executing `width_bucket` since infinite values are not valid as `upper` and `lower` bounds. I guess that this can be solves with `dropna`. FYI, pandas raises an error when there are infinite values in the series.

bach/bach/operations/cut.py
```diff
@@ -0,0 +1,195 @@
...
+        return properties_df
+
+    def _calculate_buckets(self, bucket_properties_df: 'DataFrame') -> 'Series':
+        """
+        calculates each bucket for all series values.
+        """
+        return self.series.copy_override(
+            name='bucket',
+            expression=Expression.construct(
+                # we need to make sure max value is on the last bucket
```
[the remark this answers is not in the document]
KathiaBarahona: I would expect the same, but tested multiple times and it actually assigns the max value to `self.bins + 1`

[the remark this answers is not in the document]
thijs-obj: So am I understanding correctly that the range of a bucket in postgres is `[lower, upper)`?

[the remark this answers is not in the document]
thijs-obj: i.e. it doesn't include the upper value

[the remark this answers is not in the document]
thijs-obj: I think that is not just an issue on the upper boundary of the last bucket, but for all buckets. 

Example test:
```python
def test_cut_operation_boundary() -> None:
    bins = 3
    p_series = pd.Series(data=[1, 2, 3, 4], name='a')
    series = get_from_df('cut_df', p_series.to_frame()).a

    expected = pd.cut(p_series, bins=bins, right=True)
    result = CutOperation(series=series, bins=bins, right=True)()
    compare_boundaries(expected, result)
```

but I might be misunderstanding this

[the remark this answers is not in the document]
KathiaBarahona: huh I think, I'll just remove `_calculate_buckets`... it's a bit of a headache that we can't control the way the bounds are being controlled. I think I might [this approach](https://github.com/objectiv/objectiv-analytics/blob/5ea311056c8fc9df262281107ab3c1db54ccc4a8/bach/bach/operations/cut.py#L232-L244) as I'm doing for `qcut`, it is not as elegant as I want but it's actually more reliable since ranges are already calculated and we just need to know if the value is contained. 

But thanks, that was good catch. We can talk about it tomorrow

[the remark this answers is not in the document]
KathiaBarahona: btw, this comment might sound wrong because I could use `Series.as_independent_subquery`, but since there are a bunch of calculations for getting the ranges, this might raise `Violation` errors because multiple columns from the subquery are being returned. 
```python
        # currently is not possible to reference a column from another DataFrame
        # and use the expression in the merge subquery
```

bach/tests/functional/bach/test_series_cut.py
jansentom: Thanks!

thijs-obj (approved):

PR #437, opened 2022-02-03
add trailing slash to index

borft:
add trailing / to introduction / index page

as a bonus, remove an Internal property from sphinx, as that causes the sidebar in docusaurus to disappear

### DIFF
--- a/bach/docs/source/Series.rst
+++ b/bach/docs/source/Series.rst
@@ -142,7 +142,6 @@ Sql Model
 .. autosummary::
     :toctree: Series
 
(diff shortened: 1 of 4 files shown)


bach/docs/source/Series.rst
```diff
@@ -142,7 +142,6 @@ Sql Model
 .. autosummary::
     :toctree: Series
 
-    Series.engine
```
borft, line 145: This is an `INTERNAL` property, shouldn't be in the docs

PR #440, opened 2022-02-03
Kathiabarahona/value count implementation

KathiaBarahona:
## Description
New operation for data exploration based on DataFrame.value_counts and Series.value_counts. The purpose of this operation is to count unique rows. 

## Checklist
- [X] Tests
- [X] Add docstrings
- [X] Update documentation


bach/bach/dataframe.py
```diff
@@ -2440,6 +2440,58 @@ def drop_duplicates(
...
+        # mypy fix
+        value_count_series: SeriesInt64 = cast(
+            SeriesInt64,
+            df.all_series[f'value_counts_sum'],
+        )
+        if sort:
+            value_count_series = value_count_series.sort_values(ascending=ascending)
+
+        if normalize:
+            assert isinstance(value_count_series, SeriesAbstractNumeric)
```
KathiaBarahona, line 2490: not necessary, will remove it

bach/bach/series/series.py
```diff
@@ -1309,6 +1309,54 @@ def drop_duplicates(
...
+        # sort is not needed since final result is sorted after appending empty bins
+        value_counts_result = bins_w_values_df.value_counts(normalize=normalize, sort=False)
+
+        assert isinstance(empty_bins_df, DataFrame)
+        empty_bins_df['value_counts_sum'] = 0
+        empty_bins_df.set_index(CutOperation.RANGE_SERIES_NAME, inplace=True)
+
+        # append empty bins with count 0, final result must show those ranges
+        result = value_counts_result.append(empty_bins_df.all_series['value_counts_sum'])
+        result = result.copy_override(name='value_counts_sum')
```
thijs-obj, line 1354: this line is not needed I think. We could also keep the original name of the series perhaps like Pandas does? Or would that be inconsistent with what we do in other functions?

bach/bach/series/series.py
```diff
@@ -1309,6 +1309,54 @@ def drop_duplicates(
...
+    def value_counts(
+        self,
+        bins: Optional[int] = None,
+        normalize: bool = False,
+        sort: bool = True,
+        ascending: bool = False,
+    ) -> 'Series':
+        """
+        Returns a series containing counts per unique value
+        :param bins: works only with numeric series, groups values into the request amount of bins
```
thijs-obj, line 1321: newline missing here. same in dataframe.py

bach/bach/series/series.py
```diff
@@ -1309,6 +1309,54 @@ def drop_duplicates(
         self._update_self_from_series(result)
         return None
 
+    def value_counts(
+        self,
+        bins: Optional[int] = None,
```
thijs-obj, line 1314: The order of parameters is different than Pandas, Pandas has `bins` after `ascending` (https://pandas.pydata.org/docs/reference/api/pandas.Series.value_counts.html)

thijs-obj (commented):
Some small comments, but otherwise looks good to me.

Nice tests also :ok_hand:

bach/bach/series/series.py
KathiaBarahona: it is a bit more consistent with what we do with other functions. For example agg always add a postfix to the original name of the series. I can rename this as "__value_counts"

KathiaBarahona: also, there might be a conflict between the index and column series, since the original series is in the index of the result

KathiaBarahona: aaand now I remember why  ` result = result.copy_override(name='value_counts_sum')`, it is because `append` joins the names of all series

bach/bach/series/series.py
KathiaBarahona: good catch!

bach/bach/series/series.py
thijs-obj: Ah yes, you are right. I didn't consider that append changes the name.

The reason  `agg` changes the name tho is that you can specify multiple functions. e.g. both 'min' and 'max. So one column can be turned into multiple columns, so they have to be renamed to be unique. But that's not really an issue here, or in `append()` I think.

I personally like the 'value_counts' name better than '__value_counts', but if you feel otherwise that's fine to me.

thijs-obj (approved):
non-blocker comment about renaming. Good to merge I think :+1:

PR #443, opened 2022-02-04
Tracker: new bundler tweaks

sdirosa:
Since we want to support Angular 9, I've enabled --legacy-output for Tsup. 

This changes the output from `./dist/index.mjs` to the old format `./dist/esm/index.js` and should resolve module loading issues for older bundlers and frameworks.

This PR contains also a brand new Plugin to generate HttpContexts automatically when the Tracker initializes. These contexts contain `referrer` and `user_agent` info and will be enriched further by the Collector with `remote_address`.

### DIFF
--- a/tracker/core/schema/package.json
+++ b/tracker/core/schema/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@objectiv/schema",
-  "version": "0.0.8",
+  "version": "0.0.9",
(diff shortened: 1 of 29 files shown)


sdirosa:
@ivarpruijn I believe this should fix the issue you encountered while testing. Feel free to try it out (via Verdaccio).  Else I'll publish it later today after a bit more QA love.

PR #444, opened 2022-02-04
Bach set pandas series

hendrik-obj:
Sets a pandas series as a column by first loading it as a bach DataFrame with `materialization='cte'`


bach/tests/functional/bach/test_df_setitem.py
```diff
@@ -376,3 +376,22 @@ def test_set_series_single_value():
             [3, 3055, 3, 93485, 93485, 93490, 93488, 93491, 93492]
         ]
     )
+
+
+def test_set_pandas_series():
```
thijs-obj, line 381: I think there are a lot of cases that will work, but should be tested: what if the indexes don't match, what if the series has less rows, what if there are `NaN` values in the data?

thijs-obj:
I like the simple approach, but I think the testing is a bit limited

bach/tests/functional/bach/test_df_setitem.py
hendrik-obj: Added a bunch of tests. Regarding the NaN question: I created this issue, think it is not related to this feature. https://github.com/objectiv/objectiv-analytics/issues/458

thijs-obj (approved):
lgtm

PR #445, opened 2022-02-04
dropna implementation

KathiaBarahona:
## Description
Implemented `DataFrame.dropna` and `Series.dropna`. Both of them drop rows that contain nullable values. Supports same parameters as in Pandas (except for axis=1, as we will have to execute the query and determine if we should drop the column or not)

## Checklist
- [X] Tests
- [X] Docstring
- [x] Docs


bach/bach/series/series.py
```diff
@@ -1309,6 +1309,23 @@ def drop_duplicates(
...
+        :return: a new series with dropped rows if inplace = False, otherwise None.
+        """
+        df = self.to_frame().dropna()
+        assert isinstance(df, DataFrame)
+        result_series = df.all_series[self.name]
+
+        if not inplace:
+            return result_series
+
+        self._update_self_from_series(result_series)
```
thijs-obj, line 1326: This directly contradicts the comments on line 59 to 62 of this file: https://github.com/objectiv/objectiv-analytics/blob/f9ab4c43fb76c803fed649005cf9ce396be36e8c/bach/bach/series/series.py#L59-L62

I am not sure this is a good idea. I realize that pandas allows this, but this might affect a lot of things :thinking: What if this Series is part of a DataFrame

thijs-obj, line 1326: This directly contradicts the comments on line 59 to 62 of this file: https://github.com/objectiv/objectiv-analytics/blob/f9ab4c43fb76c803fed649005cf9ce396be36e8c/bach/bach/series/series.py#L59-L62

I am not sure this is a good idea. I realize that pandas allows this, but this might affect a lot of things :thinking: What if this Series is part of a DataFrame

bach/bach/dataframe.py
```diff
@@ -2439,6 +2434,68 @@ def drop_duplicates(
...
+        :return: a new dataframe with dropped rows if inplace = False, otherwise None.
+        """
+        if axis:
+            raise ValueError('only axis = 0 is supported.')
+
+        if how not in ('any', 'all'):
+            raise ValueError(f'{how} is not a valid value for "how" parameter.')
+
+        subset = self._get_parsed_subset_of_data_columns(subset)
+        dropna_series = subset or self.data_columns
```
thijs-obj, line 2465: This seems more in line with the pandas behaviour
```suggestion
        dropna_series = self.data_columns if subset is None else subset
```

bach/bach/dataframe.py
```diff
@@ -2439,6 +2434,68 @@ def drop_duplicates(
         self._update_self_from_df(df)
         return None
 
+    def dropna(
+        self,
```
thijs-obj, line 2438:
```suggestion
        self,
        *,
```
this forces our users to explicitly write `dropna(axis=1)` instead of `dropna(1)`. The latter will give a warning with Pandas about that usage being deprecated. I think we don't need the warning and can just directly force the 'right' behaviour. Will save us a lot of deprecation messages if we want to remain compatible in the future

bach/bach/dataframe.py
```diff
@@ -2439,6 +2434,68 @@ def drop_duplicates(
...
+        axis: int = 0,
+        how: str = 'any',
+        thresh: Optional[int] = None,
+        subset: Optional[Union[str, Sequence[str]]] = None,
+        inplace: bool = False
+    ) -> Optional['DataFrame']:
+        """
+        Removes rows with missing values.
```
thijs-obj, line 2446: I think we should explain what are missing values. `NaN` ? or only sql `NULL`/pandas `None`?

thijs-obj (commented):
Left some comments

bach/bach/series/series.py
KathiaBarahona: yup, I actually mentioned about the `inplace=True` in another PR https://github.com/objectiv/objectiv-analytics/pull/428#discussion_r796762170 .

bach/bach/dataframe.py
thijs-obj: `if subset = []` then pandas checks 0 columns, we check all columns.

```python
pdf = pd.DataFrame(
    data={
        'a': [1, 2, 3, 4, 5, 6, np.nan, 7., 0, 0],
        'b': [11.9, 32.0, 43.123, 64.425124, 25.00000000001, 6, np.nan, 78., 0, 0],
        'c': [0, None, 10, 11, 11, 11, 11, 11, 11, 15],
        'd': ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j'],
        'e': [True, True, True, True, True, True, True, True, True, False],
    },
)
df = DataFrame.from_pandas(engine=engine, df=pdf.reset_index(drop=False), convert_objects=True, materialization='table', if_exists='replace', name='name of the test table')
df = df.reset_index(drop=True)
df = df.drop(columns=['index'])
print(len(pdf.dropna(subset=[])))  # 10
print(len(df.dropna(subset=[]).to_pandas()))  # 8
```

thijs-obj: I think raising an exception is a worse case: then our end user has to handle a special case. 

I can imagine that if an end users writes some function that iterates over some values for a parameter, then they don't want to handle special cases where that parameter is an empty list; much nicer if the behaviour stays consistent in that case.

KathiaBarahona: well, in that case we should just return a copy of a df? I find this behavior a bit weird/non-intuitive imho, guess that we need to specify that in the docs

thijs-obj: I guess just a copy would be good. Let's discuss tomorrow in person if you find this non-intuitive, as I feel it's the other way around :)

bach/bach/series/series.py
KathiaBarahona: @thijs-obj should we also consider removing it from DataFrame? In that case, we should just create a new issue for removing that parameter from everywhere

thijs-obj: For DataFrame, I'd say let's do what is the least amount of work. If we already have it working well and well tested we might as well keep it now I think

thijs-obj (commented):
two more suggestion. otherwise good to merge imho

PR #461, opened 2022-02-08
Kathiabarahona/qcut implementation

KathiaBarahona:
Implemented https://pandas.pydata.org/docs/reference/api/pandas.qcut.html.

Decided not to inherit from `CutOperation`, both have similar steps but  `bins != quantiles` . 

Related to the adjustment done to the lower bound, I tried to avoid it but this causes problems since calculated quantiles might be in the dataset, so a value may end up being on multiple ranges.


bach/tests/functional/bach/test_cut.py
```diff
@@ -145,3 +147,46 @@ def test_cut_calculate_bucket_ranges() -> None:
         pd.Series([pd.Interval(0.993, 3.333), pd.Interval(3.333, 5.667), pd.Interval(5.667, 8)]),
         result.sort_values(),
     )
+
+
+def test_qcut_operation() -> None:
```
jansentom, line 152: these are all nice tests!

jansentom (commented):
Nice stuff. I did not fully test / compare with pandas, but tests are nice, so I believe it works. It looks clean, and I don't have any major comments except: 
... Don't forget to add `qcut()` to the docs as well.

bach/bach/operations/cut.py
```diff
@@ -222,5 +224,118 @@ def _calculate_bucket_ranges(self, bucket_properties_df: 'DataFrame') -> 'Series
...
+        """
+        df = self.series.to_frame()
+        df.reset_index(drop=True, inplace=True)
+
+        if len(self.quantiles) == 1:
+            # need at least 2 quantiles for a range
+            df[self.RANGE_SERIES_NAME] = None
+        else:
+            quantile_ranges_df = self._get_quantile_ranges()
+            # currently is not possible to reference a column from another DataFrame
```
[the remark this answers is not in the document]
KathiaBarahona: https://github.com/objectiv/objectiv-analytics/issues/481

bach/tests/functional/bach/test_series_numeric.py
```diff
@@ -139,3 +141,129 @@ def test_aggregations_quantile():
...
+            np.testing.assert_almost_equal(exp.left, float(res.lower), decimal=2)
+            np.testing.assert_almost_equal(exp.right, float(res.upper), decimal=2)
+
+
+def test_series_qcut() -> None:
+    bounds = '(]'
+    inhabitants = get_bt_with_test_data(full_data_set=True)['inhabitants']
+
+    result = inhabitants.qcut(q=4).sort_index()
+    bin1 = NumericRange(Decimal('699.999'),  Decimal('2007.5'), bounds=bounds)
```
thijs-obj, line 215: Wouldn't it be easier to generate the following buckets, so we don't need any adjustments?
```python
    bin1 = NumericRange(Decimal('700.0'),  Decimal('2007.5'), bounds='[]')
    bin2 = NumericRange(Decimal('2007.5'),  Decimal('10120'), bounds='(]')
    bin3 = NumericRange(Decimal('10120'),  Decimal('13750'), bounds='(]')
    bin4 = NumericRange(Decimal('13750'), Decimal('93485'), bounds='(]')
```

I know I bugged you about this before, but I still don't quite get why we don't do it like this

KathiaBarahona: I tried, but the code get messier and actually less readable. IMHO in a nutshell it is just another way "to hack" the part of including the lowest value, for this I rather stick on having a similar result to the one of Pandas. Also it's a bit odd to have only the first range with those bounds.

bach/bach/operations/cut.py
```diff
@@ -222,5 +224,118 @@ def _calculate_bucket_ranges(self, bucket_properties_df: 'DataFrame') -> 'Series
...
+
+    def __init__(self, series: SeriesAbstractNumeric, q: Union[int, List[float]]) -> None:
+        self.series = series
+        self.quantiles = q if isinstance(q, list) else numpy.linspace(0, 1, q + 1).tolist()
+
+    def __call__(self, *args, **kwargs) -> SeriesFloat64:
+        """
+        Gets the quantile range per bucket and assigns the correct range to each value. If the value
+        is not contained in any range, then it will be null.
+        on it, a
```
thijs-obj, line 257: random words :)

bach/bach/operations/cut.py
```diff
@@ -222,5 +224,118 @@ def _calculate_bucket_ranges(self, bucket_properties_df: 'DataFrame') -> 'Series
...
+        """
+        df = self.series.to_frame()
+        df.reset_index(drop=True, inplace=True)
+
+        if len(self.quantiles) == 1:
+            # need at least 2 quantiles for a range
+            df[self.RANGE_SERIES_NAME] = None
+        else:
+            quantile_ranges_df = self._get_quantile_ranges()
```
thijs-obj, line 266:
```suggestion
            quantile_ranges = self._get_quantile_ranges()
```
the `_df` suffix is a bit misleading as this is a series, not a dataframe

bach/bach/series/series_numeric.py
```diff
@@ -60,6 +60,15 @@ def cut(self, bins: int, right: bool = True) -> 'SeriesAbstractNumeric':
         from bach.operations.cut import CutOperation
         return CutOperation(series=self, bins=bins, right=right)()
 
+    def qcut(self, q: Union[int, List[float]]) -> 'SeriesAbstractNumeric':
```
thijs-obj, line 63: I think the docstring should explain the return value a bit.

The values from the series become the index, and the values are range objects. That is not obvious looking at this signature and docstring

thijs-obj (commented):
One comment about the ranges and adjustment, and some small stuff.

bach/bach/operations/cut.py
KathiaBarahona: whoopps, thanks!

bach/bach/operations/cut.py
```diff
@@ -222,5 +224,117 @@ def _calculate_bucket_ranges(self, bucket_properties_df: 'DataFrame') -> 'Series
...
+        # so the value can generate duplicates
+        bound = "'(]'"
+        range_stmt = (
+            f'case when {{}} is not null\n'
+            f'then numrange(cast({{}} as numeric), cast({{}} as numeric), {bound}) end'
+        )
+        lower_bound = quantile_ranges_df.all_series['lower_bound']
+        upper_bound = quantile_ranges_df.all_series['upper_bound']
+
+        quantile_ranges_df[self.RANGE_SERIES_NAME] = lower_bound.copy_override(
```
thijs-obj, line 335: Can we materialize first?

Now the calculation for lower_bound and upper_bound are repeated multiple times in the generated SQL. I find it nearly impossible to understand the SQL

Removing the "lower_bound" and "upper_bound" columns might also be an alternative, as that removes some duplication, but haven't experimented much witht his

bach/bach/operations/cut.py
```diff
@@ -222,5 +224,117 @@ def _calculate_bucket_ranges(self, bucket_properties_df: 'DataFrame') -> 'Series
...
+            name='upper_bound'
+        )
+
+        # must not include lower bound, since a calculated quantile might also be in the dataset
+        # so the value can generate duplicates
+        bound = "'(]'"
+        range_stmt = (
+            f'case when {{}} is not null\n'
+            f'then numrange(cast({{}} as numeric), cast({{}} as numeric), {bound}) end'
+        )
```
thijs-obj, line 331:
```suggestion
        )
        quantile_ranges_df.materialize(inplace=True)
```

Checked this on one test, saved about 600 bytes of generated sql. But more importantly made the sql much more readable

bach/bach/operations/cut.py
thijs-obj: See comment above

bach/tests/functional/bach/test_series_numeric.py
thijs-obj: > I tried, but the code get messier and actually less readable

that's the part I didn't get, but that's fine.

> IMHO in a nutshell it is just another way "to hack" the part of including the lowest value

This I don't really agree with. I think changing the range values instead of being exact and setting the right inclusive/exclusive parameters seems hackish and less correct to me. There are also some extreme edge cases where it doesn't work if `_RANGE_ADJUSTMENT` is too big (example at the bottom)

Main reason I asked this, is that I looked at the generated sql, and looked at how I would write that sql by hand. And it seemed to me that the generated code was much more complex.

**Conclusion**
Having said all that. I am fine with merging this if we lower the range adjustment, if others agree. We can always look at it later if we do want to change things. 


**Example** 
Example where range adjustment is too big relative to the numbers involved
```python
    p_series = pd.Series(
        data=[1.000001, 1.000001, 1.000002, 1.000003, 1.000004, 1.000005, 1.000006, 1.0000007, 1.000008],
        name='a')
    series = get_from_df('qcut_df', p_series.to_frame()).a

    expected_w_list = pd.qcut(p_series, q=[0.25, 0.3, 0.7, 0.9])
    result_w_list = QCutOperation(series=series, q=[0.25, 0.3, 0.7, 0.9])()
    print()
    print(expected_w_list)
    print()
    print(result_w_list.to_pandas())  # shows results in wrong buckets
    compare_boundaries(expected_w_list, result_w_list)  # fails
```

PR #462, opened 2022-02-09
set private name to series results from operations

KathiaBarahona:
In order to avoid collisions between existing column names and since we don't support empty series names, added a prefix to the names of the result series of any Operation.


bach/bach/concat.py
```diff
@@ -283,7 +286,7 @@ def _get_series(self) -> Dict[str, ResultSeries]:
             dtypes.add(series.dtype)
 
         main_series = self.objects[0].copy_override(
-            name='_'.join(all_names),
+            name=self.CONCAT_SERIES_NAME,
```
KathiaBarahona, line 289: the problem of joining all series names is that sometimes the column name becomes way too long and this generates errors

thijs-obj: I think it would make more sense to take the name of the first series, i.e. leave it unchanged.

In the case that all series have the same name, which I think would be a major use-case, that would be the most logical thing to do I'd think. And in other cases it's still predictable and sort of logical too

bach/bach/series/series_numeric.py
```diff
@@ -157,11 +157,11 @@ def quantile(
             # a hack in order to avoid calling quantile_df.materialized().
             # Currently doing quantile['q'] = qt
             # will raise some errors since the expression is not an instance of AggregateFunctionExpression
-            quantile_df['q'] = agg_result.copy_override(
+            quantile_df['__q'] = agg_result.copy_override(
```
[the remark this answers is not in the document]
KathiaBarahona: yes, I know. The main issue is that Pandas normally returns a new series with an empty name. I don't feel comfortable having the possibility of returning a name that is most likely to exist as a column.

thijs-obj:
I think it makes more sense to keep the name of the first series as the name for the result of the concat operation. But my opinion is not that important :) better to ask @hendrik-obj  for review I think

thijs-obj:
@hendrik-obj do you have any opinions on this?

PR #463, opened 2022-02-09
Small type fix: series.copy_override parameter

thijs-obj:


.github/workflows/bach-tests.yml
```diff
@@ -56,7 +56,7 @@ jobs:
       - name: Typecheck with mypy
         run: |
           cd bach
-          mypy bach sql_models
+          mypy --check-untyped-defs bach sql_models
```
thijs-obj, line 59: If a function has zero annotations, it is by default ignored by mypy. Even tho mypy might know something about the types (e.g. it can infer `self`) and a function might call other functions that take/give certain types.

This change makes sure mypy doesn't ignore functions without annotations

bach/bach/series/series.py
```diff
@@ -380,7 +378,7 @@ def copy_override(
         self: T,
         dtype: Optional[str] = None,
         engine: Optional[Engine] = None,
-        base_node: Optional[TSqlModel] = None,
+        base_node: Optional[BachSqlModel] = None,
```
thijs-obj, line 381: Below we call `__init__`, which requires a `BachSqlModel`

bach/bach/partitioning.py
```diff
@@ -175,8 +180,14 @@ class GroupingSet(GroupingList):
     Abstraction to support SQLs
     GROUP BY GROUPING SETS ((colA,colB),(ColA),(ColC))
     """
-    def get_group_by_column_expression(self):
-        grouping_expr_list = [g.get_group_by_column_expression() for g in self._grouping_list]
+    def get_group_by_column_expression(self) -> Optional[Expression]:
+        grouping_optional_expr_list = [g.get_group_by_column_expression() for g in self._grouping_list]
+
+        # ensure that there are no None expressions in the list. If there are that's a bug, and we are
```
thijs-obj, line 186: mypy complained about the `Expression.construct` call below. Previously this function wasn't checked. 

This shouldn't break if it wasn't already broken, but I am not 100% sure whether the assert will always hold

bach/bach/partitioning.py
```diff
@@ -163,9 +163,14 @@ def __init__(self, grouping_list: List[GroupBy]):
         super().__init__(group_by_columns=list(group_by_columns.values()))
 
     def get_group_by_column_expression(self) -> Optional[Expression]:
-        # help mypy, our parent always returns an Expression
```
thijs-obj, line 166: this comment didn't make sense, as the expressions were not coming from the parent class.

.github/workflows/bach-tests.yml
KathiaBarahona: nice!

bach/bach/partitioning.py
```diff
@@ -163,9 +163,14 @@ def __init__(self, grouping_list: List[GroupBy]):
...
-        grouping_expr_list = [cast(Expression, g.get_group_by_column_expression())
-                              for g in self._grouping_list]
+        grouping_optional_expr_list = [cast(Expression, g.get_group_by_column_expression())
+                                       for g in self._grouping_list]
+
+        # ensure that there are no None expressions in the list. If there are that's a bug, and we are
+        # okay with breaking here. Without the cast mypy will complain as construct doesn't accept Nones
+        assert all(expr is not None for expr in grouping_optional_expr_list)
```
KathiaBarahona, line 171: shouldn't we raise an actual exception here?

bach/bach/partitioning.py
KathiaBarahona: Why is it typed as `Optional`? and also in [`GroupBy.get_group_by_column_expression`](https://github.com/objectiv/objectiv-analytics/blob/b7615e4702a4e86b4c1d3b76129161bf67174e50/bach/bach/partitioning.py#L93-L99)? If I'm not wrong, that function will always return an Expression.

KathiaBarahona: anyways, I think it's better to do smth like this
```suggestion
    def get_group_by_column_expression(self) -> Expression:
        grouping_expr_list = []
        for g in self._grouping_list:
            expr = g.get_group_by_column_expression()
            if expr:
                grouping_expr_list.append(expr)

        if not grouping_expr_list:
            raise Exception('no valid group by expressions.')

        fmtstr = ', '.join(["{}"] * len(grouping_expr_list))
        return Expression.construct(fmtstr, *grouping_expr_list)
```

bach/bach/series/series.py
```diff
@@ -380,7 +378,7 @@ def copy_override(
         self: T,
```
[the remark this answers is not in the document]
KathiaBarahona: :crying_cat_face:

bach/bach/partitioning.py
thijs-obj: Problem I had is that I don't really understand this code, and reasoning through it all would take a lot of time, so I figured I create a PR, and hopefully one of you would know if we can get `None` here and if that happens what to do. And if nobody does we can decide to either invest time or keep the code just as broken/well working as it was.

I think the new code breaks in exactly the same cases that the old code did, so it's no worse or better. Just mypy is happy now. Although you are probably right that we should raise a real excception, and not rely on `assert` as that is sometimes disabled.

Regarding your solution: I am not sure if you solution is correct, maybe if we do get a `None` here that is a bug, and if we just ignore it we might get wrong data? Again: I don't really understand this code, so you might well be right, just wondering whether you know what this does.

thijs-obj: @jansentom perhaps you have some insights here too?

jansentom: I'd be happy with raising an exception, this should not happen.

jansentom: This code is not very well tested and it's been a while since I wrote it. I suggest we just throw the exception and leave it as is.

bach/bach/partitioning.py
thijs-obj: Now raises a regular Exception: https://github.com/objectiv/objectiv-analytics/pull/463/commits/e5b7aadad2caba196593e8cf2b115a44a61d5716

bach/bach/partitioning.py
thijs-obj: Ok, now it raises a regular Exception in the case of a None

thijs-obj:
please let me know if there are any more comments @KathiaBarahona @jansentom

bach/bach/partitioning.py
```diff
@@ -163,9 +163,15 @@ def __init__(self, grouping_list: List[GroupBy]):
         super().__init__(group_by_columns=list(group_by_columns.values()))
 
     def get_group_by_column_expression(self) -> Optional[Expression]:
-        # help mypy, our parent always returns an Expression
-        grouping_expr_list = [cast(Expression, g.get_group_by_column_expression())
-                              for g in self._grouping_list]
+        grouping_optional_expr_list = [cast(Expression, g.get_group_by_column_expression())
```
KathiaBarahona, line 166: is the cast necessary? There's a similar list comprehension in `GroupingSet.get_group_by_column_expression` and there's no casting there. Also a bit misleading since it might be `None`

KathiaBarahona (approved):
lgtm, just a small comment

bach/bach/partitioning.py
thijs-obj: you are right, this is not needed. I didn't pay attention. Fixed in https://github.com/objectiv/objectiv-analytics/pull/463/commits/9cc1b598e4bd45dbdc81abbb5e7f66afe493c411

PR #464, opened 2022-02-09
Add more typing to Series

thijs-obj:
* Added typing in a lot of places
* Removed `dtype` as parameter from `copy_override()`
  * Now the return type of `copy_override()` is actually `T`
  * Added new function `copy_override_type()` for the special case where we want to change the type

### DIFF
--- a/bach/bach/concat.py
+++ b/bach/bach/concat.py
@@ -282,10 +282,9 @@ class SeriesConcatOperation(ConcatOperation[Series]):
             all_names.append(series.name)
             dtypes.add(series.dtype)
 
-        main_series = self.objects[0].copy_override(
-            name='_'.join(all_names),
-            dtype=_get_merged_series_dtype(dtypes),
-        )
+        main_series = self.objects[0]\
+            .copy_override_dtype(dtype=_get_merged_series_dtype(dtypes))\
+            .copy_override(name='_'.join(all_names))
         return self._get_result_series([main_series])
 
     def _join_series_expressions(self, obj: Series) -> Expression:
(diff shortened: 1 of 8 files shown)


KathiaBarahona:
mm not really a fan of `copy_override_type`, feels a bit hacky and with `cast('SeriesString', str_series.copy_override(experession=...))` should be enough, we still do that for other scenarios (e.g  `DataFrameorSeries`). So, I prefer sticking to a pattern for all the code.

jansentom:
I have to agree with @KathiaBarahona here. Feels like a lot of extra complexity and code-paths to remember just to get the type right, while it was not very wrong to begin with. Happy to be convinced otherwise on Monday though :)

thijs-obj:
Alright, I made an alternative fix: https://github.com/objectiv/objectiv-analytics/pull/489 Please have a look @jansentom @KathiaBarahona 

> Feels like a lot of extra complexity

I don't quite understand this comment tho. In my opinion the fix in this PR is less complex. Return types are predictable in more cases, and functions do less. The alternative (#489) is easier, but not simpler. IMHO

thijs-obj:
Discussed with Tom today. Decided to go with this one. 

Did some minor changes afterwards to fix errors after merging main back into this branch: https://github.com/objectiv/objectiv-analytics/pull/464/commits/8434444234886294d5bc84158ea73b72631dbcb2 https://github.com/objectiv/objectiv-analytics/pull/464/commits/17c65bf0d21741f04ee2915b933abaa1257c1cfb Let me know if there are any comments on that

PR #466, opened 2022-02-10
Browser/Angular Trackers: tweaks and fixes from feedback

sdirosa:
- Patch versions to 0.0.12
- Angular Tracker now supports `[applyTaggingAttributes]` in templates; spread a model-created tagger onto an element
- Browser and Angular Tracker: `ExpandableContext` does not automatically track `PressEvents` anymore, same as React
- Browser and Angular Tracker: LocationTaggers `trackVisibility` option now supports `true | false` shorthands
- Browser and Angular Tracker: Uniqueness validation does not trigger for Elements that have been removed from DOM
- Core and React Tracker: Lifted `makeIdFromString` helper function from React to Core Tracker
- Angular Tracker: Added `makeIdFromString` PipeTransform for normalizing Context identifiers directly in templates

For later:
- Add getParent API to ease further assigning a parent element to portaled components


sdirosa:
Merging this up. The `getParent` API idea will be implemented in a separate PR, as it involves also changing Core and React trackers.

Issue #430, opened 2022-02-02
Bug Report:  Min and Max failing when DataFrame has boolean series


KathiaBarahona:
Currently bach is allowing `min` and `max` aggregations for boolean series, the problem is when executing the sql, since this is functions are not allowed for boolean columns.

thijs-obj:
Make sure to also fix this TODO when that is still relevant: https://github.com/objectiv/objectiv-analytics/pull/374/files#diff-d4893a7bb88c9023e71f83191c81d74803743613b06993bc32baef020420963eR24

thijs-obj:
done, see linked PRs

PR #468, opened 2022-02-10
Bach boolean min max

thijs-obj:
* Support for min() and max() on boolean series
  * min() gives False if there is at least one False in the column
  * max() gives True if there is at least one True in the column
* Enable boolean columns in describe()

This addresses part of issue https://github.com/objectiv/objectiv-analytics/issues/430

### DIFF
--- a/bach/bach/describe.py
+++ b/bach/bach/describe.py
@@ -5,7 +5,7 @@ from typing import Optional, Union, Sequence, List, Type, Tuple
 
 from bach import (
     DataFrame, SeriesString, SeriesAbstractNumeric, DataFrameOrSeries,
-    get_series_type_from_dtype, SeriesAbstractDateTime,
+    get_series_type_from_dtype, SeriesAbstractDateTime, SeriesBoolean,
 )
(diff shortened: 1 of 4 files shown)


bach/bach/describe.py
```diff
@@ -21,9 +21,8 @@ class SupportedStats(Enum):
     MODE = 'mode'
 
 
-# TODO: Remove this constant when boolean aggregations bug (min, max) is fixed.
```
thijs-obj, line 24: I don't quite understand this TODO. We still need this, or am I misunderstanding?

For now I just added SeriesBoolean to the list.

KathiaBarahona: actually we should stop using `_SUPPORTED_SERIES_TYPES`

thijs-obj: discussed out of band. I will do some more work on this

thijs-obj: Follow up PR: https://github.com/objectiv/objectiv-analytics/pull/476

KathiaBarahona (approved):
LGTM!

Issue #467, opened 2022-02-10
Bug: Series.drop_duplicates should not support inplace=True


thijs-obj:
**Describe the bug**
`Series.drop_duplicates` has an `inplace` parameter, that mutates the Series object. In our current architecture we cannot support that , as we assume that Series are (mostly) immutable.

**Why this is a problem**

``` python
# Setup
from bach import DataFrame
from tests.functional.bach.test_data_and_utils import DB_TEST_URL
import sqlalchemy
import pandas as pd
import numpy as np

engine = sqlalchemy.create_engine(DB_TEST_URL)
pdf = pd.DataFrame({'a': [1, 2, 3, 4, 4, 4], 'b': [1, 2, 3, 4, 4, 5]})
df = DataFrame.from_pandas(engine=engine, df=pdf.reset_index(drop=False), convert_objects=True, materialization='table', if_exists='replace', name='name of the test table')
df = df.reset_index(drop=True)
df = df.drop(columns=['index'])

# Actual example
assert df.base_node == df.a.base_node # OK
df.a.drop_duplicates(inplace=True)
assert df.base_node == df.a.base_node # FAILS
```

Having different base_nodes leads to problems later on:
```python
# Setup
from bach import DataFrame
from tests.functional.bach.test_data_and_utils import DB_TEST_URL
import sqlalchemy
import pandas as pd
import numpy as np

engine = sqlalchemy.create_engine(DB_TEST_URL)
pdf = pd.DataFrame({'a': [1, 2, 3, 4, 4, 4], 'b': [1, 2, 3, 4, 4, 5]})
df = DataFrame.from_pandas(engine=engine, df=pdf.reset_index(drop=False), convert_objects=True, materialization='table', if_exists='replace', name='name of the test table')
df = df.reset_index(drop=True)
df = df.drop(columns=['index'])

# Actual example
df['x'] = (df.a == df.b)  # OK
df.a.drop_duplicates(inplace=True)
df['x'] = (df.a == df.b)  # FAILS
```

**Expected behavior**
* don't support the inplace parameter
* additional: add check to DataFrame.__init__() that all series.base_nodes match the df.base_node

PR #469, opened 2022-02-10
Bach series no mutations

thijs-obj:
* Removes the `inplace` parameter for `Series.drop_duplicates()`
  * No deprecation schedule or anything. We only just introduced it, we still have very few users, it's broken.
* Added small check in `DataFrame.__init__()` to check that base_nodes of all series match the DataFrame's

This resolves issue https://github.com/objectiv/objectiv-analytics/issues/467

**Background**
Using `Series.drop_duplicates(inplace=True)` corrupts the state of any `DataFrame` that the series is part of: The `DataFrame`'s `base_node` will be different from the `Series`'s `base_node`. This might trigger problems later on.
Example (from https://github.com/objectiv/objectiv-analytics/issues/467):
```python
# Setup
from bach import DataFrame
from tests.functional.bach.test_data_and_utils import DB_TEST_URL
import sqlalchemy
import pandas as pd
import numpy as np

engine = sqlalchemy.create_engine(DB_TEST_URL)
pdf = pd.DataFrame({'a': [1, 2, 3, 4, 4, 4], 'b': [1, 2, 3, 4, 4, 5]})
df = DataFrame.from_pandas(engine=engine, df=pdf.reset_index(drop=False), convert_objects=True, materialization='table', if_exists='replace', name='name of the test table')
df = df.reset_index(drop=True)
df = df.drop(columns=['index'])

# Actual example
df['x'] = (df.a == df.b)  # OK
df.a.drop_duplicates(inplace=True)
df['x'] = (df.a == df.b)  # FAILS
```

### DIFF
--- a/bach/bach/dataframe.py
+++ b/bach/bach/dataframe.py
@@ -196,7 +196,11 @@ class DataFrame:
                                  f'df: {value.index}, series.index: {index}')
             if value.group_by != group_by:
                 raise ValueError(f'Group_by in `series` should match dataframe. '
-                                 f'df: {value.group_by}, series.index: {group_by}')
+                                 f'df: {group_by}, series.group_by: {value.group_by}')
+            if value.base_node != base_node:
+                raise ValueError(f'Base_node in `series` should match dataframe. '
+                                 f'df: {base_node}, series.base_node: {value.base_node}')
+
             self._data[key] = value
 
         for value in index.values():
--- a/bach/bach/series/series.py
+++ b/bach/bach/series/series.py
@@ -122,20 +122,6 @@ class Series(ABC):
         self._sorted_ascending = sorted_ascending
         self._index_sorting = index_sorting
 
-    def _update_self_from_series(self, series: 'Series') -> 'Series':
-        """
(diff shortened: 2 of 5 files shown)


bach/bach/dataframe.py
```diff
@@ -196,7 +196,11 @@ def __init__(
                                  f'df: {value.index}, series.index: {index}')
             if value.group_by != group_by:
                 raise ValueError(f'Group_by in `series` should match dataframe. '
-                                 f'df: {value.group_by}, series.index: {group_by}')
+                                 f'df: {group_by}, series.group_by: {value.group_by}')
+            if value.base_node != base_node:
+                raise ValueError(f'Base_node in `series` should match dataframe. '
+                                 f'df: {base_node}, series.base_node: {value.base_node}')
```
KathiaBarahona, line 202: could you please add a test case for this?

KathiaBarahona (approved):
lgtm, just a small comment

thijs-obj:
Addressed the review comment, so merging this. Let me know if anybody has additional commetns

PR #472, opened 2022-02-11
Support multilevel index in from_pandas

hendrik-obj:


KathiaBarahona (approved):
lgtm

bach/bach/from_pandas.py
```diff
@@ -214,5 +211,5 @@ def _from_pd_shared(
 
         dtypes[str(column)] = dtype
 
-    index_dtypes[index] = dtypes.pop(index)
+    index_dtypes = {index_name: dtypes.pop(index_name) for index_name in index}
```
jansentom, line 214: duplicate index names get truncated here. Should we raise instead?

hendrik-obj: good point. fixed above.

PR #476, opened 2022-02-14
Bach describe support all types

thijs-obj:
This PR was created after discussion here: https://github.com/objectiv/objectiv-analytics/pull/468/files#r803753691

Changes
* removed `_SUPPORTED_SERIES_TYPE` constant from `describe.py`. We now support all types
* In `describe()` check whether a type has actually implemented a function before calling it
* in `SeriesJsonb`: raise `NotImplementedError` in `min()` and `max()`
* Added tests for describe for more data types

This addresses part of issue https://github.com/objectiv/objectiv-analytics/issues/430


bach/bach/describe.py
```diff
@@ -74,21 +55,72 @@ def __init__(
...
 
         if self.percentiles and any(pt < 0 or pt > 1 for pt in self.percentiles):
             raise ValueError('percentiles should be between 0 and 1.')
 
+    @staticmethod
+    def determine_series_to_describe(
+            df: DataFrame,
+            include: Optional[Union[str, Sequence[str]]],
+            exclude: Optional[Sequence[str]]
+    ) -> List[str]:
```
KathiaBarahona, line 71: Just for keeping the same indentation and style in the module
```suggestion
    def determine_series_to_describe(
        df: DataFrame,
        include: Optional[Union[str, Sequence[str]]],
        exclude: Optional[Sequence[str]],
    ) -> List[str]:
```

bach/bach/describe.py
```diff
@@ -74,21 +55,72 @@ def __init__(
...
+
+        numeric_dtypes = {col.dtype for col in df.data.values() if isinstance(col, SeriesAbstractNumeric)}
+        all_dtypes = {col.dtype for col in df.data.values()}
+
+        # process `include` parameter
+        if include is None:
+            if numeric_dtypes:
+                include_dtypes = numeric_dtypes
+            else:
+                include_dtypes = all_dtypes
```
KathiaBarahona, line 89: Let's try to avoid nested `if-else`
```suggestion
            include_dtypes = numeric_dtypes or all_dtypes
```

bach/bach/describe.py
```diff
@@ -74,21 +55,72 @@ def __init__(
...
         self.percentiles = percentiles or [0.25, 0.5, 0.75]
 
         if self.percentiles and any(pt < 0 or pt > 1 for pt in self.percentiles):
             raise ValueError('percentiles should be between 0 and 1.')
 
+    @staticmethod
+    def determine_series_to_describe(
+            df: DataFrame,
+            include: Optional[Union[str, Sequence[str]]],
+            exclude: Optional[Sequence[str]]
```
KathiaBarahona, line 70: Shouldn't `exclude` be `Optional[Union[str, Sequence[str]]`?

bach/bach/describe.py
```diff
@@ -74,21 +55,72 @@ def __init__(
...
+        # process `include` parameter
+        if include is None:
+            if numeric_dtypes:
+                include_dtypes = numeric_dtypes
+            else:
+                include_dtypes = all_dtypes
+        elif include == 'all':
+            include_dtypes = all_dtypes
+        elif isinstance(include, str):
+            raise ValueError(f'Only valid string value of include is "all", value: {include}')
```
KathiaBarahona, line 93: mmm this breaks `df.describe(include='string')`

bach/bach/describe.py
```diff
@@ -74,21 +55,72 @@ def __init__(
...
+        elif isinstance(include, abc.Sequence) and all(isinstance(item, str) for item in include):
+            include_dtypes = set(include)
+        else:
+            raise ValueError(f'Unexpected include value: {include}')
+
+        # process `exclude` parameter
+        if exclude is None:
+            exclude_dtypes = set()
+        elif isinstance(exclude, str):
+            raise ValueError(f'exclude cannot be a string, value: {exclude}')
```
KathiaBarahona, line 103: This is not true, user should be able to provide a single string if just wants to exclude 1 type

bach/bach/describe.py
```diff
@@ -74,21 +55,72 @@ def __init__(
...
+            if numeric_dtypes:
+                include_dtypes = numeric_dtypes
+            else:
+                include_dtypes = all_dtypes
+        elif include == 'all':
+            include_dtypes = all_dtypes
+        elif isinstance(include, str):
+            raise ValueError(f'Only valid string value of include is "all", value: {include}')
+        elif isinstance(include, abc.Sequence) and all(isinstance(item, str) for item in include):
+            include_dtypes = set(include)
```
KathiaBarahona, line 95: should validate if the dtype is actually supported by bach.

bach/bach/describe.py
thijs-obj: OK.

I think both is equally valid tho. `x or y` is still an if-else in my book ;)

bach/bach/describe.py
thijs-obj: aah right. I didn't realize we need to support that.

Looking at pandas docs they don't support this https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.describe.html
It's either the string 'all', or a list of dtypes, or None

bach/bach/describe.py
thijs-obj: Do we need this?

Pandas doesn't support this https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.describe.html exclude is a list or None. 

Can easily add this tho if this is important

bach/bach/describe.py
thijs-obj: good point. Fixed in https://github.com/objectiv/objectiv-analytics/pull/476/commits/689106dff38002f721d25c1e5e2b3c375c23097a

bach/bach/describe.py
```diff
@@ -74,21 +55,73 @@ def __init__(
...
+        else:
+            raise ValueError(f'Unexpected include value: {include}')
+        # resolve dtype-aliasses to the actual dtype
+        include_dtypes = {get_series_type_from_dtype(dtype).dtype for dtype in include_dtypes}  # type: ignore
```
thijs-obj, line 96: added `# type: ignore` here because mypy doesn't understand that `Series.dtype` is a string.

The fact that `Series.dtype` is a `property`, `classmethod`, and `abstractmethod` is a bit too much for myp.

I will clean this up in a small followup PR, but that's a bit of out of scope here I think

thijs-obj: https://github.com/objectiv/objectiv-analytics/pull/483

bach/bach/describe.py
thijs-obj: Fixed https://github.com/objectiv/objectiv-analytics/pull/476/commits/642188ee49ca7e123b537efd9778518445dd66ae

bach/bach/describe.py
thijs-obj: Had out of band discussion: Pandas actually does support a single dtype.

Updated code to match that behaviour. https://github.com/objectiv/objectiv-analytics/pull/476/commits/642188ee49ca7e123b537efd9778518445dd66ae

bach/bach/describe.py
thijs-obj: https://github.com/objectiv/objectiv-analytics/pull/476/commits/642188ee49ca7e123b537efd9778518445dd66ae

bach/bach/describe.py
```diff
@@ -74,21 +55,73 @@ def __init__(
...
+        self.series_to_describe = self.determine_series_to_describe(self.df, include, exclude)
 
         self.datetime_is_numeric = datetime_is_numeric
         self.percentiles = percentiles or [0.25, 0.5, 0.75]
 
         if self.percentiles and any(pt < 0 or pt > 1 for pt in self.percentiles):
             raise ValueError('percentiles should be between 0 and 1.')
 
+    @staticmethod
+    def determine_series_to_describe(
```
KathiaBarahona, line 67: is there a way we can avoid duplicated code? it seems that processing `include` and `exclude` block codes are basically the same

KathiaBarahona (commented):
nice tests!

bach/bach/describe.py
thijs-obj: I think this is the most logical way to do that: https://github.com/objectiv/objectiv-analytics/pull/476/commits/59f944a9e1888ff338f63485d5b5b0dc72480ee5

Didn't do that initially because it's actually more code, and the indirection of having to follow the function call made it a bit more complex imho. Let me know what you think.

KathiaBarahona: also, is there a reason this is a `staticmethod`? why not a `classmethod` instead, I mean we a referencing to the class when calling `str_or_sequence_to_dtypes`

KathiaBarahona (approved):
lgtm, just a small comment

bach/bach/describe.py
thijs-obj: changed

PR #478, opened 2022-02-14
Add missing comment

thijs-obj:


bach/bach/series/series.py
```diff
@@ -1349,6 +1349,8 @@ def drop_duplicates(self: T, keep: Union[str, bool] = 'first') -> T:
         """
         Return a series with duplicated rows removed.
 
+        Note that we don't support 'inplace' as a parameter, as Series are immutable.
```
thijs-obj, line 1352: I think one of the conclusions from this morning was that we should have a docstring explaining this. But I am not sure this is actually an improvement.

There are more parameters we don't support, and generally we don't offer an explanation. In the future we might also have users that only know Bach, and don't know pandas. To them such comments won't make any sense. WDYT?

jansentom: I sadly agree.

KathiaBarahona (approved):
lgtm

PR #480, opened 2022-02-14
Kathiabarahona/timedelta numeric support

KathiaBarahona:
## Description
- support for `TimeDelta.quantile`
- Created `TimeDeltaOperation`
    * Has following accessors:
         - days
         - seconds
         - microseconds
         - components
        
Also verified if currently we are supporting `min`, `max`, `mean` .. etc for timedeltas and it's working properly. Will be adding tests for this since I didn't notice any.


bach/bach/series/series_datetime.py
```diff
@@ -39,6 +42,53 @@ def sql_format(self, format_str: str) -> SeriesString:
...
+    @property
+    def components(self) -> DataFrame:
+        formats = {
+            'days': 'DD',
+            'hours': 'HH24',
+            'minutes': 'MI',
+            'seconds': 'SS',
+            'milliseconds': 'MS',
+            'microseconds': 'US',
+        }
```
KathiaBarahona, line 55: will create an enum for this, I guess we might need it for other situations

bach/bach/series/series_datetime.py
```diff
@@ -39,6 +42,53 @@ def sql_format(self, format_str: str) -> SeriesString:
...
+    @property
+    def seconds(self) -> SeriesInt64:
+        seconds_series = (self._get_epoch() % _SECONDS_IN_DAY) // 1
+
+        seconds_series = seconds_series.astype('int64')
+        return cast(SeriesInt64, seconds_series.copy_override(name='seconds'))
+
+    @property
+    def microseconds(self) -> SeriesInt64:
+        microseconds_series = (self._get_epoch() % 1) % 1 * 10**6
```
jansentom, line 79: Why is the double mod 1 required? Could use a comment if it wasn't an oversight.

bach/bach/series/series_datetime.py
```diff
@@ -39,6 +42,53 @@ def sql_format(self, format_str: str) -> SeriesString:
...
+        microseconds_series //= 1
+
+        microseconds_series = microseconds_series.astype('int64')
+        return cast(SeriesInt64, microseconds_series.copy_override(name='microseconds'))
+
+    def _get_epoch(self) -> SeriesFloat64:
```
jansentom, line 85: I was a bit confused by the precision of this thing. It returns the epoch in seconds, with more precion behind the comma. That took me a while, maybe nice to add a comment?

bach/bach/series/series_datetime.py
```diff
@@ -39,6 +42,53 @@ def sql_format(self, format_str: str) -> SeriesString:
         return str_series
 
 
+class TimeDeltaOperation(DateTimeOperation):
```
jansentom, line 45: nice to have this!

jansentom: Let's match the case of the Series
```suggestion
class TimedeltaOperation(DateTimeOperation):
```

bach/tests/functional/bach/test_series_timedelta.py
```diff
@@ -85,3 +89,88 @@ def test_to_pandas():
...
+
+
+def test_timedelta_operations():
+    pdf = pd.DataFrame(
+        data={
+            'start_date': [
+                datetime.datetime(year=2022, month=1, day=d) for d in range(1, 17)
+            ],
+            'end_date': [
+                datetime.datetime(year=2022, month=m, day=d) for m in range(2, 4) for d in range(1, 9)
```
jansentom, line 101: should we test negative tds too? (all tests)

bach/bach/series/series_datetime.py
jansentom: nice to do this with Bach arithmetic btw!

jansentom (approved):
Did a high level scan only, but from that perspective it looks nice. Left some comments for you,

bach/bach/series/series_datetime.py
KathiaBarahona: yes, to be honest I haven't add any comment/docstring since it is a bit complicated to explain. Took an evening for me to even figure out that formula

bach/tests/functional/bach/test_series_timedelta.py
KathiaBarahona: I actually test a negative tds here:
https://github.com/objectiv/objectiv-analytics/blob/27f7a5efbccdc3c347e7870bda05f2494bdab405/bach/tests/functional/bach/test_series_timedelta.py#L115-L149

KathiaBarahona: can add the expected data to make a bit more clear

bach/bach/series/series_datetime.py
KathiaBarahona: actually, will make this a property called `total_seconds`. I think it will be just nice to have the interval as a numerical value

PR #482, opened 2022-02-15
drop_duplicates w sort_by

KathiaBarahona:
Added `sort_by` and `ascending` to `DataFrame.drop_duplicates`. I didn't for `Series.drop_duplicates` because it is a bit odd, by default the sorting is performed using the index and in case it has no index, it doesn't matter the order :woman_shrugging: .


thijs-obj (approved):
LGTM

PR #483, opened 2022-02-15
Make Series.dtype a simple class level string instead of a property-c…

thijs-obj:
…lassmethod-abstractmethod

### DIFF
--- a/bach/bach/series/series.py
+++ b/bach/bach/series/series.py
@@ -36,7 +36,8 @@ WrappedWindow = Union['Window', 'DataFrame']
 
 class Series(ABC):
     """
-    A Series that represents the generic type and its specific operations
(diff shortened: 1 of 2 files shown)


bach/bach/series/series.py
```diff
@@ -136,19 +159,6 @@ def _update_self_from_series(self, series: 'Series') -> 'Series':
         self._index_sorting = series._index_sorting
         return self
 
-    @property
-    @classmethod
-    @abstractmethod
-    def dtype(cls) -> str:
```
jansentom, line 142: Bye, you'll not be missed.

jansentom: @borft Check this out, no more errors on `make html`

jansentom (approved):
LGTM!

thijs-obj:
one approval, so will merge this when the tests pass

PR #490, opened 2022-02-17
Small collector docker fix

borft:


backend/docker/Dockerfile
```diff
@@ -92,11 +92,11 @@ COPY docker/gunicorn.conf.py /etc/
 CMD /services/entry_point.sh
 
 # python requirements
-COPY requirements.txt /services/
+COPY requirements.in /services/
```
borft, line 95: @thijs-obj use the list of packages without the hashes, because they are not cross-platform. wdyt?

thijs-obj: make sense to me. Would be great to have a comment in this file explaining just that, I think

thijs-obj: In the future we should maybe see how others do it. Because checking the hashes is quite nice of course. But I think for now this should suffice

borft: done

PR #492, opened 2022-02-18
Kathiabarahona/df fillna

KathiaBarahona:
`fillna` for `DataFrame`. Supporting `value`, `ffill`, `bfill` methods for filling null gaps.

Still missing tests, docstrings and disclaimer about non-deterministic results


bach/bach/series/series.py
```diff
@@ -1351,6 +1351,18 @@ def window_nth_value(self, n: int, window: WrappedWindow = None):
             self.dtype
         )
 
+    def window_sum(self, window: WrappedWindow = None):
+        """
+        Returns value evaluated at the row that is the n'th row of the window frame.
+        (counting from 1); returns NULL if there is no such row.
+        """
```
KathiaBarahona, line 1358: I'll update this, also need to add a test case

bach/bach/dataframe.py
```diff
@@ -2624,6 +2625,112 @@ def dropna(
...
+
+        self._update_self_from_df(df)
+        return None
+
+    def ffill(
+        self,
+        sort_by: Optional[Union[str, Sequence[str]]] = None,
+        ascending: Union[bool, List[bool]] = True,
+        inplace: bool = False,
+    ) -> Optional['DataFrame']:
```
KathiaBarahona, line 2667: here too

bach/bach/dataframe.py
```diff
@@ -2624,6 +2625,112 @@ def dropna(
...
+
+        self._update_self_from_df(df)
+        return None
+
+    def bfill(
+        self,
+        sort_by: Optional[Union[str, Sequence[str]]] = None,
+        ascending: Union[bool, List[bool]] = True,
+        inplace: bool = False,
+    ) -> Optional['DataFrame']:
```
KathiaBarahona, line 2713: same :)

bach/bach/dataframe.py
```diff
@@ -2624,6 +2625,112 @@ def dropna(
...
+                    f'case when {{}} is null then 0 else 1 end', self.all_series[series_name],
+                ),
+                name=partition_name,
+            )
+
+            df[partition_name] = df.all_series[partition_name].window_sum(
+                window=Window([], order_by=df.order_by),
+            )
+
+        df.materialize(inplace=True)
```
jansentom, line 2692: What happens here is quite smart/magical. I needed to look at the SQL generated to actually understand:
in pseudo-sql:
```sql
sum(case when "series_name" is null then 0 else 1 end) over ( order by "df.order_by" asc RANGE BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW) as "__partition_series_name"
```
I do think it needs a very detailed explanation here in the code.

jansentom: Please give name to the materialization btw
```suggestion
        df.materialize(node_name='fillna_partitioning', inplace=True)
```

bach/tests/functional/bach/test_df_fillna.py
```diff
@@ -0,0 +1,115 @@
...
+
+DATA = [
+    [None, None, None, None],
+    [3, 4, None, 1],
+    [None, 3, None, 4],
+    [None, 2, None, 0],
+    [None, None, None, None],
+    [1, None, None, None],
+    [None, None, None, None],
+    [None, None, 1, None],
```
jansentom, line 14: Easier to read:

```suggestion
    [None, None, None, None],
    [3,    4,    None, 1],
    [None, 3,    None, 4],
    [None, 2,    None, 0],
    [None, None, None, None],
    [1,    None, None, None],
    [None, None, None, None],
    [None, None, 1,    None],
```

jansentom: I'm also missing test cases with sparse data, e.g. `None, 1, None, None, 2, None, 3, None`

bach/tests/functional/bach/test_df_fillna.py
```diff
@@ -0,0 +1,115 @@
+import numpy as np
```
jansentom, line 1: Don't think this is used?

bach/tests/functional/bach/test_df_fillna.py
```diff
@@ -0,0 +1,115 @@
...
+import pandas as pd
```
jansentom, line 2: Does pandas give different results in these tests? Could we compare to pandas output?

bach/bach/series/series.py
```diff
@@ -1351,6 +1351,18 @@ def window_nth_value(self, n: int, window: WrappedWindow = None):
             self.dtype
         )
 
+    def window_sum(self, window: WrappedWindow = None):
```
jansentom, line 1354: I don't think this method is required. Also see comment in `ffill()` implementation above.

bach/bach/dataframe.py
```diff
@@ -2624,6 +2625,112 @@ def dropna(
...
+        if sort_by:
+            df = df.sort_values(by=sort_by, ascending=ascending)
+
+        if not sort_by and not df.order_by:
+            raise Exception('dataframe must be sorted in order to apply forward fill.')
+
+        # create partition per each non-nullable value in all columns to be interpolated
+        for series_name in self.data_columns:
+            partition_name = f'__partition_{series_name}'
+            df[partition_name] = self.all_series[series_name].copy_override(
```
jansentom, line 2681: This needs to become a `SeriesInt64`, and if that's the case, you should be able to use `SeriesNumeric.sum()` instead of the new `window_sum` implementation.

jansentom (changes requested):
Smart way to do this, took me a while to get why it worked. I do have some problems with some parts of the implementation though, but they should not be super hard to address.

bach/tests/functional/bach/test_df_fillna.py
KathiaBarahona: we definitely can

KathiaBarahona: oh :crying_cat_face: gives different results, seems because of the way they sort NaNs.

bach/tests/functional/bach/test_df_fillna.py
KathiaBarahona: thanks, found a bug with this one!

jansentom: 662fq8

bach/bach/dataframe.py
```diff
@@ -2624,6 +2626,155 @@ def dropna(
...
+            or a dictionary specifying which scalar/series to use for each series.
+        :param method: Method to use for filling NULL values on all DataFrame series. Supported values:
+           - "ffill"/"pad": Fill missing values by propagating the last non-nullable value in the series.
+           - "bfill"/"backfill": Fill missing values with the next non-nullable value in the series.
+        :param axis: only ``axis=0`` is supported.
+        :param sort_by: series label or sequence of labels used to sort values.
+            Sorting of values is needed since result might be non-deterministic, as rows with NULLs might
+            yield different results affecting the values to be propagated when using a filling method.
+        :param ascending: Whether to sort ascending (True) or descending (False). If this is a list, then the
+            `by` must also be a list and ``len(ascending) == len(by)``.
```
thijs-obj, line 2651:
```suggestion
            `sort_by` must also be a list and ``len(ascending) == len(sort_by)``.
```

thijs-obj: same for `ffill` and `bfill`

bach/bach/dataframe.py
```diff
@@ -2624,6 +2626,155 @@ def dropna(
...
+
+        :return: a new dataframe with filled missing values.
+
+        .. note::
+            sort_by is required if DataFrame has no order_by.
```
thijs-obj, line 2656:
```suggestion
            sort_by is required if method is specified and the DataFrame has no order_by.
```

bach/bach/dataframe.py
```diff
@@ -2624,6 +2626,155 @@ def dropna(
...
+        :param axis: only ``axis=0`` is supported.
+        :param sort_by: series label or sequence of labels used to sort values.
+            Sorting of values is needed since result might be non-deterministic, as rows with NULLs might
+            yield different results affecting the values to be propagated when using a filling method.
+        :param ascending: Whether to sort ascending (True) or descending (False). If this is a list, then the
+            `by` must also be a list and ``len(ascending) == len(by)``.
+
+        :return: a new dataframe with filled missing values.
+
+        .. note::
```
thijs-obj, line 2655: I think we should have some note, either here or in some other place. That if the sort_by is not deterministic, that the resulting data in the DataFrame is then non-deterministic and 'unstable', i.e. the data of this operation might change retro-actively after doing other operations on the resulting DataFrame

bach/bach/dataframe.py
```diff
@@ -45,6 +45,7 @@
 #     - dict of axis labels -> functions, function names or list of such.
 
 Level = Union[int, List[int], str, List[str]]
+Scalar = Optional[Union[int, float, str, UUID]]
```
thijs-obj, line 48:
```suggestion
Scalar = Union[int, float, str, bool, date, datetime, time, UUID]
```

This should work with date etc. too right?

I realize that my suggestion is not exhaustive either, but seems nice to have the python types for the 'basic' series at least. 
Also I realize that  `bool` is technically included already as it's a subtype of `int`, but I think it's nice to be explicit

bach/bach/dataframe.py
```diff
@@ -756,6 +757,7 @@ def set_savepoint(self, name: str, materialization: Union[Materialization, str]
         Set the current state as a savepoint in `self.savepoints`.
 
         :param save_points: Savepoints object that's responsible for tracking all savepoints.
+        :param name: Name for the savepoint. This will be the name of the table or view if that's set as
```
thijs-obj, line 760:
```suggestion
```

line is a duplicate of the line below it

bach/bach/dataframe.py
```diff
@@ -2624,6 +2626,155 @@ def dropna(
...
+        ascending: Union[bool, List[bool]] = True,
+    ) -> 'DataFrame':
+        """
+        Fill missing values by propagating the last non-nullable value in each series.
+
+        :param sort_by: series label or sequence of labels used to sort values.
+            Sorting of values is needed since result might be non-deterministic, as rows with NULLs might
+            yield different results affecting the values to be propagated when using a filling method.
+        :param ascending: Whether to sort ascending (True) or descending (False). If this is a list, then the
+            `by` must also be a list and ``len(ascending) == len(by)``.
```
thijs-obj, line 2691:
```suggestion
            `sort_by` must also be a list and ``len(ascending) == len(sort_by)``.
```

bach/bach/dataframe.py
```diff
@@ -2624,6 +2626,155 @@ def dropna(
...
+        ascending: Union[bool, List[bool]] = True,
+    ) -> 'DataFrame':
+        """
+        Fill missing values by using the next non-nullable value in each series.
+
+        :param sort_by: series label or sequence of labels used to sort values.
+            Sorting of values is needed since result might be non-deterministic, as rows with NULLs might
+            yield different results affecting the values to be propagated when using a filling method.
+        :param ascending: Whether to sort ascending (True) or descending (False). If this is a list, then the
+            `by` must also be a list and ``len(ascending) == len(by)``.
```
thijs-obj, line 2755:
```suggestion
            `sort_by` must also be a list and ``len(ascending) == len(sort_by)``.
```

bach/bach/dataframe.py
```diff
@@ -2624,6 +2626,155 @@ def dropna(
...
+
+        df = self.copy()
+
+        if sort_by:
+            df = df.sort_values(by=sort_by, ascending=ascending)
+
+        if not sort_by and not df.order_by:
+            raise Exception('dataframe must be sorted in order to apply forward fill.')
+
+        # create a partition column for each series to be filled
```
thijs-obj, line 2709: I very much appreciate this explanation :ok_hand:

bach/bach/dataframe.py
```diff
@@ -2624,6 +2626,155 @@ def dropna(
...
+        from bach.partitioning import Window, WindowFrameMode
+        from bach.series.series_numeric import SeriesInt64
+
+        df = self.copy()
+
+        if sort_by:
+            df = df.sort_values(by=sort_by, ascending=ascending)
+
+        if not sort_by and not df.order_by:
+            raise Exception('dataframe must be sorted in order to apply forward fill.')
```
thijs-obj, line 2707:
```suggestion
            raise Exception('dataframe must be sorted in order to apply forward or backward fill.')
````

This exception can also be indirectly triggered through `bfill()`

File: bach/tests/functional/bach/test_df_fillna.py (hunk 5 above)
Comment (KathiaBarahona) [reply bach/tests/functional/bach/test_df_fillna.py]: LMAO

File: bach/docs/source/core_concepts.rst [hunk 19]
```diff
@@ -36,8 +36,9 @@ database. This also results in a few differences in how DataFrames form both lib
 situations:
 
 * The order of rows in a Bach DataFrame can be non-deterministic. If there is not a deterministic
-  :py:meth:`DataFrame.sort_values()` call, then the order of the rows that the data-transfer
-  functions return can be unpredictable.
+  :py:meth:`DataFrame.sort_values()` or :py:meth:`DataFrame.fillna()` call, then the order of the rows that the data-transfer
```
thijs-obj, line 39: I think we should have a more comprehensive explanation here. See also my comment above. 

`fillna()` (and `drop_duplicates()`)  are a special case of tricky. If you do `fillna()` on a DataFrame that's not deterministically sorted, and then you continue working with that frame, then even twenty operations later the data is still non-deterministic. So you might have a dataframe whose data is 'unstable' and you might never notice. You might do `head()` in between to check on the data, and you might sort the dataframe multiple times afterwards, but none of that fixes the unstable data situation. In the end all operations are compressed to one big sql query, and if somewhere at the start there is a non-deterministic part, then the whole query is non-deterministic.

Not really sure what the right words are to explain this. We could also do this as a separate issue, as the maybe the explanation should be a whole subsection in this file, but we definitely need to explain this well.


P.S. `window_row_number()`, `window_rank()`, and related functions have the same problem potential, but I imagine they are less likely to be used so that helps a bit there.

bach/bach/dataframe.py
```diff
@@ -2624,6 +2626,155 @@ def dropna(
...
+        .. note::
+            sort_by is required if DataFrame has no order_by.
+        """
+        df = self.copy()
+
+        if sort_by:
+            df = df.sort_values(by=sort_by, ascending=ascending)
+
+        main_order_by = copy(df._order_by)
+        # bfill is similar to ffill, the difference is that the order is reversed.
```
thijs-obj, line 2768: nice method of doing this :ok_hand:

bach/bach/dataframe.py
```diff
@@ -2624,6 +2626,155 @@ def dropna(
...
+        df._order_by = [
```
thijs-obj, line 2769: Could we use `copy_override()` here?

Generally we almost never assign anything to `df._x` but always call `df.copy_override(x=)`. This makes it a lot easier to check that we don't accidentally mutate state. For example just searching for `self._` in this file make it clear that there are only a few places where we do in-place changes, and it's easy to check all occurrences. If we have a lot of legitimate `self._x=` usage instead of `df.copy_override()` then it will be harder to audit that in the future

thijs-obj: for example something like
```python
        main_order_by = df.order_by
        reverse_order_by = [
            SortColumn(expression=ob.expression, asc=not ob.asc)
            for ob in main_order_by
        ]
        return df.copy_override(sort_by=reverse_order_by).ffill().copy_override(sort_by=main_order_by)
```

thijs-obj: This opinion of my might be controversial with other reviewers.

thijs-obj (commented):
I like this PR a lot.

Seems very complete, good docstrings, tests, straightforward code, and inline comments explaining things :+1:

Do have a few comments here and there.

bach/bach/dataframe.py
KathiaBarahona: oh wow I'm  not even sure how I added that change. Thanks!

bach/bach/dataframe.py
KathiaBarahona: sounds good to me

bach/docs/source/core_concepts.rst
KathiaBarahona: I think it is better to create an issue that addresses this, because to be honest I was/am not really sure how to explain it in a clear way, maybe it is one of those situations that it's better to explain through examples than words. What about creating an specific section just for this?

bach/bach/dataframe.py
```diff
@@ -2624,6 +2626,155 @@ def dropna(
...
+        # |   3  |       2       |     3    |
+        # | NULL |       2       |     3    |
+        # |   4  |       3       |     4    |
+        for series_name, series in self._data.items():
+            partition_name = f'__partition_{series.name}'
+            partition_expr = Expression.construct(f'case when {{}} is null then 0 else 1 end', series)
+            partition_series = cast(
+                SeriesInt64,
+                series.copy_override(expression=partition_expr, name=partition_name).astype('int64'),
+            )
```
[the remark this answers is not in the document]
KathiaBarahona: ohhh shoot, I should test for all dtypes

bach/bach/dataframe.py
KathiaBarahona: it sorta should. I'd just copied the types supported in `const_to_series` since that's what we call on Series.fillna

bach/bach/dataframe.py
KathiaBarahona: Will add it there. I created an issue for it as well, https://github.com/objectiv/objectiv-analytics/issues/498

thijs-obj: I think we should use the `.. warning::` syntax for the warning about 'if sort_by is non-deterministic'. For an example on how that looks see https://staging.objectiv.io/docs/modeling/DataFrame/bach.DataFrame.window

Also good to do that as part of https://github.com/objectiv/objectiv-analytics/issues/498, but then we must not forget to pick up that issue soon

thijs-obj: Can we copy the warning to `bfill()` and `ffill()` too?

thijs-obj (approved):
two small comments. otherwise looks good to me.

jansentom (approved):
LGTM, thanks for the fixes!

PR #495, opened 2022-02-21
remove inplace from DataFrame

KathiaBarahona:
Removed all `inplace` params from most of DataFrame's functions, except for `materialize`, as `set_savepoint` actually updates `self`.

## Checklist
- [X] Make sure all notebooks that use this are updated
- [X] Update tests
- [] Make sure that changelog is clear on this


bach/bach/operations/describe.py
```diff
@@ -177,17 +177,19 @@ def _calculate_stat(
         if not series_to_aggregate:
             return None
 
-        stat_df = self.df[series_to_aggregate]
```
thijs-obj, line 180: maybe we should add a function to DataFrame to do this in a type safe manner, i.e. a function that will always return a DataFrame. Anyways, something for another time

thijs-obj (approved):

PR #496, opened 2022-02-21
Preliminary Snowplow sink support

borft:
Adds preliminary support for pushing events to the snowplow pipeline, using GCP pubsub


backend/requirements.txt
```diff
@@ -4,41 +4,138 @@
 #
```
borft, line 4: @thijs-obj So I regenerated this file, as it seemed to be broken, but that of course also resulted in a lot of updates of other packages. wdyt?

PR #506, opened 2022-02-23
remove series calling private method, fixing outer merge

KathiaBarahona:
Refactoring of `_get_column_name_expr_dtype` and `_determine_result_columns`.  This solves a problem we have when doing outer joins, since we are considering only left table values when having conflicting on columns. This way we also remove the hack of calling `df._index_merge` in Series class.

Let me know what do you think about always applying `COALESCE` on columns involved in the merge `on`.  Just want to avoid passing the how on multiple functions just to evaluate if we are doing `how == How.outer`. Not sure if doing this might be actually inefficient. 

Still missing changing the code from `DataFrame._index_merge`

### DIFF
--- a/bach/bach/merge.py
+++ b/bach/bach/merge.py
@@ -1,6 +1,7 @@
 """
 Copyright 2021 Objectiv B.V.
 """
+from collections import defaultdict
 from copy import copy
 from enum import Enum
 from typing import Union, List, Tuple, Optional, Dict, Set, NamedTuple, Hashable
@@ -126,33 +127,36 @@ def _determine_result_columns(
     Determine which columns should be in the DataFrame after merging left and right, with the given
     left_on and right_on values.
     """
-    filter_columns_from_right = _get_filter_columns_from_right(left_on, right_on)
-    left_index = left.index
-    if right.index:
-        right_index = {
-            label: series for label, series in right.index.items()
-            if label not in filter_columns_from_right
-        }
-    else:
-        right_index = {}
-
-    left_data = left.data
-    if isinstance(right, DataFrame):
-        right_data = right.data
-    elif isinstance(right, Series):
-        right_data = {right.name: right}
-    else:
+    if not isinstance(right, (DataFrame, Series)):
         raise TypeError(f'Right should be DataFrameOrSeries type: {type(right)}')
-    right_data = {
-        label: series for label, series in right_data.items()
-        if label not in filter_columns_from_right
-    }
 
-    conflicting = set({**left_index, **left_data}.keys()).intersection(set({**right_index, **right_data}))
-    new_index_list = _get_column_name_expr_dtype(left_index, conflicting, suffixes[0], 'l')
-    new_index_list += _get_column_name_expr_dtype(right_index, conflicting, suffixes[1], 'r')
-    new_data_list = _get_column_name_expr_dtype(left_data, conflicting, suffixes[0], 'l')
-    new_data_list += _get_column_name_expr_dtype(right_data, conflicting, suffixes[1], 'r')
+    left_df = left.copy()
+    right_df = right.copy() if isinstance(right, DataFrame) else right.to_frame()
+
+    conflicting_on = {l_on for l_on, r_on in zip(left_on, right_on) if l_on == r_on}
+    conflicting = (
+        (set(left_df.index) | set(left_df.data)) & (set(right_df.index) | set(right_df.data))
+    )
+    # don't add suffixes to on columns
+    # need to consider values from both objects (important when how = How.outer)
+    conflicting -= conflicting_on
+
+    new_index_list = _get_merged_result_series(
+        left_series=left_df.index,
+        right_series=right_df.index,
+        suffixes=suffixes,
+        conflicting_names=conflicting,
+        conflicting_on=conflicting_on,
+    )
+
+    new_data_list = _get_merged_result_series(
+        left_series=left_df.data,
+        right_series=right_df.data,
+        suffixes=suffixes,
+        conflicting_names=conflicting,
+        conflicting_on=conflicting_on,
+    )
+
     _check_no_column_name_conflicts(new_index_list + new_data_list)
     return new_index_list, new_data_list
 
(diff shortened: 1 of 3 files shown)


bach/bach/merge.py
```diff
@@ -126,33 +127,36 @@ def _determine_result_columns(
...
-    new_data_list += _get_column_name_expr_dtype(right_data, conflicting, suffixes[1], 'r')
+
+    left_df = left.copy()
+    right_df = right.copy() if isinstance(right, DataFrame) else right.to_frame()
+
+    conflicting_on = {l_on for l_on, r_on in zip(left_on, right_on) if l_on == r_on}
+    conflicting = (
+        (set(left_df.index) | set(left_df.data)) & (set(right_df.index) | set(right_df.data))
+    )
+    # don't add suffixes to on columns
```
KathiaBarahona, line 140:
```suggestion
    # don't add suffixes to conflicted on columns
```

bach/bach/dataframe.py
```diff
@@ -992,43 +992,38 @@ def _index_merge(
...
-                        right_index=True,
-                        how=how,
-                        suffixes=('', '__remove'))
+        # align index names, this way we have all matched indexes in a single series after merge
+        aligned_index = {index_name: value.index[value_index_name].copy_override(name=index_name)}
+        other = value.copy_override(
+            index=aligned_index,
+            name=key,
+            group_by=GroupBy(list(aligned_index.values())) if value.group_by else None,
+        )
```
KathiaBarahona, line 1013: In a future this can be replaced by `Series.rename`, although renaming of indexes is still not supported

bach/bach/series/series.py
```diff
@@ -504,19 +504,53 @@ def _get_supported(
...
+        ):
+            raise ValueError('dtypes of indexes to be merged should be the same')
+
+        # align index names, this way we have all matched indexes in a single series
+        other_cp = other.copy_override(
+            index={
+                caller_idx.name: other_idx.copy_override(name=caller_idx.name)
+                for caller_idx, other_idx in zip(self.index.values(), other.index.values())
+            }
+        )
```
KathiaBarahona, line 538: This block actually enables merging with multi-level, please let me know if it's ok, else I'll just change the condition to support only in 1 level

bach/bach/merge.py
```diff
@@ -166,54 +170,49 @@ def _check_no_column_name_conflicts(result_columns: List[ResultSeries]):
...
-        left: DataFrame,
-        right: DataFrameOrSeries,
-        how: str,
-        on: Union[str, List[str], None],
-        left_on: Union[str, List[str],  None],  # todo: also support array-like arguments?
-        right_on: Union[str, List[str], None],
-        left_index: bool,
-        right_index: bool,
-        suffixes: Tuple[str, str]
+    left: DataFrame,
```
thijs-obj, line 207: I'd prefer it if we do cosmetic changes to do all functions in a file or, only the ones that are actually modified. Changing indenting on some functions makes the PR bigger, without bringing consistency to the file

bach/bach/series/series.py
```diff
@@ -504,19 +504,53 @@ def _get_supported(
...
-                    return cast('Series', df[self.name]), cast('Series', df['__other'])
-                    # todo pandas: if name of both series are same, use that name of result. we always use
-                    #  name of self
+                    return self._index_merge(other)
 
         if other.dtype.lower() not in supported_dtypes:
             raise TypeError(f'{operation_name} not supported between {self.dtype} and {other.dtype}.')
         return self, other
 
+    def _index_merge(self, other: 'Series') -> Tuple['Series', 'Series']:
```
thijs-obj, line 513: bit long example, but this works on main and not on this branch.
The gist of it is that if the `groupb_y` on the two `Series` don't match, then it gives an error.

```python
# Setup
from bach import DataFrame
from tests.functional.bach.test_data_and_utils import DB_TEST_URL
import sqlalchemy
import pandas as pd
import numpy as np

engine = sqlalchemy.create_engine(DB_TEST_URL)
pdf1 = pd.DataFrame(data={'a': [1, 2, 3], 'b': [2, 2, 2], 'c': [3, 3, 3]},)
pdf1 = pdf1.set_index('a')
pdf2 = pd.DataFrame(data={'x': [1, 2, 3, 4], 'y': [2, 2, 4, 4], 'z': [1, 2, 3, 4]},)
pdf2 = pdf2.set_index('x')
df1 = DataFrame.from_pandas(engine=engine, df=pdf1.reset_index(drop=False), convert_objects=True, materialization='table', if_exists='replace', name='test_table1')
df2 = DataFrame.from_pandas(engine=engine, df=pdf2.reset_index(drop=False), convert_objects=True, materialization='table', if_exists='replace', name='test_table2')
df1 = df1.reset_index(drop=True)
df1 = df1.set_index('a')
df2 = df2.reset_index(drop=True)
df2 = df2.set_index('x')

# works on main, not on current branch
xx = df1.c + df2.groupby('y')['z'].sum()
```

bach/bach/merge.py
```diff
@@ -166,54 +170,49 @@ def _check_no_column_name_conflicts(result_columns: List[ResultSeries]):
...
+        table_alias = 'l' if suffix == suffixes[0] else 'r'
+        for series_name, series in source_series.items():
+            new_name = series_name
+            expr = series.expression.resolve_column_references(table_alias)
+
+            if series_name in conflicting_on:
+                if table_alias == 'r':
+                    continue
+                r_expr = right_series[series_name].expression.resolve_column_references('r')
+                expr = Expression.construct(f'COALESCE({expr.to_sql()}, {r_expr.to_sql()})')
```
thijs-obj, line 192: I think just always doing coalesce here is OK.
We can later optimize in different paths for 'left', 'right' 'inner', 'cross', and 'outer' if we want to.

thijs-obj (changes requested):
Generally the changes in `merge.py` are a nice improvement I think.

**mssing: tests**
I do think we should have some tests for the cases that we previously didn't support and that now work. Otherwise we'll just get regressions in the future.

**other**
Also left some (mostly small) comments on other things

bach/bach/series/series.py
thijs-obj: Seems ok to me. @hendrik-obj  ?

hendrik-obj: yes, looks like it's working. good feature.

bach/tests/functional/bach/test_df_merge.py
```diff
@@ -3,11 +3,12 @@
 """
 from decimal import Decimal
 
+import pandas as pd
 import pytest
 
 from bach import DataFrame
 from tests.functional.bach.test_data_and_utils import get_bt_with_test_data, get_bt_with_food_data, \
-    assert_equals_data, get_bt_with_railway_data
+    assert_equals_data, get_bt_with_railway_data, get_from_df
```
hendrik-obj, line 11: There is an, admittedly untested, situation where the behaviour is now different from main.
When merging a series to a data frame using the name of the index of the series. ie:
```
bt = get_bt_with_test_data(True)
bt.reset_index().merge(bt.city, on='_index_skating_order')
```
Also goes wrong when series have different base nodes. Maybe a bit of an edge case but we should either not support it with an exception or implement it (and test it).

hendrik-obj (commented):
Nice improvement overall.
One comment about an (untested) scenario that gives an error.

bach/bach/series/series.py
KathiaBarahona: ouch! thanks, seems I gotta rename the indexes in the GroupBy

bach/tests/functional/bach/test_df_merge.py
KathiaBarahona: uh hah! actually its failing because `_index_skating_order` is in the index for `bt.city` series.

bach/bach/dataframe.py
```diff
@@ -981,54 +981,48 @@ def __setitem__(self,
...
         Internal method used by __setitem__ to set a series using a merge on index. Modifies the DataFrame
         with the added column. The DataFrames index name is the same as the original DataFrame's.
 
         :param key: name of the column to set.
         :param value: Series that is set.
-        :param how: 'left' or 'outer'.
         """
-
-        # todo This method's functionality will be implementd in merge()
+        # todo This method's functionality will be implemented in merge()
```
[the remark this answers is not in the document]
KathiaBarahona: I agree, since it's only used when setting items. Can't find other place we use it. I'll remove the comment

thijs-obj (commented):
As discussed yesterday: would be great to have a specific test for what failed first and works now, the scenario of `merge(how=outer/right)` with null values in the left dataframe.

thijs-obj (approved):

PR #508, opened 2022-02-24
preliminary support for automatically adding MarketingContexts

borft:
Looks for utm query parameters, and if so creates and adds a MarketingContext based on those


backend/objectiv_backend/end_points/collector.py
```diff
@@ -223,6 +226,43 @@ def add_http_context_to_event(event: EventData, request: Request):
...
+        return
+    else:
+        path_context = path_contexts[0]
+
+    query_string = urlparse(str(path_context.get('id', ''))).query
+    parsed_qs = parse_qs(query_string)
+
+    # for now we only support utm, but other mappings are possible
+    # all mappings that result in a valid MarketingContext will be added
+    mappings = {
```
borft, line 244: I think this needs to be a little bit more configurable in the future, but atleast there's already a mechanism to support multiple mappings

backend/objectiv_backend/end_points/collector.py
```diff
@@ -223,6 +226,43 @@ def add_http_context_to_event(event: EventData, request: Request):
...
+            'content': 'utm_content'
+        }
+    }
+
+    for mapping_type in mappings:
+        mapping = mappings[mapping_type]
+        marketing_context_fields = {f: str(parsed_qs.get(mf, [''])[0]) for f, mf in mapping.items()}
+        marketing_context_fields['id'] = mapping_type
+
+        try:
```
borft, line 259: not sure if this is super nice, but this way, the `MappingContext` gets to decide if the fields provided are good enough to create a valid instance of the context. Which means this code doesn't need to know about the specifics of the schema.

jansentom: hmm, @thijs-obj prob has an opinion here

backend/objectiv_backend/end_points/collector.py
jansentom: quite elegant

backend/objectiv_backend/end_points/collector.py
```diff
@@ -223,6 +226,43 @@ def add_http_context_to_event(event: EventData, request: Request):
...
+            'medium': 'utm_medium',
+            'campaign': 'utm_campaign',
+            'term': 'utm_term',
+            'content': 'utm_content'
+        }
+    }
+
+    for mapping_type in mappings:
+        mapping = mappings[mapping_type]
+        marketing_context_fields = {f: str(parsed_qs.get(mf, [''])[0]) for f, mf in mapping.items()}
```
jansentom, line 256: lazy question: Is this None vs '' in line with what we agreed earlier?

backend/objectiv_backend/end_points/collector.py
```diff
@@ -223,6 +226,43 @@ def add_http_context_to_event(event: EventData, request: Request):
...
+        marketing_context_fields['id'] = mapping_type
+
+        try:
+            add_global_context_to_event(event, MarketingContext(**marketing_context_fields))
+        except ValueError:
+            # couldn't create a marketing context for this mapping, no biggie
```
jansentom, line 262: Let's explain here why it's no biggie instead

jansentom (approved):

backend/objectiv_backend/end_points/collector.py
borft: No, this was a bug. I've changed the logic here:
- if there is a value, use that as a string
- if there's no value, don;'t set it at all. If this is an optional, the ContextClass will fix it, otherwise we fail, which is also OK

backend/objectiv_backend/end_points/collector.py
borft: done

backend/objectiv_backend/end_points/collector.py
thijs-obj: Can we change the comment below to say this. Reading the code I was confused why we expect this to fail in some cases. Data not adhering to schema is a good reason of course

borft: makes sense, done

thijs-obj (approved):

PR #502, opened 2022-02-22
WIP: Bach bq initial support main merged

thijs-obj:
DO NOT review. Only meant as an easy reference for myself to see what changes have been made, and what I want to port to other branches

PR #509, opened 2022-02-24
Bach from_table() add all_dtypes argument

thijs-obj:
Change: 
* Add `all_dtypes` parameter to `DataFrame.from_table()` and `DataFrame.from_model()` that allows the user to specify what the dtypes are.
* Bonus change: use `DataFrame.from_model()` and `from_table()` in `from_pandas()` code

Why: 
* Without the `all_dtypes` argument, `DataFrame.from_table()` always has to query the database to find out the table structure.
* BigQuery queries generally take at least 2 seconds. So adding this makes testing with BigQuery a lot more pleasant.

This will make the diff of "the bigquery branch" (https://github.com/objectiv/objectiv-analytics/pull/502 WIP) smaller


thijs-obj:
Please have a look @jansentom @KathiaBarahona, no hurry

KathiaBarahona:
What about having a MixIn containing all this classmethods? I feel that in a future we might need to add more things to it and the DataFrame class is getting quite big

thijs-obj:
> What about having a MixIn containing all this classmethods? I feel that in a future we might need to add more things to it and the DataFrame class is getting quite big

Maybe a good thing to look at the whole class and think about how we want to manage this in the future and what code we want to move out of the class and where. 

I think I'd rather move some of the helper functions involved in this PR to another file, and keep the main functions here than introduce a mixin now.

thijs-obj:
In https://github.com/objectiv/objectiv-analytics/pull/509/commits/e956687fb7f413b5c7278dc9466f86f03a4472a1 I moved a bunch of code to other files from dataframe.py

bach/bach/dataframe.py
```diff
@@ -362,80 +364,109 @@ def __eq__(self, other: Any) -> bool:
...
+        If all_dtypes is not specified, then a transaction scoped temporary table with will be created with
+        0 result rows from the model. The meta data of this table will be used to deduce the dtypes.
 
         :param engine: an sqlalchemy engine for the database.
         :param model: an SqlModel that specifies the queries to instantiate as DataFrame.
         :param index: list of column names that make up the index. At least one column needs to be
             selected for the index.
+        :param all_dtypes: Optional. Mapping from column name to dtype.
+            Must contain all index and data columns.
+            Must be in same order as the columns appear in the the sql-model.
```
KathiaBarahona, line 426: where is this validated?

bach/bach/from_database.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def get_dtypes_from_table(engine: Engine, table_name: str) -> Dict[str, str]:
+    """ Query database to get dtypes of the given table. """
+    sql = f"""
+        select column_name, data_type
+        from information_schema.columns
+        where table_name = '{table_name}'
+        order by ordinal_position;
+    """
```
KathiaBarahona, line 36: can we have another function that returns this sql with the correct table_name? this way we avoid duplicated code.
smth like:

```python
def _get_table_metadata_query(table_name:str) -> str:
    return f"""
        select column_name, data_type
        from information_schema.columns
        where table_name = '{table_name}'
        order by ordinal_position;
    """
```

bach/tests/functional/bach/test_df_from.py
```diff
@@ -0,0 +1,124 @@
...
+    _create_test_table(engine, table_name)
+    sql_model: SqlModel = CustomSqlModelBuilder(sql=f'select * from {table_name}')()
+
+    df = DataFrame.from_model(engine=engine, model=sql_model, index=['a'])
+    assert df.index_dtypes == {'a': 'int64'}
+    assert df.dtypes == {'b': 'string', 'c': 'float64', 'd': 'date', 'e': 'timestamp'}
+    assert df.is_materialized
+    assert df.base_node.columns == ('a', 'b', 'c', 'd', 'e')
+    # there should only be a single model that selects from the table, not a whole tree
+    # todo: in the future introduce a special SqlModel type 'source', so we don't even need a first model
```
jansentom, line 62: This is a really nice idea. Can you add that to the issue tracker?

jansentom (approved):
Nice improvement!

bach/bach/from_database.py
thijs-obj: honest quetion: Why should we avoid this duplicated code?

bach/bach/dataframe.py
thijs-obj: It's not. The called of this function must make sure that this is the case

bach/bach/from_database.py
KathiaBarahona: feels like both functions that are doing this query are doing the same thing.  but  that's just my opinion :)

thijs-obj: I fully agree that both functions partially do the same thing.

I just don't think it improves the code if we remove this small code duplication. Splitting this of in a separate function leads to more lines in this file, and I personally find it harder to read and don't see much upside to that approach.

PR #515, opened 2022-03-02
stack implementation

KathiaBarahona:
In order to implement `DataFrame.loc`, `DataFrame.stack`, since from cases (accessing index value), the DataFrame's columns are stacked into a new index.

### DIFF
--- a/bach/bach/dataframe.py
+++ b/bach/bach/dataframe.py
@@ -2760,6 +2760,41 @@ class DataFrame:
             )
         return subset
 
+    def stack(self, *, level: int = -1, dropna: bool = True) -> 'Series':
+        """
+        Stacks all data_columns into a single index series.
+
+        :param level: only ``level=-1`` is supported, column axis supports only single levels.
+        :param dropna: Whether to drop rows that contain missing values. If the caller has
+            at least an index series, this might generate different combinations between
+            the index and the stacked values.
+
+        :return: a reshaped series that includes a new index containing the caller's column labels as values.
+        """
+        if level != -1:
+            raise NotImplementedError('column axis supports only one level.')
+
+        df = self.copy()
+        if df.group_by:
+            df = df.materialize()
+
+        dc_dfs = []
+        # convert each data column series to DataFrame and use series name as new index value
+        for dc in df.data_columns:
+            dc_df = df.all_series[dc].copy_override(name='__stacked').to_frame()
+            dc_df[f'__stacked_index'] = dc
+            dc_dfs.append(dc_df)
+
+        # concat all dataframes to get new_index and stacked values in two single series
+        from bach.operations.concat import DataFrameConcatOperation
+        stacked_df = DataFrameConcatOperation(dc_dfs)()
+
+        # append the stacked index to the initial indexes
+        stacked_df = stacked_df.set_index(list(self.index_columns + ['__stacked_index']))
+        stacked_df = stacked_df.dropna() if dropna else stacked_df
+
+        return stacked_df.all_series['__stacked']
+
 
 def dict_name_series_equals(a: Dict[str, 'Series'], b: Dict[str, 'Series']):
     """
--- a/bach/bach/dataframe.py
+++ b/bach/bach/dataframe.py
@@ -2775,12 +2775,12 @@ class DataFrame:
         """
         df = self.copy()
         if df.group_by:
-            df = df.materialize()
+            df = df.materialize('stack')
 
         dc_dfs = []
         # convert each data column series to DataFrame and use series name as new index value
-        for series_name, series in df._data.items():
-            dc_df = df.all_series[series_name].copy_override(name='__stacked').to_frame()
+        for series_name, series in df.data.items():
+            dc_df = series.copy_override(name='__stacked').to_frame()
             dc_df['__stacked_index'] = series_name
             dc_dfs.append(dc_df)
 
--- a/bach/bach/dataframe.py
+++ b/bach/bach/dataframe.py
@@ -2768,8 +2768,8 @@ class DataFrame:
             at least an index series, this might generate different combinations between
             the index and the stacked values.
 
-        :return: a reshaped series that includes a new index containing the caller's column labels as values.
-
+        :return: a reshaped series that includes a new index (named "__stacked_index")
+            containing the caller's column labels as values.
         .. note::
             ``level`` parameter is not supported since multilevel columns are not allowed.
         """
--- a/bach/bach/dataframe.py
+++ b/bach/bach/dataframe.py
@@ -1560,7 +1560,7 @@ class DataFrame:
         :returns: a new DataFrame with the specified ordering,
          otherwise it updates the original and returns None.
         """
-        sort_by = self.index_columns if not level else self._get_indexes_by_level(level)
+        sort_by = self.index_columns if level is None else self._get_indexes_by_level(level)
         df = self.sort_values(by=sort_by, ascending=ascending)
         return df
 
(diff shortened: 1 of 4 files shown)


bach/docs/source/DataFrame.rst
```diff
@@ -162,6 +162,7 @@ Reshaping, indexing, sorting & merging
     DataFrame.fillna
     DataFrame.ffill
     DataFrame.bfill
+    DataFrame.stack
```
jansentom, line 165: Hey, we're missing unstack here!

bach/bach/dataframe.py
```diff
@@ -2760,6 +2760,41 @@ def _get_parsed_subset_of_data_columns(
...
+
+        df = self.copy()
+        if df.group_by:
+            df = df.materialize()
+
+        dc_dfs = []
+        # convert each data column series to DataFrame and use series name as new index value
+        for dc in df.data_columns:
+            dc_df = df.all_series[dc].copy_override(name='__stacked').to_frame()
+            dc_df[f'__stacked_index'] = dc
```
jansentom, line 2785:
```suggestion
            dc_df['__stacked_index'] = dc
```

bach/bach/dataframe.py
```diff
@@ -2760,6 +2760,41 @@ def _get_parsed_subset_of_data_columns(
...
+            dc_dfs.append(dc_df)
+
+        # concat all dataframes to get new_index and stacked values in two single series
+        from bach.operations.concat import DataFrameConcatOperation
+        stacked_df = DataFrameConcatOperation(dc_dfs)()
+
+        # append the stacked index to the initial indexes
+        stacked_df = stacked_df.set_index(list(self.index_columns + ['__stacked_index']))
```
jansentom, line 2793: If  `__stacked_index` already existed in the df, this breaks silently. Should we check/avoid that?

bach/bach/dataframe.py
```diff
@@ -2760,6 +2760,41 @@ def _get_parsed_subset_of_data_columns(
...
+
+        :param level: only ``level=-1`` is supported, column axis supports only single levels.
+        :param dropna: Whether to drop rows that contain missing values. If the caller has
+            at least an index series, this might generate different combinations between
+            the index and the stacked values.
+
+        :return: a reshaped series that includes a new index containing the caller's column labels as values.
+        """
+        if level != -1:
+            raise NotImplementedError('column axis supports only one level.')
```
jansentom, line 2775: If there is just one Series (column), stacking does not make a lot of sense. Should we just return a copy of the only series in there in that case?

bach/bach/dataframe.py
```diff
@@ -2760,6 +2760,41 @@ def _get_parsed_subset_of_data_columns(
...
+
+        df = self.copy()
+        if df.group_by:
+            df = df.materialize()
```
jansentom, line 2779: please set a name when materializing, this makes the sql much more easy to understand

bach/bach/dataframe.py
```diff
@@ -2760,6 +2760,41 @@ def _get_parsed_subset_of_data_columns(
...
+    def stack(self, *, level: int = -1, dropna: bool = True) -> 'Series':
+        """
+        Stacks all data_columns into a single index series.
+
+        :param level: only ``level=-1`` is supported, column axis supports only single levels.
+        :param dropna: Whether to drop rows that contain missing values. If the caller has
+            at least an index series, this might generate different combinations between
+            the index and the stacked values.
+
+        :return: a reshaped series that includes a new index containing the caller's column labels as values.
```
jansentom, line 2772: Can you please elaborate a bit more here on series naming?

bach/tests/functional/bach/test_df_stack.py
```diff
@@ -0,0 +1,56 @@
+import pandas as pd
+import pytest
+
+from tests.functional.bach.test_data_and_utils import get_bt_with_test_data, assert_equals_data
+
+
```
jansentom, line 6: These tests are nice, thanks! I think there should be some tests for null value handling; it hurts my brain (yes, it's sensitive :) to try to figure out what happens.

bach/bach/dataframe.py
```diff
@@ -2760,6 +2760,41 @@ def _get_parsed_subset_of_data_columns(
             )
         return subset
 
+    def stack(self, *, level: int = -1, dropna: bool = True) -> 'Series':
+        """
+        Stacks all data_columns into a single index series.
+
+        :param level: only ``level=-1`` is supported, column axis supports only single levels.
```
jansentom, line 2767: Shouldn't we remove this parameter and add a note in the docs? It makes no sense for us since we can never allow multilevel columns

jansentom (commented):
A few, but mostly minor comments. Cool to have this functionality!

bach/docs/source/DataFrame.rst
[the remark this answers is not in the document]
KathiaBarahona: We don't have`DataFrame.unstack`.  Was going to do it in this PR, but not as easy as I thought (need to go through Series.unstack to understand better). https://github.com/objectiv/objectiv-analytics/issues/521

bach/bach/dataframe.py
KathiaBarahona: Yes, I actually thought about it... buuuut we don't do it for other functions with "hardcoded internal" names (e.g quantiles). Maybe we should create smth like:
```python
    def _generate_series_name(self, base_name: str) -> str:
        name_regex = re.compile(rf'^{base_name}(_(?P<post_fix_num>\d+))?$')

        post_fix = 0
        for series_name in self._data.keys():
            match = name_regex.match(series_name)
            if not match:
                continue
            curr_post_fix = match.group('post_fix_num')
            post_fix = max(post_fix, 1 if not curr_post_fix else int(curr_post_fix) + 1)

        if not post_fix:
            return base_name

        return f'{base_name}_{post_fix}'
```
But definitely prefer to deal all this situations in a separate issue

bach/bach/dataframe.py
KathiaBarahona: yup, I agree

bach/tests/functional/bach/test_df_stack.py
KathiaBarahona: uhh will try to, didn't wanted to write 50 rows with null values haha

bach/bach/dataframe.py
KathiaBarahona: mmm yes and no, since the series name should be an index value. If we return just a copy, it'll stop being Pandas compatible

bach/bach/dataframe.py
KathiaBarahona: sure

bach/docs/source/DataFrame.rst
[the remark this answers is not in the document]
jansentom: ah doh! Thanks

bach/bach/dataframe.py
jansentom: It makes sense to deal with these cases in a seperate issue indeed. Solution you propose would work I think.

bach/bach/dataframe.py
jansentom: yeah, or just `stack`?

jansentom (approved):

bach/bach/dataframe.py
```diff
@@ -2760,6 +2760,40 @@ def _get_parsed_subset_of_data_columns(
...
+            ``level`` parameter is not supported since multilevel columns are not allowed.
+        """
+        df = self.copy()
+        if df.group_by:
+            df = df.materialize()
+
+        dc_dfs = []
+        # convert each data column series to DataFrame and use series name as new index value
+        for series_name, series in df._data.items():
+            dc_df = df.all_series[series_name].copy_override(name='__stacked').to_frame()
```
thijs-obj, line 2783:
```suggestion
        for series_name, series in df.data.items():
            dc_df = series.copy_override(name='__stacked').to_frame()
```

minor thing. `series` was never used.

bach/bach/dataframe.py
thijs-obj: I think we should make it a rule that any column starting with `__` can be broken by us, without warning.

Only tricky thing is if we return series/dataframes with such names, then users are bound to run into issues. So we probably shouldn't do that then

bach/bach/dataframe.py
```diff
@@ -2760,6 +2760,40 @@ def _get_parsed_subset_of_data_columns(
             )
         return subset
 
+    def stack(self, dropna: bool = True) -> 'Series':
+        """
+        Stacks all data_columns into a single index series.
```
thijs-obj, line 2765: Question: What type will this column have, if the columns are not all of the same type? Should we not raise an exception then? How does pandas handle that?

thijs-obj (commented):
Left a question and some minor comments.

bach/bach/dataframe.py
KathiaBarahona: afff thanks

bach/bach/dataframe.py
KathiaBarahona: will convert them to "string" (ConcatOperation implementation is in charge of this). Pandas just assigns "object" type to the series. I can add a test for this

PR #517, opened 2022-03-02
Bach bq limited support

thijs-obj:
Changes:
* Some tests are now 'multi-database tests'. They take either a `dialect` or `engine` as argument, which can be set to either a Postgres or BigQuery dialect/engine
  * new `pytest` commandline options:
    * `--all`: Run the multi-database tests against both Postgres and BigQuery
    * `--big-query`: Run the multi-database tests against BigQuery
  * In all cases the non-multi database tests (most tests) run against Postgres 
* `quote_identifier` has an extra argument: dialect. Depending on Postgres or BigQuery identifiers are quoted differently.
  * As a result we have to pass the dialect in a huge number of places. All other changes basically are needed because of this change.

General idea:
* All Postgres code should work always. There should be no dependency on any BigQuery library ever.
* Slowly make more tests multi-database, and make the Bach code multi-database to make those test pass
* In the future we'll also run the BigQuery tests in ci (not now)

Note: To see the BigQuery data (and run the tests locally), you might need some additional access and/or credentials of a service account.

This is ready for review. There are small todos still in one function, but as this is quite big it might be worth having a look at the rest already @jansentom @KathiaBarahona


bach/tests/unit/sql_models/test_util.py
```diff
@@ -20,10 +21,25 @@ def test_extract_format_fields_nested():
...
+        assert quote_identifier(dialect, 'te"st') == '"te""st"'
+        assert quote_identifier(dialect, '"te""st"') == '"""te""""st"""'
+        assert quote_identifier(dialect, '`te`st`') == '"`te`st`"'
+        assert quote_identifier(dialect, 'te%st') == '"te%st"'
+    elif is_bigquery(dialect):
+        # https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#identifiers
+        assert quote_identifier(dialect, 'test') == '`test`'
+        assert quote_identifier(dialect, 'te"st') == '`te""st`'
+        assert quote_identifier(dialect, '"te""st"') == r'`""te""""st""`'
+        # TODO: CHECK, is the implementation of BigQueryDialect correct?? Seems odd
```
thijs-obj, line 37: This I still need to do, but the other 99% of this PR can already be reviewed

bach/sql_models/util.py
```diff
@@ -33,23 +35,37 @@ def extract_format_fields(format_string: str, nested=1) -> Set[str]:
     return extract_format_fields(new_format_string, nested=nested-1)
 
 
-def quote_identifier(name: str) -> str:
+def quote_identifier(dialect: Dialect, name: str) -> str:
```
thijs-obj, line 38: Here is  the 'main' change of this PR

bach/sql_models/util.py
```diff
@@ -33,23 +35,37 @@ def extract_format_fields(format_string: str, nested=1) -> Set[str]:
...
+        # more 'logical' would be: dialect.preparer(dialect).quote_identifier(value=name)
+        # But it seems that goes wrong in case there is a `%` in the value. Which sort of makes sense, as
+        # sqlalchemy already escapes that for later on.
+
+        # postgres spec: https://www.postgresql.org/docs/14/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS
+        replaced_chars = name.replace('"', '""')
+        return f'"{replaced_chars}"'
+
+    if is_bigquery(dialect):
+        # todo: check whether this is efficient and correct
```
thijs-obj, line 65: Ironically, here is a todo that I still need to look into. But the rest of this PR is ready for review

bach/tests/conftest.py
```diff
@@ -0,0 +1,76 @@
...
+from typing import NamedTuple, Optional
+
+from sqlalchemy import create_engine
+from sqlalchemy.engine import Engine, Dialect
+
+DB_PG_TEST_URL = os.environ.get('OBJ_DB_PG_TEST_URL', 'postgresql://objectiv:@localhost:5432/objectiv')
+DB_BQ_TEST_URL = os.environ.get('OBJ_DB_BQ_TEST_URL', 'bigquery://objectiv-snowplow-test-2/bach_test')
+DB_BQ_CREDENTIALS_PATH = os.environ.get(
+    'OBJ_DB_BQ_CREDENTIALS_PATH',
+    '/home/thijs/.work_secret/objectiv-snowplow-test-2_bach-big-query-testing.json'
```
jansentom, line 19: haha, yeah :)

bach/tests/conftest.py
```diff
@@ -0,0 +1,76 @@
...
+)
+
+
+def pytest_addoption(parser):
+    parser.addoption('--postgres', action='store_true', help='run the functional tests for Postgres')
+    parser.addoption('--big-query', action='store_true', help='run the functional tests for BigQuery')
+    parser.addoption('--all', action='store_true', help='run the functional tests for BigQuery')
+
+
+def pytest_generate_tests(metafunc):
```
jansentom, line 29: noice

bach/tests/conftest.py
```diff
@@ -0,0 +1,76 @@
...
+    # Import locally. This way a missing library doesn't break anything, if we don't hit this code path
+    from sqlalchemy.dialects.postgresql.base import PGDialect
+    return EngineDialect(None, PGDialect())
+
+
+def get_bigquery_engine_dialect(need_engine: bool = True) -> EngineDialect:
+    if need_engine:
+        engine = create_engine(DB_BQ_TEST_URL, credentials_path=DB_BQ_CREDENTIALS_PATH)
+        return EngineDialect(engine, engine.dialect)
+    # Import locally. This way a missing library doesn't break anything, if we don't hit this code path
```
jansentom, line 74: Like this!

bach/tests/unit/sql_models/test_sql_generator_materialization.py
```diff
@@ -65,22 +68,26 @@ def test_edge_node_materialization():
...
     expected_query = '''
         select l.key, l.value + r.value as value
         from "ValueModel___0cf6df2c76283647c26101b0e472e617" as l
         inner join "ValueModel___5d92f18dfba41ab7101110d9679b9faf" as r on l.key=r.key
     '''
-
-    # TODO: the naming of these views and tables is something we need to improve
+    expected_table = 'create table "ValueModel___5d92f18dfba41ab7101110d9679b9faf" as select \'a\' as key, 2 as value'
+    expected_view = 'create view "ValueModel___0cf6df2c76283647c26101b0e472e617" as select \'a\' as key, 1 as value'
+    if is_bigquery(dialect):
```
jansentom, line 80: nasty :(

jansentom (commented):
Looks like a good start. I like Dialect vs Engine; good choice!

bach/tests/unit/sql_models/test_sql_generator_materialization.py
thijs-obj: yeah this is not nice at all.

For unittests like these we need to do some hacks and/or duplicating the implementation. The Functional tests take a long time to run, so it's not realistic to get very good test coverage there. So we need all the unit-tests that we can to get some okayish test-coverage

bach/tests/unit/sql_models/test_util.py
thijs-obj: This will actually be a bit bigger thing. BigQuery has very different requirements for column names than Postgres. So we might actually have to limit what we allow throughout Bach, or at least when the dialect is BigQuery. Let's discuss in person

thijs-obj: https://cloud.google.com/bigquery/docs/schemas#column_names aliasses (i.e. `"col as "x"`) seem to have the same limitations

bach/requirements-bigquery.txt
```diff
@@ -0,0 +1,6 @@
+# Later on we'll create a separate packages `bach` and `bach[bigquery]`. `bach[bigquery]` will then have
```
thijs-obj, line 1: at first this was in `requirements-dev.txt`, but then I realized that we use that file for the CI tests. And the CI tests only run Postgres tests, so they should break if we accidentally hit a bigquery code-path

thijs-obj:
I think @KathiaBarahona and @hendrik-obj had no comments. So I am merging this.

bach/tests/conftest.py
thijs-obj: Smallest PR ever: https://github.com/objectiv/objectiv-analytics/pull/527

PR #518, opened 2022-03-02
fix non-deterministic sorting in tests

thijs-obj:
In most tests we didn't force a certain sorting. This PR makes it so that `assert_equals_data()` always sorts the data. Forcing a sorting changed the sorting of result data for some tests with Postgres; those tests are fixed too.

This should get rid of a lot of non-determinism in tests, but probably doesn't fix everything.

### DIFF
--- a/bach/tests/functional/bach/test_data_and_utils.py
+++ b/bach/tests/functional/bach/test_data_and_utils.py
@@ -207,6 +207,8 @@ def assert_equals_data(
 
     if order_by:
         bt = bt.sort_values(order_by)
+    elif not bt.order_by:
+        bt = bt.sort_index()
 
     if not use_to_pandas:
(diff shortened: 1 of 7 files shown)


bach/tests/functional/bach/test_data_and_utils.py
```diff
@@ -207,6 +207,8 @@ def assert_equals_data(
 
     if order_by:
         bt = bt.sort_values(order_by)
+    elif not bt.order_by:
```
thijs-obj, line 210: not sure why we didn't have this before

KathiaBarahona (approved):
nice! lgtm

thijs-obj:
Also changed a modelhub test, but not expecting any comments, so I am going to merge this.
Let me know if you have any comments @hendrik-obj

Issue #524, opened 2022-03-07
Bug Report: `SeriesUuid.min()` broken


thijs-obj:
**Describe the bug**
* `SeriesUuid.min()` (and probably `max()` etc.) generates invalid SQL.
* This also breaks `describe(include='all')` on DataFrames with UUID columns

**Steps To Reproduce**
1. create dataframe with uuid columns
2. call `dataframe.describe().to_pandas()`

**Expected behavior**
* `SeriesUuid.min()` should raise an `NotImplementedError` as Postgres doesn't support min() on uuid columns. (alternative: magic to get it to to work, not impossible, probably useless).
* `df.describe(include='all')` should work on DataFrames with uuid columns. For the uuid columns we should not get a 'min' value.

**Additional context**
A user encountered this problem in the wild

PR #525, opened 2022-03-07
fix describe uuid

KathiaBarahona:
Raise `NotImplemented` for `min` and `max` in UUID series


thijs-obj:
This resolves #524

thijs-obj:
lgtm

PR #532, opened 2022-03-10
removing objectivframe

hendrik-obj:


jansentom (commented):
Overall: very nice improvement; nothing was lost while a lot of simplicity was gained! 
I dod find it a bit confusing that we call the DataFrame `data` is quite a few places, why not use `df`?

hendrik-obj:
> Overall: very nice improvement; nothing was lost while a lot of simplicity was gained! I do find it a bit confusing that we call the DataFrame `data` in quite a few places, why not use `df`?

Used `df` before, but I thought might be a bit confusing because it is very pandas, and it already says it needs a Bach DataFrame. happy to revert it to `df` though. lmk @jansentom

PR #533, opened 2022-03-11
Save mapping between columns and expressions in `BachSqlModel`

KathiaBarahona:
Save mapping between columns and expressions in BachSqlModel. Needed for on series boolean merge


jansentom (approved):
LGTM, thanks for the simple PR

PR #534, opened 2022-03-11
Add `TableColumnReferenceToken` to know when a table is referenced by a column

KathiaBarahona:
Created TableColumnReferenceToken, this way we can know the table being references by a column.


bach/bach/expression.py
```diff
@@ -282,18 +299,49 @@ def has_windowed_aggregate_function(self) -> bool:
...
-                result.append(data_item.resolve(dialect, table_name))
+                result.append(data_item.resolve(table_name))
             else:
                 result.append(data_item)
         return self.__class__(result)
 
+    def remove_table_column_references(self) -> Tuple[str, str, 'Expression']:
+        """
+        removes all table references from this expression.
+        Returns first table_name and column_name found and a new expression without table column references
```
jansentom, line 326: Should it not return as list of (table_name, column_name) tuples? The first one is quite random :)

jansentom (approved):
One minor comment, rest LGTM

bach/bach/expression.py
KathiaBarahona: I thought about returning a mapping between table_names and list of columns, buuuuuut make things more complicated and in theory we shouldn't be allowing multiple table refs in an expression.

KathiaBarahona: also, usually the first column found in the expression is the name of the series. It's "hacky" but prefer to call it an heuristic ;)

jansentom: is the "in theory" guarded aka do we raise when that happens? Maybe we should check in this method that it's actually u
unique or raise, that'd be nice?

KathiaBarahona: nope, :thinking: the only place where we set the table alias is in "merge" and it's hardcoded. Not sure, in that case probably we will need to "preprocess" the expressions in the SqlModel

KathiaBarahona: but will raise it here :P

thijs-obj: I don't think it makes much sense to raise the exception only here. This is an okay second place to do it. But I think that if we say that `"we shouldn't be allowing multiple table refs in an expression"`, then we should raise that exception when we create such an illegal `Expression`.

thijs-obj: Also, in a merge could we not have an expression that references two different tables, or am I misunderstanding that?

thijs-obj: I do agree that returning a list or map of table-names wouldn't be very useful for code that calls this probably.

PR #540, opened 2022-03-14
WIP: Fix metabase startup order

borft:


docker-compose.yaml
```diff
@@ -43,6 +43,7 @@ services:
       - "127.0.0.1:8081:8081"
     networks:
       - obj
+    restart: unless-stopped
```
borft, line 46: set a default policy to restart, unless the container was stopped manually, to recover from fails

analysis/docker/entrypoint.sh
```diff
@@ -1,17 +1,26 @@
 #!/bin/bash
 
-# load virtualenv
-source /services/venv/bin/activate
+set -e
+
 
 # wait a bit to give PG a chance to start up
-sleep 10
+until PGPASSWORD=$POSTGRES_PASSWORD psql -h "${POSTGRES_HOSTNAME}" -U "${POSTGRES_USER}" -c '\q'; do
```
borft, line 7: we wait until the db is accessible, rather than just 10s, so in most cases we actually start-up faster

docker-compose.yaml
```diff
@@ -79,9 +81,9 @@ services:
     ports:
       - "127.0.0.1:8888:8888"
       - "127.0.0.1:8053:8053"
-
     networks:
       - obj
+    restart: unless-stopped
     depends_on:
```
borft, line 87: even though we depend_on the notebook container, docker doesn't wait for it to be ready, so this dependency doesn't really achieve what we want (to have everything needed up, running and bootstrapped, before we start up the metabase container)

analysis/docker/entrypoint.sh
```diff
@@ -1,17 +1,26 @@
...
+  >&2 echo "Postgres is unavailable - sleeping"
```
jansentom, line 8:
```suggestion
  >&2 echo "Postgres is not yet available - sleeping 1s"
```

jansentom (approved):

docker-compose.yaml
jansentom: Can't test it at the moment because I'm doing a repo reshuffle locally, but what happens if the user stops metabase, then stops all, and then restarts all. Does MB stay off in that case (horrible troubleshooting ahead)

borft: no. it's not that complicated ;)
1. stop metabase, which means everything is running but metabase
2. stop all --> everything is now stopped
3. start all --> everything is running

PR #542, opened 2022-03-14
Revert Merge Implementation 

KathiaBarahona:
**Description**

One of the advantages of Bach over Pandas, is that Bach objects preserve a history of all the nodes that are generated after each operation. Therefore it will be nice to be able to revert some of this operation, such as Merge. By reverting merge operation, we can internally be able to update the node's query with further information that the user requires, this way we can avoid the creation of nested subqueries and be more efficient.

For example:
`df['a'] + df2['b'] + df3['c']` will current generate the following:

``` sql
with join_1 as (SELECT "l"."a", "r"."b" from "df1" OUTER JOIN ON ... ),
first_result as (SELECT "a" + "b" AS "a" from "join_1"),
join_2 as (SELECT "l"."a", "r"."c" from "first_result" OUTER JOIN ON ... )
SELECT "a" + "c" from "join_2"
```

If we are able to revert the merge created during `df['a'] + df2['b']` when observing ` + df3['c']`, bach will be able to know that another column from `df2` is required, this way it can just update `join_1` subquery by just adding `c`. Resulting:
``` sql
with join_1 as (SELECT "l"."a", "r"."b", "r"."c" from "df1" OUTER JOIN ON ... ),
SELECT "a" + "b" +  "c" from "join_1"
```

### DIFF
--- a/bach/bach/merge.py
+++ b/bach/bach/merge.py
@@ -335,8 +335,8 @@ def _get_merge_sql_model(
     join_type_expr = Expression.construct('full outer' if how == How.outer else how.value)
 
     return MergeSqlModel.get_instance(
+        column_expressions={rc.name: rc.expression for rc in new_column_list},
         dialect=dialect,
-        column_names=tuple(rc.name for rc in new_column_list),
         columns_expr=columns_expr,
         join_type_expr=join_type_expr,
         on_clause=on_clause,
@@ -359,18 +359,20 @@ def _get_expression(df_series: DataFrameOrSeries, label: str) -> Expression:
 
 class MergeSqlModel(BachSqlModel):
     @classmethod
-    def get_instance(cls,
-                     *,
-                     dialect: Dialect,
-                     column_names: Tuple[str, ...],
-                     columns_expr: Expression,
-                     join_type_expr: Expression,
-                     on_clause: Expression,
-                     left_node: BachSqlModel,
-                     right_node: BachSqlModel,
-                     variables: Dict['DtypeNamePair', Hashable]) -> 'MergeSqlModel':
+    def get_instance(
+        cls,
+        *,
+        dialect: Dialect,
+        column_expressions: Dict[str, Expression],
+        columns_expr: Expression,
+        join_type_expr: Expression,
+        on_clause: Expression,
+        left_node: BachSqlModel,
+        right_node: BachSqlModel,
+        variables: Dict['DtypeNamePair', Hashable],
+    ) -> 'MergeSqlModel':
         """
-        :param column_names: tuple with the column_names in order
+        :param column_expressions: mapping between column_names and their expressions (in order)
         :param columns_expr: A single expression that expresses projecting all needed columns from either
             left or right
         :param join_type_expr: expression expressing the join type, e.g. an expression that represents
@@ -404,5 +406,5 @@ class MergeSqlModel(BachSqlModel):
             references=references,
             materialization=Materialization.CTE,
             materialization_name=None,
-            columns=column_names
+            column_expressions=column_expressions,
         )
--- a/bach/bach/merge.py
+++ b/bach/bach/merge.py
@@ -1,20 +1,20 @@
 """
 Copyright 2021 Objectiv B.V.
 """
-from collections import defaultdict
 from copy import copy
 from enum import Enum
-from typing import Union, List, Tuple, Optional, Dict, Set, NamedTuple, Hashable, cast
+from typing import Union, List, Tuple, Optional, Dict, Set, Hashable, cast
 
 from sqlalchemy.engine import Dialect
 
 from bach import DataFrameOrSeries, DataFrame, ColumnNames, Series
 from bach.dataframe import DtypeNamePair
-from bach.expression import Expression, join_expressions, ColumnReferenceToken, TableColumnReferenceToken, \
+
+from bach.expression import Expression, join_expressions, TableColumnReferenceToken, \
     NonAtomicExpression
 from bach.utils import ResultSeries, get_result_series_dtype_mapping
 from sql_models.model import Materialization, CustomSqlModelBuilder
-from bach.sql_model import BachSqlModel, get_variable_values_sql, filter_variables, construct_references
+from bach.sql_model import BachSqlModel, construct_references
 
 
 class How(Enum):
@@ -306,10 +306,13 @@ def merge(
 def revert_merge(base: DataFrame) -> Tuple[DataFrame, DataFrame]:
     """
     Splits a merged dataframe into two new frames that have the same structure as the original frames
-    that were combined.
+    that were combined. Base dataframe should not be aggregated since original indexes can be lost.
     """
     if not isinstance(base.base_node, MergeSqlModel):
         raise Exception('can only revert merge if DataFrame base_node is MergeSqlModel instance.')
+
+    if base.group_by:
+        raise Exception('cannot revert merge on aggregated frame.')
 
     left_node = cast(BachSqlModel, base.base_node.references['left_node'])
     right_node = cast(BachSqlModel, base.base_node.references['right_node'])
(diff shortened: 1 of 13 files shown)


bach/bach/merge.py
```diff
@@ -302,6 +303,86 @@ def merge(
...
+
+    left_series = {}
+    right_series = {}
+
+    left_node = cast(BachSqlModel, base.base_node.references['left_node'])
+    right_node = cast(BachSqlModel, base.base_node.references['right_node'])
+
+    expressions_to_parse: Dict[str, List[Expression]] = {}
+    for series_name, original_expr in base.base_node.column_expressions.items():
+        if series_name not in series_to_unmerge or not original_expr.has_table_column_references:
```
jansentom, line 350: the last part loses any column that does not have a table column expression (constants only?)

jansentom (commented):
This is waaaay more elegant than I expected. 
Tests are quite nice too, although I do think some tests are lacking though, mainly for unmerging df's that had operations applied after merge, like unmaterialized groupby(), but also simple arithmetic. If find that part of the code a bit hard to follow, but as I understand it, it magically drops / removes those dataframe properties?

bach/bach/merge.py
```diff
@@ -302,6 +303,86 @@ def merge(
...
+            for sub_expr in original_expr.data
+            if isinstance(sub_expr, (Expression, TableColumnReferenceToken))
+        ]
+
+    for series_name, expressions in expressions_to_parse.items():
+        series = base.all_series[series_name]
+        for sub_expr in expressions:
+            table_name, column_name, expr = sub_expr.remove_table_column_references()
+
+            if table_name == 'l':
```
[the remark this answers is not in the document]
KathiaBarahona: I agree, tho we should do it in another PR, since it involves changing a bunch of tests

bach/bach/merge.py
KathiaBarahona: yeeh :/ I only had two options for constants:
1. Ignore them (we don't have a way to know which df originally had it or it's actually a new column from the current df)
2. Add them to both results

I went with the first

bach/tests/functional/bach/test_revert_merge.py
```diff
@@ -0,0 +1,158 @@
...
+    bt1 = get_bt_with_test_data(full_data_set=False)[['city']]
+    bt2 = get_bt_with_test_data(full_data_set=False)[['inhabitants']]
+    merged = bt1.merge(bt2, on='_index_skating_order')
+    merged = merged.groupby('city').sum()
+
+    with pytest.raises(Exception, match=r'aggregated frame'):
+        revert_merge(merged)
+
+
+def _compare_source_with_replicate(original_source: DataFrame, replicate: DataFrame) -> None:
```
thijs-obj, line 149: Question: what is the reason we don't just do the following?

```
assert original_source == replicate
```

Are there properties we do not expect to be the same? A comment here might help

Only problem I can see is that there is something with the `.engine` that might not compare correctly.

bach/bach/merge.py
```diff
@@ -302,6 +303,117 @@ def merge(
...
+    left_index: Dict[str, 'Series'],
+    right_index: Dict[str, 'Series'],
+) -> Tuple[Dict[str, 'Series'], Dict[str, 'Series']]:
+    """
+    Determine which series are on each original dataframe based on the node references
+    from current MergeSqlModel. When identifying each column original source, a new series
+    will be created and replicate the original structure.
+
+    .. note::
+    Series without table column references are added to both final results.
```
thijs-obj, line 349: Question: Is this the only case in which the DataFrames that `revert_merge()` returns are different than the original DataFrames from before the `merge()`? 

Would be good to add this note to the docstring of `revert_merge()`. If I want to us `revert_merge()`, I am not going to read the docstring of the functions it relies on, so I'll miss this one.

bach/bach/merge.py
```diff
@@ -302,6 +303,117 @@ def merge(
...
+    expressions_to_parse: Dict[str, List[Expression]] = {}
+    for series_name, original_expr in base.base_node.column_expressions.items():
+        if series_name not in series_to_unmerge:
+            continue
+
+        if isinstance(original_expr, NonAtomicExpression) or not original_expr.has_table_column_references:
+            expressions_to_parse[series_name] = [original_expr]
+            continue
+
+        expressions_to_parse[series_name] = [
```
thijs-obj, line 367: I don't quite understand how this works. Question: Can you explain this bit to me?

I am gathering that `expressions_to_parse[series_name]` is a list of Expressions. But what I don't understand is:
1. Why can we just leave out any `sub_expr` if it is not an instance of `Expression` or `TableColumnReferenceToken`. Will filtering the tokens like that not change the combined Expression?
2. Below we iterate over the `List[Expression]` that we assign here, (see https://github.com/objectiv/objectiv-analytics/pull/542/files#diff-ba7b48bfc14bb63ec48a4874ddfec28d908a40d9f3e71cce5c851232ed368df6R380). So if there are multiple sub-expressions, only the last one gets assigned to `left_series[column_name]` and/or `right_series[column_name]`, am I understanding this correctly? what happens with the other `sub_expr` parts?

thijs-obj (commented):
Cool idea.

Left a bunch of questions

bach/bach/merge.py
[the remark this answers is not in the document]
thijs-obj: There seems agreement here. Does that follow-up PR already exist, or do we have an issue to make sure this happens?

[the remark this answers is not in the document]
thijs-obj: I am in favor of the constant value approach I think. We now have the knowledge that the left one is called 'l' in sql duplicated 4 times. Seems appropriate to that a constant for that

bach/tests/functional/bach/test_revert_merge.py
KathiaBarahona: Because it's quite impossible to return the same expressions. There's a part we have to "reconstruct" the expressions of each series (which ends up flatten). so:

```
[[RawToken('('), ColumnReference('a'), RawToken('+'), [....]] 
```
Is not the same as:
```
[RawToken('('), ColumnReference('a'), RawToken('+'), ....]

```

Also, properties such as constant values, order bys are lost after materializing, so for example you can have:

```
df1['c'] = 1
df3 = df1.merge(df2)
```
There's no way to know that series ```c```  comes from df1 or df2 or it was actually added directly to df3, since it's just a constant.

bach/bach/merge.py
KathiaBarahona: why not reading it? :) the only thing that ``revert_merge`` does is call this function. Anyways, currently it is the only case that this can happen.

bach/bach/merge.py
KathiaBarahona: I think both these [tests](https://github.com/objectiv/objectiv-analytics/blob/b94467a5fb8d4e11ec4db1167cca38e888c561cd/bach/tests/functional/bach/test_revert_merge.py#L113-L136) might help for comprehension, and probably some comment in the code is needed, since when I was doing this it was a bit clearer for me :woman_shrugging: 

1. The main reason is due to the scenario when merging on shared columns, we need to get rid of the ```COALESCE```  function to get the actual expressions of those columns, for example:
```COALESCE("l"."X", "r"."X" + 1)``` translates into: ```[RawToken('COALESCE'), RawToken('('), TableColumnReference(table_name='l', column_name='X'), RawToken(','), NonAtomicExpression([TableColumnReference(table_name='r', column_name='X), RawToken('+'), ConstValueExpression([...])) ]```. But, in this case we are interested only in the arguments.  I kinda remember considering creating a new ```FunctionExpression``` lass and just calling smth like ```FunctionExpression.get_arguments```, but this meant a lot of changes and felt a bit hacky.

2. Related to 1. In this case parsing the ``COALESCE`` will return 2 subexpressions. 1 for "l" and the other one for "r".


Thanks for this questions. made me realize that the logic is way too bounded to the ```COALESCE``` case, I guess I did this approach because I wanted to avoid smth hacky like (which will break expressions that where actually construct by ```fillna```): 
if ```expression.data[0].raw == ```COALESCE```:
    ...
  

If you have a better idea, let me know

PR #552, opened 2022-03-16
Overload typing of `DataFrame.__get_item__`

KathiaBarahona:
Used typing's [overload](https://docs.python.org/3/library/typing.html#typing.overload) decorator. This way mypy can stop complaining about smth like `df['a']` being a DataFrame.  

Just changed the `df.all_series[..]` in some places for testing purposes, in theory using it is not wrong.


jansentom (approved):
A-ok! Very neat fix

PR #570, opened 2022-03-23
`Dataframe.unstack()`, support multi levels

KathiaBarahona:
Moved `Series.unstack` code to `DataFrame` (most of the operations where using DataFrame's method). 

Also started supporting unstack using other levels besides `level=-1`


bach/bach/dataframe.py
```diff
@@ -2994,6 +2994,62 @@ def stack(self, dropna: bool = True) -> 'Series':
...
+
+                df[new_column_name] = None
+                # previous statement will change dtype to string because value is None
+                df[new_column_name] = df[new_column_name].astype(df[curr_col].dtype)
+                df.loc[df[index_to_unstack] == column, new_column_name] = df[curr_col]
+
+        df = df.groupby(remaining_indexes).aggregate(aggregation)
+        df = df.rename(columns={col: col.replace(f'_{aggregation}', '') for col in df.data_columns})
+
+        df = df.materialize('unstack')
```
KathiaBarahona, line 3048: had to add this materialize since df might be sorted by columns that no longer exist

bach/bach/dataframe.py
```diff
@@ -1334,7 +1334,7 @@ def reset_index(
 
             new_index = {idx: series for idx, series in df.index.items() if idx not in levels_to_remove}
 
-        df._data = {n: s.copy_override(index={}, index_sorting=[]) for n, s in series.items()}
+        df._data = {n: s.copy_override(index=new_index, index_sorting=[]) for n, s in series.items()}
```
jansentom, line 1337: buh, good catch

bach/bach/dataframe.py
```diff
@@ -2999,6 +2999,62 @@ def stack(self, dropna: bool = True) -> 'Series':
...
+        if level >= len(self.index):
+            raise IndexError(f'Too many levels. DataFrame/Series has only {len(self.index)} levels.')
+
+        if type(aggregation) != str:
+            raise TypeError('invalid aggregation method')
+
+        index_to_unstack = self.index_columns[level]
+        values = self.index[index_to_unstack].unique().to_numpy()
+
+        if None in values or numpy.nan in values:
```
jansentom, line 3033: Do we need a test for this?

bach/bach/dataframe.py
jansentom: Should we do that only when that's the case?

bach/bach/dataframe.py
```diff
@@ -2999,6 +2999,62 @@ def stack(self, dropna: bool = True) -> 'Series':
 
         return stacked_df.all_series['__stacked']
 
+    def unstack(
+        self, level=-1, fill_value: Optional[Scalar] = None, aggregation: str = 'max',
```
jansentom, line 3003: Should we support multiple levels, or named levels too like pandas does? Both can be useful imho

bach/bach/dataframe.py
KathiaBarahona: we could, already tried and looks quite nasty IMHO.

bach/bach/dataframe.py
KathiaBarahona: If I'm not wrong, @hendrik-obj tested it when doing ```Series.unstack```. I'd just copied what was already done :)

bach/bach/dataframe.py
KathiaBarahona: uiii, didn't noticed the ```can pass level name``` from the docs. Good catch!

bach/bach/dataframe.py
jansentom: It's kind of nice that we don't materialize when not required. Let's use @thijs-obj as a tiebreaker here. WDYT? (no pressure ;)

bach/bach/dataframe.py
jansentom: Let's test as close to where it's implemented as possible and move this test over in that case.

bach/bach/dataframe.py
jansentom: Did you implement this in another PR? Can't really find it?

jansentom (commented):
Some open ends left, but this should not be a lot of work anymore. Nice functionality!

bach/bach/dataframe.py
thijs-obj: I don't think materializing here is a big deal, so if that's easier let's do that. 

> had to add this materialize since df might be sorted by columns that no longer exist

Makes more sense to add a comment about this in the code, than in this PR imho

bach/bach/dataframe.py
```diff
@@ -2999,6 +2999,62 @@ def stack(self, dropna: bool = True) -> 'Series':
...
+    ) -> 'DataFrame':
+        """
+        Pivot a level of the index labels.
+
+        Returns a(n unsorted) DataFrame with the values of the unstacked index as columns. In case of
+        duplicate index values that are unstacked, `aggregation` is used to aggregate the values.
+
+        DataFrame's index should be of at least two levels to unstack.
+
```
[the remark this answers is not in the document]
thijs-obj: If we can update the docs for Series.unstack() also, then that would be great

bach/bach/dataframe.py
thijs-obj: Didn't fully review this PR. But I think my input was only needed here, right?

bach/bach/dataframe.py
KathiaBarahona: will do in this one, I'm going to pick it up rn

KathiaBarahona:
@jansentom if you have more questions, let's address it in  a different issue

PR #571, opened 2022-03-23
fix order of columns in __getitem__

KathiaBarahona:
DataFrame should respect the order of selected columns in __getitems__ when a list is passed.

### DIFF
--- a/bach/bach/dataframe.py
+++ b/bach/bach/dataframe.py
@@ -1039,7 +1039,12 @@ class DataFrame:
             key_set = set(key)
             if not key_set.issubset(set(self.data_columns)):
                 raise KeyError(f"Keys {key_set.difference(set(self.data_columns))} not in data_columns")
-            selected_data = {key: data for key, data in self.data.items() if key in key_set}
+
+            if isinstance(key, set):
+                selected_data = {col: data for col, data in self.data.items() if col in key_set}
+            else:
+                # should change order of columns in select
+                selected_data = {col: self.data[col] for col in key}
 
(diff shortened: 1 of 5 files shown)


bach/bach/dataframe.py
```diff
@@ -1039,7 +1039,12 @@ def __getitem__(self, key):
             key_set = set(key)
             if not key_set.issubset(set(self.data_columns)):
                 raise KeyError(f"Keys {key_set.difference(set(self.data_columns))} not in data_columns")
-            selected_data = {key: data for key, data in self.data.items() if key in key_set}
+
+            if isinstance(key, set):
+                selected_data = {col: data for col, data in self.data.items() if col in key_set}
+            else:
+                # should change order of columns in select
+                selected_data = {col: self.data[col] for col in key}
```
jansentom, line 1047: I think it would be more easy to read if you'd create a new top-level `if isinstance(key, list)` for this case? It does not use any of the code in the combined one anyway.

jansentom (commented):
One comment, rest makes sense to me

bach/bach/dataframe.py
KathiaBarahona: yeh, but will have to duplicate the same logic of:
```python
    key_set = set(key)
    if not key_set.issubset(set(self.data_columns)):
        raise KeyError(f"Keys {key_set.difference(set(self.data_columns))} not in data_columns")
```

jansentom: yeah, true. That'd be lame. Let's leave it like this.

PR #572, opened 2022-03-23
Support of ``DataFrame.get_dummies`` and ``SeriesString.get_dummies``

KathiaBarahona:
Works similar as `Pandas.get_dummies`, main difference is that we always add a `prefix`, this is mainly because we don't know the values during runtime.

Also not supporting `sparse` (we just cannot)
and not supporting `drop_first` (doesn't make much sense IMHO) @hendrik-obj do you consider we might need it in a future?


bach/bach/dataframe.py
```diff
@@ -3055,6 +3055,88 @@ def unstack(
 
         return df.fillna(value=fill_value)
 
+    def get_dummies(
+        self,
+        prefix: Optional[Union[str, List[str], Dict[str, str]]] = None,
+        prefix_sep: str = '_',
+        dummy_na: bool = False,
+        columns: Optional[List[str]] = None,
+        dtype: str = 'int64',
```
KathiaBarahona, line 3064: should we limit `dtype` to: `string, int64, float64`? Doesn't make much sense allowing other dtypes. Also we can fill nulls with '' in case `dtype==string` (that is what pandas does)

jansentom (approved):
Nice feature, I think it's very well done like this. @hendrik-obj should have a look as well though.

PR #577, opened 2022-03-28
HttpContext: Make remote_address optional

sdirosa:
This refactors how HttpContext remote_address gets factored.

Docs update: https://github.com/objectiv/objectiv.io/pull/362


sdirosa:
@borft I've to finish fixing up the backed tests, but you can already take a look at this change.

I'm going to check if we can make referrer and user_agent also optional, instead of sending empty strings as we do now.

borft (approved):
lgtm

PR #578, opened 2022-03-28
Support ```DataFrame.minmax_scale```

KathiaBarahona:
Similar to #574 . Based on https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.minmax_scale.html#sklearn.preprocessing.minmax_scale

### DIFF
--- a/bach/bach/dataframe.py
+++ b/bach/bach/dataframe.py
@@ -3030,16 +3030,41 @@ class DataFrame:
 
     def minmax_scale(self, feature_range: Tuple[int, int] = (0, 1)) -> 'DataFrame':
         """
-        Scales all numeric series based on a given range. The transformation to be performed per series
-        is as follows:
-        .. code-block:: python
-            feature_std = (feature - min_feature) / (max_feature - min_feature)
+        Scales all numeric series based on a given range.
 
-            # min, max = feature_range
-            scaled_feature = feature_std * (max - min) + min
-
-        :param feature_range: ```tuple(min, max)``` desired range to use for scaling.
+        :param feature_range: ``tuple(min, max)`` desired range to use for scaling.
         :return: DataFrame
+
+        Each transformation per feature is performed as follows:
+
+        .. testsetup:: minmax_scale
+           :skipif: engine is None
+
+           import pandas
+           data = {'index': ['a', 'b', 'c', 'd'], 'feature': [1, 2, 3, 4]}
+           pdf = pandas.DataFrame(data)
+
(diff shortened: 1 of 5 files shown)


bach/tests/functional/bach/test_df_scale.py
```diff
@@ -100,3 +100,62 @@ def test_standard_scale() -> None:
...
     )
+
+
+def test_min_max_scale() -> None:
+    numerical_cols = ['skating_order', 'inhabitants', 'founding']
+    all_cols = ['city'] + numerical_cols
+    bt = get_bt_with_test_data(full_data_set=True)[all_cols]
+    pdf = bt.to_pandas()
+
+    min_so = 1
```
jansentom, line 111: Do we want to support min=11, max=1? If so, we should test, if not, we should add a check on the arguments.

bach/bach/dataframe.py
```diff
@@ -19,7 +19,7 @@
 from bach.from_database import get_dtypes_from_table, get_dtypes_from_model
 from bach.sql_model import BachSqlModel, CurrentNodeSqlModel, get_variable_values_sql
 from bach.types import get_series_type_from_dtype
-from bach.utils import escape_parameter_characters
+from bach.utils import escape_parameter_characters, FeatureRange
```
jansentom, line 22: not used here?

bach/bach/dataframe.py
```diff
@@ -3024,6 +3024,47 @@ def scale(self, with_mean: bool = True, with_std: bool = True) -> 'DataFrame':
         from bach.preprocessing.scalers import StandardScaler
         return StandardScaler(training_df=self, with_mean=with_mean, with_std=with_std).transform()
 
+    def minmax_scale(self, feature_range: Tuple[int, int] = (0, 1)) -> 'DataFrame':
```
jansentom, line 3027: think we should also support this directly on SeriesNumeric.

jansentom (changes requested):
Easy to review as well, thanks. Nice feature to have too. 
This only big concern is the last of support for this on SeriesNumeric. Rest should be simple fixes.

bach/bach/dataframe.py
KathiaBarahona: forgot to remove it, good catch!

bach/tests/functional/bach/test_df_scale.py
KathiaBarahona: Oh no, that's "min skating order", I wrote each min max per column, but in the first assert it's just for range (0, 1) which is the default

jansentom (approved):
awesome

PR #585, opened 2022-03-29
Use category links in Modeling docs

ivarpruijn:
Uses new types of Docusaurus navigation items to directly show the respective overview page for a category, instead of another 'Overview' link in the menu under it.

As an added benefit, this also makes almost all breadcrumb items clickable.


bach/docs/source/index.rst
```diff
@@ -10,7 +10,7 @@ Modeling
     models
 
 .. toctree::
-    :maxdepth: 6
+    :maxdepth: 7
```
ivarpruijn, line 13: Done to generate links in the 'Reference by function' items in the sidebar. Without this, items like 'Attributes and underlying data' don't get any links, as they only have subsections such as 'Axes'.

ivarpruijn (commented):
@jansentom Placed a few comments for relevant things to look at, feel free to ignore the rest.

PR #587, opened 2022-03-30
Use regular properties instead of @property@classmethod for type rela…

thijs-obj:
Small change in the Series class: Instead of using `@property` and `@classmethod` as decorators on `dtype_aliases`, `supported_db_dtype`, and `supported_value_types` just define them as class level properties.

Reasons:
1. This is in line with how we already define `Series.dtype`
2. This makes things a bit simpler for the future and easier for `mypy`. Some `cast()` and `type: ignore` could be removed because of this

This should not change any behaviour. No tests were changed.


KathiaBarahona (approved):
nice!

PR #594, opened 2022-03-31
tiny docs fixes

hendrik-obj:


notebooks/sklearn-example.ipynb
```diff
@@ -51,7 +51,7 @@
     "# instantiate the model hub\n",
     "modelhub = ModelHub(time_aggregation='YYYY-MM-DD')\n",
     "# get the Bach DataFrame with Objectiv data\n",
-    "df = modelhub.get_objectiv_dataframe(start_date='2022-01-04')"
+    "df = modelhub.get_objectiv_dataframe(start_date='2022-02-02')"
```
vincenthoogsteder, line 54: @hendrik-obj shall we just do start_date='2022-03-01' here as well, like we have in the docs? So then we're sure folks always see the same slapping this on the demo data

PR #595, opened 2022-03-31
Support from_table() for BigQuery basic data types

thijs-obj:
Changes:
* In `Series`: instead of a single `supported_db_dtype` have a db_dtype for each supported database
  * For the most basic types we have defined the BigQuery type. Not all are yet supported
* `DataFrame.from_table()` can determine the column types automatically on BigQuery, for the basic types


thijs-obj:
Cleaner PR: https://github.com/objectiv/objectiv-analytics/pull/599

PR #597, opened 2022-03-31
Bach: `method` param for Cut operation

KathiaBarahona:
Added ``method`` param for `CutOperation`.
``pandas``: No changes, just performs the same steps as the current implementation
``bach``: No boundary adjustments are performed, we only create the bins based on the step and just change first/last 
interval bounds to include both lower and upper edges


bach/tests/functional/bach/test_series_value_counts.py
```diff
@@ -103,3 +116,8 @@ def test_value_counts_w_groupby() -> None:
 def test_bins_error() -> None:
     with pytest.raises(ValueError, match='Cannot calculate bins for non numeric series.'):
         get_bt_with_railway_data()['town'].value_counts(bins=4)
+
+
+def test_method_error() -> None:
+    with pytest.raises(ValueError, match=r'"whatever" is not a valid method'):
```
jansentom, line 122: lol

bach/bach/series/series.py
```diff
@@ -1557,6 +1558,13 @@ def value_counts(
         :param ascending: sorts values in ascending order if true.
         :param bins: works only with numeric series, groups values into the request amount of bins
             and counts values based on each range.
+        :param method: Method to use for calculating bin ranges.
+            Supported values:
+
+                - "pandas" (default): Performs bound adjustments based on Pandas implementation.
+
+                - "bach": No bound adjustments are performed. Instead, first interval includes both
```
jansentom, line 1566: Shall we call this "closed" referring to the closed nature of the overarching interval. (or something similar, open to suggestions). pandas is being weird here, and we don't have to define our name here imho. 
The parameter could also be called "interval" maybe? So you get interval="closed" for ours, interval="pandas" for the weird one? Again, open to suggestions that are less generic than method.

KathiaBarahona: mmm can be confused with ``right`` parameter I guess, since that defines the type of boundaries. Maybe a boolean instead (``apply_adjustments``)

PR #598, opened 2022-03-31
Consistently use Postgres specific test url in tests

thijs-obj:
In an earlier PR, I introduced a new DB URL in `conftest.py` to distinguish Postgres from BigQuery (https://github.com/objectiv/objectiv-analytics/pull/517/files#diff-ac4981ffe5414f6a0da82969dd8fea67abe29feb825bbaa985495b5aefb35b23). I forgot to use that consistently tho. This PR changes that.

### DIFF
--- a/bach/tests/functional/bach/test_df_from_pandas.py
+++ b/bach/tests/functional/bach/test_df_from_pandas.py
@@ -5,8 +5,9 @@ import pytest
 import sqlalchemy
 
 from bach import DataFrame
+from tests.conftest import DB_PG_TEST_URL
 from tests.functional.bach.test_data_and_utils import get_pandas_df, TEST_DATA_CITIES, CITIES_COLUMNS, \
-    DB_TEST_URL, assert_equals_data
+    assert_equals_data
 import datetime
 from uuid import UUID
 import pandas as pd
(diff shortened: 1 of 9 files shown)


modelhub/tests_modelhub/functional/modelhub/data_and_utils.py
```diff
@@ -10,7 +10,7 @@
 import sqlalchemy
 import os
 
-DB_TEST_URL = os.environ.get('OBJ_DB_TEST_URL', 'postgresql://objectiv:@localhost:5432/objectiv')
+DB_TEST_URL = os.environ.get('OBJ_DB_PG_TEST_URL', 'postgresql://objectiv:@localhost:5432/objectiv')
```
thijs-obj, line 13: changed modelhub environment variable too, for consistency.

bach/tests/functional/bach/test_df_from_pandas.py
```diff
@@ -65,7 +66,7 @@ def test_from_pandas_table():
 
 def test_from_pandas_table_injection():
     pdf = get_pandas_df(TEST_DATA_INJECTION, COLUMNS_INJECTION)
-    engine = sqlalchemy.create_engine(DB_TEST_URL)
+    engine = sqlalchemy.create_engine(DB_PG_TEST_URL)
```
KathiaBarahona, line 69: isn't it better to have a fixture for this? I think it will be much easier have one and later parameterize it when testing for BigQuery

thijs-obj: What exactly do you mean when you say a fixture?

We have the "magic" `engine` parameter for test functions that work for both postgres and bigquery, and when running the tests with `--big-query` those tests will run against big query. Is that what you mean, or something else?

thijs-obj: Discussed this out-of-band. My goal with this PR was to create the smallest possible PR that would make all follow-up PRs work, while keeping consistency. So that's why I didn't want to structurally change anything.

But I made the change now in the latest commit: 04580fd062479fe00cef6abd2bd8d58c345fe541. Can you have another look @KathiaBarahona

KathiaBarahona (approved):
lgtm

PR #602, opened 2022-03-31
Speed up test_df_setitem: set all columns in a test in one go

thijs-obj:
Speed up of a test: do less queries.

output of `pytest tests/functional/bach/test_df_setitem.py --all`
* before: `28 passed in 61.93s`
* after: `28 passed in 32.58s`

The BigQuery tests are quite slow, so speeding things up a bit now and then is nice as otherwise a lot of time is spent waiting.

### DIFF
--- a/bach/tests/functional/bach/test_df_setitem.py
+++ b/bach/tests/functional/bach/test_df_setitem.py
@@ -17,44 +17,53 @@ from tests.functional.bach.test_data_and_utils import get_bt_with_test_data, ass
     assert_equals_data, CITIES_INDEX_AND_COLUMNS, get_bt_with_railway_data, get_df_with_test_data, run_query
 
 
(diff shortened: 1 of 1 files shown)


bach/tests/functional/bach/test_df_setitem.py
```diff
@@ -17,44 +17,53 @@
     assert_equals_data, CITIES_INDEX_AND_COLUMNS, get_bt_with_railway_data, get_df_with_test_data, run_query
 
 
-def check_set_const(engine, constant, expected_series: Type[Series], expected_pg_db_type: str):
+def check_set_const(engine, constants: list, expected_series: Type[Series], expected_pg_db_type: str):
```
KathiaBarahona, line 20: just to follow consistency
```suggestion
def check_set_const(engine, constants: List[Any], expected_series: Type[Series], expected_pg_db_type: str):
```

KathiaBarahona (approved):
lgtm

PR #606, opened 2022-04-04
fix append in _resolve_merge_expression_references

KathiaBarahona:
Small bug in _resolve_merge_expression_references, this was actually changing the order of tokens. Noticed thanks from this branch https://github.com/objectiv/objectiv-analytics/pull/601 since the expression structure changed for binary operations.

### DIFF
--- a/bach/tests/unit/bach/test_merge.py
+++ b/bach/tests/unit/bach/test_merge.py
@@ -7,7 +7,7 @@ from sqlalchemy.dialects.postgresql.base import PGDialect
 from bach.expression import Expression
 from bach.merge import (
     _determine_merge_on, _determine_result_columns, ResultSeries, merge, How, MergeOn,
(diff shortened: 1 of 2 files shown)


bach/tests/unit/bach/test_merge.py
```diff
@@ -334,6 +334,13 @@ def test_verify_on_conflicts_conditional(dialect) -> None:
         )
 
 
+def test__resolve_merge_expression_reference(dialect) -> None:
```
thijs-obj, line 337: This test also passes with the old code in `bach/bach/merge.py`. Can we change this to fail with the old code and pass with the new code?

KathiaBarahona: hah sorry, forgot to materialize

thijs-obj (approved):
lgtm

PR #608, opened 2022-04-04
Bach BQ: convert more tests

thijs-obj:
Changes: Tests only
* Convert some tests from Postgres only to multi-database
* Convert some postgres-only tests to stop using `get_bt_with_test_data()` as preparation for porting the code to support BQ
* Modified and moved some unit tests from functional to unit test folder

This is all preparation for more BigQuery support in the code. I'd like to not touch tests and actual code too much in the same PR in these cases, as that makes it easier to see there are no regressions


bach/tests/functional/bach/test_series_numeric.py
```diff
@@ -69,34 +75,6 @@ def test_round_integer():
         pd.testing.assert_series_equal(expected2, result2, check_names=False, check_index=False)
 
 
-def test_dataframe_agg_skipna_parameter():
```
thijs-obj, line 72: These tests were modified to use fake data and then moved to the unittest folder, see here: https://github.com/objectiv/objectiv-analytics/pull/608/files#diff-57274a134bb9bf31f41e5532b1899153fec9bd81337b64de68ba54dad32421d0

thijs-obj:
sorry, pushed one more commit to this branch after opening for review. 82d634f

bach/tests/functional/bach/test_df_setitem.py
```diff
@@ -25,10 +25,9 @@ def check_set_const(engine, constants: List[Any], expected_series: Type[Series],
         column_names.append(column_name)
         bt[column_name] = constant
 
-    if is_postgres(engine):
```
KathiaBarahona, line 28: I think we should keep this if

bach/tests/functional/bach/test_data_and_utils.py
```diff
@@ -288,13 +288,21 @@ def assert_postgres_type(
...
+    NOTE: If series.engine is not a Postgres engine, then this function simply returns without doing any
+    asserts!
+
     :param series: Series object to check the type of
     :param expected_db_type: one of the types listed on https://www.postgresql.org/docs/current/datatype.html
     :param expected_series_type: Subclass of Series
     """
+    engine = series.engine
+    if not is_postgres(engine):
+        return
```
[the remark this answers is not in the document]
thijs-obj: With the name of this function containing the word `postgres`, I don't think the chance of fake positives is that big.
This function is used 29 times. If we raise an Exception here, then we'll have to have the `if is_postgres(engine):` in all those places. Do you really think that's cleaner?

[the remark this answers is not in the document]
thijs-obj: discussed out-of-band, leaving as is for now

bach/tests/functional/bach/test_df_setitem.py
thijs-obj: See other discussion https://github.com/objectiv/objectiv-analytics/pull/608#discussion_r843722206

PR #610, opened 2022-04-05
Change merge in `cut` and `qcut`

KathiaBarahona:
Changed old hack for referencing to "range" column in qcut and cut merge. Now merges based on series boolean (still a bit hacky because we don't have contains operator for SeriesNumeric)


thijs-obj (approved):
LGTM. Left one suggestion

PR #612, opened 2022-04-05
Ignore index attribute in `CutOperation`

KathiaBarahona:
For plotting a hist, we need to preserve the index (contains the label of the numerical columns).

### DIFF
--- a/bach/bach/operations/cut.py
+++ b/bach/bach/operations/cut.py
@@ -17,6 +17,7 @@ class CutOperation:
     series: a numerical series to be segmented into bins
     bins: number of equal-width bins the series will be divided into.
     right: indicates if bin ranges should include the rightmost edge (lower bound).
+    ignore_index: if true, original index will be dropped
     include_empty_bins: if true, it will return bins that contain no values in series.
 
     returns a new Numerical Series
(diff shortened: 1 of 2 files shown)


bach/bach/operations/cut.py
```diff
@@ -17,6 +17,7 @@ class CutOperation:
     series: a numerical series to be segmented into bins
     bins: number of equal-width bins the series will be divided into.
     right: indicates if bin ranges should include the rightmost edge (lower bound).
+    ignore_index: if true, original index will be dropped
```
jansentom, line 20: shouldn't this be called drop_index in line with reset_index(..)?

jansentom (approved):
Looks good, but I left a question of variable naming.

bach/bach/operations/cut.py
KathiaBarahona: mmm it's actually in line with append(..) https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.append.html

jansentom: okay, since pandas can't make up its mind, okay for me to stick to this too.

jansentom (approved):

PR #614, opened 2022-04-05
draft: Bach bq groupby changes

thijs-obj:


bach/bach/dataframe.py
```diff
@@ -1545,25 +1543,64 @@ def groupby(
...
         df = self
-        if self._group_by:
+        if df._group_by:
             # We need to materialize this node first, we can't stack aggregations (yet)
-            df = self.materialize(node_name='nested_groupby')
+            df = df.materialize(node_name='nested_groupby')
 
         group_by: GroupBy
         if isinstance(by, tuple):
+            if not is_postgres(self.engine):
```
thijs-obj, line 1552: All these `if is_postgres()`  and `if is_bigquery()` tend to get a bit ugly. But I suggest we'll just do it this way now, and then when we support 3 or 4 databases we can evaluate whether we need to change how we do this in general or for certain functions

KathiaBarahona: I agree

bach/bach/dataframe.py
```diff
@@ -1545,25 +1543,65 @@ def groupby(
...
+                raise DatabaseNotSupportedException(
+                    self.engine,
+                    f'GroupingSets are not supported for this SQL dialect. Only grouping by one or more '
+                    f'Series is supported. To use this functionality: specify a Series, a list of Series, a '
+                    f'Series name, or a list of Series names. Dialect: {self.engine.name}')
```
KathiaBarahona, line 1557: do we have a test covering this?

thijs-obj: no. Do you think we should?

bach/bach/dataframe.py
```diff
@@ -1545,25 +1543,65 @@ def groupby(
...
 
+        # This is the normal group-by case
+        by_mypy = cast(Union[str, 'Series',
+                             List[DataFrame._GroupBySingleType], None], by)
+        group_by_columns = df._partition_by_series(by_mypy)
+
+        group_by_expressions = any(
+            gbc.expression != Expression.column_reference(gbc.name) for gbc in group_by_columns
+        )
+        if is_bigquery(self.engine) and group_by_expressions:
```
KathiaBarahona, line 1584: can we inverse the logic on this condition? the function is easier to read if we return the normal path.

bach/bach/dataframe.py
```diff
@@ -1545,25 +1543,65 @@ def groupby(
...
-        elif isinstance(by, list) and len([b for b in by if isinstance(b, (tuple, list))]) > 0:
+            return DataFrame._groupby_to_frame(df, group_by)
+
+        if isinstance(by, list) and len([b for b in by if isinstance(b, (tuple, list))]) > 0:
+            if not is_postgres(self.engine):
+                raise DatabaseNotSupportedException(
+                    self.engine,
+                    f'GroupingLists are not supported for this SQL dialect. Only grouping by one or more '
+                    f'Series is supported. To use this functionality: specify a Series, a list of Series, a '
+                    f'Series name, or a list of Series names. Dialect: {self.engine.name}')
```
KathiaBarahona, line 1571: maybe the ``GroupList`` and ```GroupSet`` classes should be in charge of this? Tho, that means we need to provide the engine to them.  The thing is this function looks more scarier than it seems

bach/bach/dataframe.py
KathiaBarahona: yes, just for coverage :)

bach/bach/dataframe.py
KathiaBarahona, line 1584: naming was a bit odd
```suggestion
        has_group_by_expressions = any(
            gbc.expression != Expression.column_reference(gbc.name) for gbc in group_by_columns
        )
        if is_bigquery(self.engine) and has_group_by_expressions:
```

bach/bach/dataframe.py
```diff
@@ -1545,25 +1543,65 @@ def groupby(
...
+            #   e.g. df.groupby([df.x >= 0, df.x == 0])
+            #   See https://github.com/objectiv/objectiv-analytics/issues/616
+            df_new = df.copy()
+            gbc_names: List[str] = [gbc.name for gbc in group_by_columns]
+            for gbc in group_by_columns:
+                df_new[gbc.name] = gbc
+            df_new = df_new.materialize(node_name='bq_materialize_before_groupby')
+            group_by_columns = [df_new[gbc_name] for gbc_name in gbc_names]
+            group_by = GroupBy(group_by_columns=group_by_columns)
+            return DataFrame._groupby_to_frame(df_new, group_by)
```
KathiaBarahona, line 1601: maybe? it is already iterating over `group_by_columns` and `_partition_by_series` already does the job of getting group by columns, so just consistency
```suggestion
            gbc_names = []
            for gbc in group_by_columns:
                gbc_names.append(gbc.name)
                df_new[gbc.name] = gbc

            df_new = df_new.materialize(node_name='bq_materialize_before_groupby')
            group_by_columns = df_new._partition_by_series(gbc_names)
            group_by = GroupBy(group_by_columns=group_by_columns)
            return DataFrame._groupby_to_frame(df_new, group_by)
```

bach/bach/dataframe.py
thijs-obj: yeah I think I like your naming better.

`group_by_expression` was supposed to read as 'do we group by an expression', but agreed that that's not clear.

bach/bach/dataframe.py
thijs-obj: Not really an improvement imho. Because if I read this now I have to click through to `_partition_by_series()` and understand what that does to understand this fragment.

However, I don't care about this very strongly and since you think it's better, I will apply this change too

bach/bach/dataframe.py
thijs-obj: If we put the functionality there, then if we ever do support GroupingLists for BigQuery, then all the logic for materializing etc. would also need to go inside the GroupList I'd think. I don't think that's the way to go.

To be honest, I'd be happy if we get rid of this code. Hence my question to (mainly @jansentom) in the PR description: "Do we actually ever use this functionality on Postgres?"

bach/bach/dataframe.py
thijs-obj: Personally I like functions that use the pattern
```
if special condition 1:
    ...
    return
if special condition 2:
    ...
    return
if special condition 3:
    ...
    return
... normal case logic
return
```

better than
```
if special condition 1:
    ...
    return
if special condition 2:
    ...
    return
if not special condition 3:
    ... normal case logic
    return
...
return
```

The first just seems more consistent and predictable than the second.

But again. I want to optimize for readability, and as I already wrote this and thus know the code, I'll apply this suggestion.

thijs-obj: Fixed in 0a26f522970ab6b14d6737e73eb640d55d41eb39

bach/bach/dataframe.py
KathiaBarahona: IMHO, we should get rid of it if it's not being applied. But, yeah @jansentom  what do you think?

thijs-obj:
Discussed out-of-band: Kathia is OK with these changes.

There is one open question, https://github.com/objectiv/objectiv-analytics/pull/614#discussion_r844879927. As Tom is out, I'll merge this now, and we can address that question later @jansentom

bach/bach/dataframe.py
jansentom: These are all nice optimizations that can only be done be a DB engine: single pass, multi level aggregations.
They are also quite easy to support if not natively supported (many groupby's concatenated). I'd be in favor of keeping the support in place but also agree that a "not supported" is good enough for now. 
PS BQ should support ROLLUP as well btw, redshift supports nothing

thijs-obj: ack. Let's leave it as is for now.

PR #622, opened 2022-04-06
fix renaming of other in __set_item_with_merge

KathiaBarahona:
When calling `__set_item_with_merge` with an aggregated series, the `merge` call materialized `other.base_node`. Therefore it changed the node and the condition in charged of making reference to the correct column name was not working correctly.

### DIFF
--- a/bach/bach/series/series.py
+++ b/bach/bach/series/series.py
@@ -582,7 +582,7 @@ class Series(ABC):
 
         mod_other_name = other.name
         if (
-            other.base_node == right.base_node
+            (other.base_node == right.base_node or not update_column_references)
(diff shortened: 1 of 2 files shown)


bach/bach/series/series.py
```diff
@@ -582,7 +582,7 @@ def __set_item_with_merge(self, other: 'Series') -> Tuple['Series', 'Series']:
 
         mod_other_name = other.name
         if (
-            other.base_node == right.base_node
+            (other.base_node == right.base_node or not update_column_references)
```
KathiaBarahona, line 585: should always modify name if its not updating column references in expression.

thijs-obj (approved):

PR #625, opened 2022-04-06
Bach check valid column names

thijs-obj:
## Changes
* main change: in `Series.__init__()` we check that the `name` is something that that database allows for column names
* supporting changes:
  * added `is_valid_column_name()` function
  * added/updated tests

## Impact
Creating a series with a 'wrong' name now leads to an error directly, instead of when the sql runs.

### DIFF
--- a/bach/tests/unit/bach/test_series.py
+++ b/bach/tests/unit/bach/test_series.py
@@ -4,12 +4,13 @@ Copyright 2021 Objectiv B.V.
 from bach import get_series_type_from_dtype
 from bach.expression import Expression
 from bach.partitioning import GroupBy
-from tests.unit.bach.util import get_fake_df
+from tests.unit.bach.util import get_fake_df, FakeEngine
 
 
 def test_equals():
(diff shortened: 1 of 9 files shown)


bach/tests/unit/bach/test_series.py
```diff
@@ -30,59 +31,62 @@ def test_equals():
...
+    engine = left.engine
+    engine_other = FakeEngine(dialect=engine.dialect, url='sql://some_other_string')
+
     int_type = get_series_type_from_dtype('int64')
     float_type = get_series_type_from_dtype('float64')
 
     expr_test = Expression.construct('test')
     expr_other = Expression.construct('test::text')
 
-    sleft = int_type(engine=None, base_node=None, index={}, name='test',
```
thijs-obj, line 39: Nothing really changed for the `Series.equales()` logic (which is what this test tests), but this broke because we now use `engine.dialect` in `Series.__init__()`

KathiaBarahona (approved):

PR #626, opened 2022-04-07
change notebook header

hendrik-obj:


vincenthoogsteder:
small edit, make Bach in capitals to be consistent. LGTM now

PR #627, opened 2022-04-07
Developer tools package

sdirosa:
Follow up of tracker-better-errors-and-tests, where we move most of the validation logic and error logging into their own package.


sdirosa:
Package is pretty much done, some small tweaks to be done. Will work on those in the parent branch.

PR #630, opened 2022-04-07
Bach categorize unittests: mutli-db or db-independent

thijs-obj:
With this PR all unittest (not the functional tests!) are either classified as multi-database or database-independent.

## Changes
There are two main changes in this PR.
1. This first change is a bit of a meta-change. This introduces the `db_independent` mark, that we can use to mark unittests as database independent (`@pytest.mark.db_independent`). Basically this mark means: we checked this test, it doesn't need to support multiple databases. Commits: ce6a85e0fa6813ab1946562e48fd87a30788e1fa, a49303603107fa49d1764c58e840c72c96a9a244, and d7309bba9993cfe63467c1b214869970e0fb2636
3. Changed all unittests, either added the marking, or added `dialect` as a parameter and made them work for both databases. Commits: c43bd50a26023167712e1322a82ee301b96fa778 and 392a6fba663a22572321444159f253a7d4db5567

## All *unit* tests converted
With the combination of the mark and the `--all` option we can now verify that all tests have either been converted, or checked to be database independent. 
```
pytest tests/unit/ -m db_independent
...
41 passed, 72 deselected
```
that is, there are `41` db-independent tests, and `72` multi-database tests.
So if we run the tests for both BigQuery and Postgres we expect in total there to be `41 + 2 * 72 = 185` tests
```
pytest tests/unit/ --all
...
185 passed
```

### DIFF
--- a/bach/Makefile
+++ b/bach/Makefile
@@ -3,7 +3,7 @@
 tests:
 	mypy --check-untyped-defs bach sql_models
 	pycodestyle bach sql_models
-	pytest tests/ -W error::DeprecationWarning
+	pytest tests/
 
 
 tests-bigquery:
--- a/bach/Makefile
+++ b/bach/Makefile
@@ -3,7 +3,7 @@
 tests:
 	mypy --check-untyped-defs bach sql_models
 	pycodestyle bach sql_models
-	pytest tests/
+	pytest tests/unit tests/functional tests/
 
 
 tests-bigquery:
@@ -12,7 +12,7 @@ tests-bigquery:
 # against Postgres
 	mypy --check-untyped-defs bach sql_models
 	pycodestyle bach sql_models
-	pytest --big-query tests/unit tests/functional -W error::DeprecationWarning
+	pytest --big-query tests/unit tests/functional tests/
 
 
 tests-all:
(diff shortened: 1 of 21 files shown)


thijs-obj:
sorry, pushed one more commit after marking ready for review. It's very small tho

bach/Makefile
```diff
@@ -3,7 +3,7 @@
 tests:
 	mypy --check-untyped-defs bach sql_models
 	pycodestyle bach sql_models
-	pytest tests/ -W error::DeprecationWarning
+	pytest tests/unit tests/functional tests/
```
thijs-obj, line 6: The ` -W error::DeprecationWarning` is now in `pytest.ini`
Listing `tests/unit` first makes sure the fast unittests run before the slow functional tests, and finally `tests/` make sure we won't forget any tests if we ever add more directories

bach/tests/unit/sql_models/test_util.py
```diff
@@ -1,9 +1,13 @@
...
 Copyright 2021 Objectiv B.V.
 """
+import pytest
+
 from sql_models.util import extract_format_fields, quote_identifier, quote_string, is_postgres,\
-    is_bigquery, DatabaseNotSupportedException
+    is_bigquery
+
 
+pytestmark = [pytest.mark.db_independent]  # mark all tests here as database independent.
```
KathiaBarahona, line 10: nice, didn't know this was possible

KathiaBarahona (approved):
lgtm! I really like this PR:)  really cool theres the option of running only db independent tests

jansentom (approved):
LIKE!

PR #632, opened 2022-04-08
Revert Merge: Improve _determine_series_per_source

KathiaBarahona:
Added `merge_on` property to `MergeSqlModel`. This way we know which columns where "coalesced", this way parsing the expression is easier and less hacky.

### DIFF
--- a/bach/bach/merge.py
+++ b/bach/bach/merge.py
@@ -427,18 +427,28 @@ def _determine_series_per_source(
     right_node = cast(BachSqlModel, base.base_node.references['right_node'])
 
     expressions_to_parse: Dict[str, List[Expression]] = {}
+
+    merge_node = cast(MergeSqlModel, base.base_node)
+    merge_on = merge_node.merge_on
+
+    coalesced_series = [
+        left_on for left_on, right_on in zip(merge_on.left, merge_on.right) if left_on == right_on
+    ]
     for series_name, original_expr in base.base_node.column_expressions.items():
         if series_name not in series_to_unmerge:
             continue
 
-        if isinstance(original_expr, NonAtomicExpression) or not original_expr.has_table_column_references:
+        if series_name not in coalesced_series:
             expressions_to_parse[series_name] = [original_expr]
             continue
 
+        # removes first-level tokens from COALESCE expression
+        # I.e COALESCE(expr1, expr2)
+        # expression_to_parse[series] = [expr1, expr2]
         expressions_to_parse[series_name] = [
-            sub_expr if isinstance(sub_expr, Expression) else Expression([sub_expr])
+            sub_expr
             for sub_expr in original_expr.data
-            if isinstance(sub_expr, (Expression, TableColumnReferenceToken))
+            if isinstance(sub_expr, Expression)
         ]
 
     default_series = base.all_series[base.data_columns[0]]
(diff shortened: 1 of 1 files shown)


bach/bach/merge.py
```diff
@@ -606,9 +617,14 @@ def _get_expression(df_series: DataFrameOrSeries, label: str) -> Expression:
 
 
 class MergeSqlModel(BachSqlModel):
-    def __init__(self, *args, **kwargs):
+    def __init__(self, merge_on: MergeOn, *args, **kwargs):
```
thijs-obj, line 620: We need to implement `copy_override()` too when we add parameters here

bach/bach/merge.py
```diff
@@ -427,18 +427,28 @@ def _determine_series_per_source(
...
 
+        # removes first-level tokens from COALESCE expression
+        # I.e COALESCE(expr1, expr2)
+        # expression_to_parse[series] = [expr1, expr2]
         expressions_to_parse[series_name] = [
-            sub_expr if isinstance(sub_expr, Expression) else Expression([sub_expr])
+            sub_expr
             for sub_expr in original_expr.data
-            if isinstance(sub_expr, (Expression, TableColumnReferenceToken))
+            if isinstance(sub_expr, Expression)
```
thijs-obj, line 451: I don't get why this is correct.

Let's say `original_expr` is something like `COALESCE(((l.a + 1) * l.a), r.a + 2))`

1. Then the subexpressions we are interested in are `((l.a + 1) * l.a)`  and `r.a. + 2`, correct?
2. Then `isinstance(sub_expr, Expression)` would filter out the `(`, as it is a token and not an expression, correct?
3. So we would get the expresison `(l.a + 1) * l.a` instead of  `((l.a + 1) * l.a)`, correct?
4. Would the above not be a problem?


Am I misunderstanding what is happening here?

thijs-obj (commented):
Left a question and a comment

bach/bach/merge.py
KathiaBarahona: actually it should filter only `COALESCE`, `(`, `,`,`)`. The expression tree is:

```
Main Expression
├── RawToken
│   └── COALESCE
├── RawToken
│   └── (
├── NonAtomicExpression
│   ├── RawToken
│   │   └── (
│   ├── NonAtomicExpression
│   │   ├── RawToken
│   │   │   └── (
│   │   ├── TableColumnReferenceToken
│   │   │   └── "l"."a"
│   │   ├── RawToken
│   │   │   └── +
│   │   ├── ConstValueToken
│   │   │   └── 1
│   │   └── RawToken
│   │       └── )
│   ├── RawToken
│   │   └── *
│   ├── TableColumnReferenceToken
│   │   └── "l"."a"
│   └── RawToken
│       └── )
├── RawToken
│   └── ,
├── NonAtomicExpression
│   ├── TableColumnReferenceToken
│   │   └── "r"."a"
│   ├── RawToken
│   │   └── +
│   └── ConstValueToken
│       └── 2
└── RawToken
    └── )
```

It should always be this way since both args from the COALESCE come directly from each series expression (Root is always an expression class).

thijs-obj: ok. this makes sense to me

Thanks for the clear explanation :+1:

thijs-obj (approved):

PR #636, opened 2022-04-11
Bach bq load data from cte

thijs-obj:
**Main Change:**  `DataFrame.from_pandas()` support for BigQuery when `materialization='cte'`.  Note that `materialization='table'` is NOT yet supported

**Other Changes:**
* `DataFrame.from_pandas()` now complains if the DataFrame's columns are not all unique and strings.
  * changed some tests that used a column named `0`, renamed that column to`'0'`
* Modified some tests for `from_pandas()` and added some
  *  A lot of tests in `test_df_from_pandas.py` have not been made multi-database yet, because they either test `materialization='table'` or they use types that we do not support yet.
* Converted some tests that previously used `get_from_df()` to now use `DataFrame.from_pandas()` directly. This change makes those tests work with BigQuery too. Unfortunately it wasn't trivial to convert all tests, so left others alone for now.

### DIFF
--- a/bach/bach/from_pandas.py
+++ b/bach/bach/from_pandas.py
@@ -4,13 +4,14 @@ Copyright 2021 Objectiv B.V.
 from typing import Tuple, Dict
 
 import pandas
-from sqlalchemy.engine import Engine
+from sqlalchemy.engine import Engine, Dialect
 
 from bach import DataFrame, get_series_type_from_dtype
 from bach.types import value_to_dtype
 from bach.expression import Expression, join_expressions
+from bach.utils import is_valid_column_name
 from sql_models.model import CustomSqlModelBuilder
-from sql_models.util import quote_identifier
+from sql_models.util import quote_identifier, DatabaseNotSupportedException, is_postgres, is_bigquery
 
 
 def from_pandas(engine: Engine,
@@ -58,7 +59,12 @@ def from_pandas_store_table(engine: Engine,
         * append: Insert new values to the existing table.
     """
     # todo add dtypes argument that explicitly let's you set the supported dtypes for pandas columns
-    df_copy, index_dtypes, all_dtypes = _from_pd_shared(df, convert_objects, cte=False)
+    df_copy, index_dtypes, all_dtypes = _from_pd_shared(
+        dialect=engine.dialect,
+        df=df,
+        convert_objects=convert_objects,
+        cte=False
+    )
 
     conn = engine.connect()
     df_copy.to_sql(name=table_name, con=conn, if_exists=if_exists, index=False)
@@ -76,7 +82,7 @@ def from_pandas_ephemeral(
 ) -> DataFrame:
     """
     Instantiate a new DataFrame based on the content of a Pandas DataFrame. The data will be represented
-    using a `select * from values()` query.
+    using a `select * from values()` query, or something similar depending on the database dialect.
 
     Warning: This method is only suited for small quantities of data.
     For anything over a dozen kilobytes of data it is recommended to store the data in a table in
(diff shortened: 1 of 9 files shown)


bach/tests/functional/bach/test_df_drop_duplicates.py
```diff
@@ -1,18 +1,19 @@
...
+def test_df_basic_drop_duplicates(engine) -> None:
     pdf = pd.DataFrame(
         data={
             'a': [1, 1, 2, 3, 4, 4, 5],
             'b': ['a', 'a', 'b', 'c', 'd', 'e', 'e'],
         }
     )
 
-    df = get_from_df('drop_dup_table', pdf)
+    df = DataFrame.from_pandas(engine=engine, df=pdf, convert_objects=True)
```
thijs-obj, line 16: This change is rather important. `get_from_df()` writes a table to the database, whereas `DataFrame.from_pandas()` with `materialization='cte'` (the default) doesn't. This changes this functional test from a test that writes to the database to a read-only test.

Having these tests read-only is much better, because:
* read-only tests can be trivially parallelized
* we currently use a shared BigQuery database for all tests. So writing tests could run into race-conditions, even if we don't parellelize tests, when multiple persons run tests at the same time.
* BigQuery has limits on how often one can write to a table, so writing tests might break if we run them too often dueing a day (https://cloud.google.com/bigquery/quotas#load_job_per_table.long)
* read-only tests save an operation, so they are slightly faster

Of course we'll need some tests that write to the database. But for tests that not strictly need it, we are better of using `materializtion='cte'` for the test data

KathiaBarahona: should we wait for this PR and care about parallelization later?

bach/bach/from_pandas.py
```diff
@@ -108,12 +119,34 @@ def from_pandas_ephemeral(
...
+        #         ('row 1', cast(1234 as bigint), cast(-13.37 as double precision)),
+        #         ('row 2', cast(1337 as bigint), cast(31.337 as double precision))
+        #     ) as t("a", "b", "c")
+        column_names_expr = join_expressions(
+            [Expression.raw(quote_identifier(engine.dialect, column_name)) for column_name in
+             all_dtypes.keys()]
+        )
+        column_names_str = column_names_expr.to_sql(engine.dialect)
+        sql = f'select * from (values \n{all_values_str}\n) as t({column_names_str})\n'
+    elif is_bigquery(engine):
```
jansentom, line 134: nice, these inline comments.

KathiaBarahona (approved):
lgtm, tho I prefer having a separate PR that includes the migration of all tests from ``get_from_df`` => ``DataFrame.from_pandas``

jansentom (commented):
nice stuff!

thijs-obj:
>

> lgtm, tho I prefer having a separate PR that includes the migration of all tests from `get_from_df` => `DataFrame.from_pandas`

I only migrated a small subset of the tests that use `get_from_df`. Migrating the other tests is a bit tricky, and I haven't even started on it. What makes things tricky is this bug that we still haven't fixed: https://github.com/objectiv/objectiv-analytics/issues/458

bach/tests/functional/bach/test_df_drop_duplicates.py
thijs-obj: I think we could start on parallelization already. We just cannot parallelize tests that use `get_from_df()` until after we migrated them all (which is not always trivial, see other comment).

I imagine we'll always have some tests that write to the database, so we need some mechanism to not parallelize such tests anyway. Therefore, I think some tests using `get_from_df()` shouldn't be a show-stopper for parallelization

PR #638, opened 2022-04-12
Kathiabarahona/series string fix tests

KathiaBarahona:
* Speeded up `test_string_slice`. Instead of executing a query per slice, we just execute one containing all slices we need to test.
* Changed the way we quote string constants for bigquery, uses raw strings to preserve escape characters.

### DIFF
--- a/bach/sql_models/util.py
+++ b/bach/sql_models/util.py
@@ -1,6 +1,7 @@
 """
 Copyright 2021 Objectiv B.V.
 """
+import re
 import string
 from typing import Set, Union, Optional
 
@@ -70,7 +71,7 @@ def quote_identifier(dialect: Dialect, name: str) -> str:
     raise DatabaseNotSupportedException(dialect)
 
 
-def quote_string(value: str) -> str:
+def quote_string(dialect_engine: Union[Dialect, Engine], value: str) -> str:
     """
     Add single quotes around the value and escape any quotes in the value.
 
@@ -78,13 +79,22 @@ def quote_string(value: str) -> str:
     See https://www.postgresql.org/docs/14/sql-syntax-lexical.html#SQL-SYNTAX-CONSTANTS
 
     Examples:
-    >>> quote_string("test")
+    >>> from sqlalchemy.dialects.postgresql.base import PGDialect
+    >>> quote_string(PGDialect(), "test")
     "'test'"
-    >>> quote_string("te'st")
+    >>> quote_string(PGDialect(), "te'st")
     "'te''st'"
-    >>> quote_string("'te''st'")
+    >>> quote_string(PGDialect(), "'te''st'")
     "'''te''''st'''"
     """
+
+    if is_bigquery(dialect_engine):
+        # raw string,
+        # Quoted or triple-quoted literals that have the raw string literal prefix (r or R)
+        # are interpreted as raw/regex strings.
+        # See https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#string_and_bytes_literals
+        return f'r"""{value}"""'
+
     replaced_chars = value.replace("'", "''")
     return f"'{replaced_chars}'"
 
--- a/bach/tests/functional/bach/test_series_string.py
+++ b/bach/tests/functional/bach/test_series_string.py
@@ -8,14 +8,14 @@ from tests.functional.bach.test_data_and_utils import get_bt_with_test_data, ass
     get_df_with_test_data
 
 
-def test_from_const(pg_engine):
+def test_from_const(engine):
     # TODO: BigQuery
     a = 'a string'
     b = 'a string\'"\'\' "" \\ with quotes'
     c = None
     d = '\'\'!@&*(HJD☢%'
 
-    bt = get_df_with_test_data(pg_engine)[['city']]
+    bt = get_df_with_test_data(engine)[['city']]
     bt['a'] = a
     bt['b'] = b
     bt['c'] = SeriesString.from_const(base=bt, value=c, name='temp')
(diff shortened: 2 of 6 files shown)


bach/tests/functional/bach/test_series_string.py
```diff
@@ -1,19 +1,21 @@
 """
```
thijs-obj, line 1: Is there a reason why `test_get_dummies()` at the bottom wasn't converted? It seems to work with BigQuery

leaving the comment here, as I cannot comment on unchanged lines

bach/tests/functional/bach/test_series_string.py
```diff
@@ -30,11 +32,10 @@ def test_from_const(pg_engine):
 
 
 def test_string_slice(engine):
-    bt = get_df_with_test_data(engine)
+    bt = get_df_with_test_data(engine)[['city']]
     # TODO: this test passes on BigQuery, but is very slow
```
thijs-obj, line 36: I think you can remove this todo ;)

thijs-obj: very nice speed-up!

bach/tests/functional/bach/test_series_string.py
```diff
@@ -1,19 +1,21 @@
...
+import pandas as pd
+
 from bach import Series, SeriesString, DataFrame
 from tests.functional.bach.test_data_and_utils import get_bt_with_test_data, assert_equals_data, \
     get_df_with_test_data
 
 
-def test_from_const(pg_engine):
+def test_from_const(engine):
     # TODO: BigQuery
```
thijs-obj, line 12: should be resolved

bach/tests/functional/bach/test_series_string.py
```diff
@@ -1,19 +1,21 @@
...
     a = 'a string'
     b = 'a string\'"\'\' "" \\ with quotes'
     c = None
     d = '\'\'!@&*(HJD☢%'
```
thijs-obj, line 17: This string doesn't seem to work: ```e = 'test"""test'```. 

Whole test for convenient copy paste:
```
def test_from_const(engine):
    # TODO: BigQuery
    a = 'a string'
    b = 'a string\'"\'\' "" \\ with quotes'
    c = None
    d = '\'\'!@&*(HJD☢%'
    e = 'test"""test'

    bt = get_df_with_test_data(engine)[['city']]
    bt['a'] = a
    bt['b'] = b
    bt['c'] = SeriesString.from_const(base=bt, value=c, name='temp')
    bt['d'] = SeriesString.from_const(base=bt, value=d, name='temp')
    bt['e'] = e
    assert_equals_data(
        bt,
        expected_columns=['_index_skating_order', 'city', 'a', 'b', 'c', 'd', 'e'],
        expected_data=[
            [1, 'Ljouwert', a, b, c, d, e],
            [2, 'Snits', a, b, c, d, e],
            [3, 'Drylts', a, b, c, d, e]
        ]
    )
```

bach/sql_models/util.py
```diff
@@ -70,21 +71,30 @@ def quote_identifier(dialect: Dialect, name: str) -> str:
     raise DatabaseNotSupportedException(dialect)
 
 
-def quote_string(value: str) -> str:
+def quote_string(dialect_engine: Union[Dialect, Engine], value: str) -> str:
     """
     Add single quotes around the value and escape any quotes in the value.
 
     This is in accordance with the Postgres string notation format, no guarantees for other databses.
     See https://www.postgresql.org/docs/14/sql-syntax-lexical.html#SQL-SYNTAX-CONSTANTS
```
thijs-obj, line 79: I know I forgot to update the description above. But I think we need to have a more database independent description here.


Perhaps something as simple as this?
```suggestion
    Quote and escape string value for the given database dialect.
```

and then move the link to postgresql documentation below in the code as an inline comment?

thijs-obj (changes requested):
Left some comments. I think there's a problem with strings with 3 consecutive double quotes in them

bach/tests/functional/bach/test_series_string.py
KathiaBarahona: oops. forgot to remove comments

bach/tests/functional/bach/test_series_string.py
KathiaBarahona: because it is out of the scope of the PR, I was focused more on speeding the test and fixing the string constants

bach/tests/functional/bach/test_series_string.py
KathiaBarahona: should be ok now

bach/sql_models/util.py
KathiaBarahona: then we should do it for ```quote_identifier```, I just followed the same pattern

bach/tests/functional/bach/test_series_string.py
thijs-obj: ok

bach/sql_models/util.py
thijs-obj: Fair point, I forgot to update  `quote_identifier()`'s docstring after making it multi-database. It still talked about Postgres being a default etc. Created a small PR to fix that: https://github.com/objectiv/objectiv-analytics/pull/642

thijs-obj: P.S. I didn't mean that we should remove the example. We should just update the description to cover both the Postgres and BigQuery implementation, whereas previously it only described the Postgres part of the implementation.

thijs-obj (approved):

PR #642, opened 2022-04-13
Bach: Update quote_identifier() docstring and test

thijs-obj:

PR #645, opened 2022-04-13
Fix quote_identifiers after earlier mess-up

thijs-obj:
Messed something up in https://github.com/objectiv/objectiv-analytics/pull/642 that broke BigQuery tests


KathiaBarahona (approved):
lgtm!

PR #646, opened 2022-04-13
Make tests, that don't work yet on BigQuery, Postgres-only

thijs-obj:


KathiaBarahona (approved):
lgtm, my bad for not testing for bigquery :crying_cat_face:

PR #647, opened 2022-04-13
Improve navigation after waiting for event to be tracked

afroald:
When `waitUntilTracked=true` on an anchor the default browser behavior is cancelled to be able to wait for tracking. Unfortunately the default behavior cannot be resumed. This is an issue for two reasons as far as I've noticed:

1. User selected preferences regarding navigation like behavior on ctrl/cmd+click break.
2. Attributes like `target` and `rel` get ignored. Breaking `rel="noopener noreferrer"` probably [isn't a big deal anymore](https://mathiasbynens.github.io/rel-noopener/) for security, but it's not good either.

This PR proposes a hacky but slightly improved way of resuming navigation after it was cancelled. By creating a new anchor with the same attributes we keep the behavior related to those attributes. Regarding cmd+click, the goal here is to do something that will be right more often than it is wrong, since correct behavior is not possible.

Please let me know if you think this is an improvement. A better solution than this PR would probably be to remove the `waitUntilTracked` option entirely.

### DIFF
--- a/tracker/trackers/react/src/common/factories/makeAnchorClickHandler.tsx
+++ b/tracker/trackers/react/src/common/factories/makeAnchorClickHandler.tsx
@@ -59,7 +59,24 @@ export const makeAnchorClickHandler =
       // Execute onClick prop, if any.
       props.onClick && props.onClick(event);
 
-      // Resume navigation.
-      window.location.href = props.anchorHref;
+      // Resume navigation by duplicating the anchor to preserve behavior tied to attributes like
+      // `target` and `rel`. During tests in a jsdom environment `currentTarget` doesn't get
+      // set so in that case we have to fall back on `target`.
+      const originalAnchor = event.currentTarget || event.target;
+      const newAnchor = document.createElement('a');
+      for (const attribute of originalAnchor.getAttributeNames()) {
+        // We know for sure the attribute exists at this point, so we have to convince TypeScript
+        newAnchor.setAttribute(attribute, originalAnchor.getAttribute(attribute) as string);
+      }
+
+      // Check if the ctrl or meta key was pressed. In this case the user probably wanted to open
+      // the link in a new tab, but we can't know for sure since it depends on platform and user
+      // preferences. We lost proper behavior when we `preventDefault()`ed so the best we can do is
+      // make a guess and hope this is right in most cases.
+      if (event.metaKey || event.ctrlKey) {
+        newAnchor.setAttribute('target', '_blank');
+      }
+
+      newAnchor.click();
     }
   };


sdirosa:
@afroald 
Thanks for the PR!

Just for context. The only goal of `waitUntilTracked` is to attempt to track session-ending / outgoing links, e.g. opening in new tabs or, on mobile, opening new windows. 

It's best-effort at best and is not meant to be used on internal links as, indeed, it breaks some of their normal behavior. 

If `waitUntilTracked` interferes in any way with the normal functionality of the target Application it should not be used. A custom implementation is recommended for those cases. Pretty much as you would do with other trackers like GA. 

---

> Unfortunately the default behavior cannot be resumed.

I agree. Not 100%.
We tried a couple of approaches. The one in place now, which you revised, was the least disruptive and the simplest.

Another one, which is currently used by Browser Tracker, is to clone the Event and re-dispatch it ([see makeClickEventHandler](https://github.com/objectiv/objectiv-analytics/blob/main/tracker/trackers/browser/src/mutationObserver/makeClickEventHandler.ts)). 

This only works with native event handlers though, not with SyntheticEvents. We can't use it for React.

---

> Regarding cmd+click, the goal here is to do something that will be right more often than it is wrong, since correct behavior is not possible.

Not for the moment, we may be able to change this in the future though. I'd be happy to discuss the longer term plan with you if you like. In short, as we rely less on Collector for validation we may be able to use Beacon API for these events. 

As they are outgoing and, practically, the last Event we track before the session ends, we don't really care if we don't know whether the collector received them or not, nor of its response (Beacon is send-and-forget and has no response).

---

> Please let me know if you think this is an improvement. 

Yes. I think this is already better than what we have! 

**Could you make sure to fix the test coverage so it's 100%?**

The relevant tests, implicitly testing this handler, are here: https://github.com/objectiv/objectiv-analytics/blob/main/tracker/trackers/react/tests/TrackedLinkContext.test.tsx#L198-L260

Feel free to create a test suite for makeAnchorClickHandler on its own if it's easier for you.

---

> A better solution than this PR would probably be to remove the waitUntilTracked option entirely.

We have been using this feature for several months and it works good enough for tracking social links and other outgoing ones. Again, it's opt-in. Perhaps we can simply point out its limitations in the docs though. What do you think? Would that be an improvement?

sdirosa (changes requested):
This looks like an improvement to me. Only remark, as mentioned in my other comment, is about the test coverage. Let me know if you need help with that.

afroald:
> Just for context. The only goal of waitUntilTracked is to attempt to track session-ending / outgoing links, e.g. opening in new tabs or, on mobile, opening new windows.
>
> It's best-effort at best and is not meant to be used on internal links as, indeed, it breaks some of their normal behavior.
>
> If waitUntilTracked interferes in any way with the normal functionality of the target Application it should not be used. A custom implementation is recommended for those cases. Pretty much as you would do with other trackers like GA.

I understand, it just seems to me the current implementation doesn't quite match this description. It _always_ interferes by breaking UX to _assure_ tracking succeeds.

Dispatching a clone of the event is a good idea. I tried it earlier and couldn't get it to work with React like you said. With some more persistence it does work which gives a still hacky but nicer solution 🙂 The `nativeEvent` property of synthetic events [seems safe to use](https://reactjs.org/docs/events.html).  Let me know what you think.

By the way, what is a good way to test this branch in an actual website, like objectiv.io?

sdirosa:
> I understand, it just seems to me the current implementation doesn't quite match this description. It _always_ interferes by breaking UX to _assure_ tracking succeeds.

It's just my take on a possible approach to guarantee traceability. If the trade-off is unreasonable I'd say the best bet is for developers to write their own async click/press handlers. After-all, it's opt-in.

> Dispatching a clone of the event is a good idea. I tried it earlier and couldn't get it to work with React like you said. With some more persistence it does work which gives a still hacky but nicer solution 🙂 The `nativeEvent` property of synthetic events [seems safe to use](https://reactjs.org/docs/events.html). Let me know what you think.

I never managed to get it work nicely with React, it always produced unexpected behavior (mainly the events would not propagate to parents and children as expected). 

If you actually have a working PoC please do share it, I'd be curious to see how that functions!

> By the way, what is a good way to test this branch in an actual website, like objectiv.io?

You can use Verdaccio, [as described here](https://github.com/objectiv/objectiv-analytics/blob/main/tracker/CONTRIBUTING.md#building--publishing-packages). 

In short you start up a dockerized version of verdaccio, add a .yarnrc to the target project containing

```"@objectiv:registry" "http://127.0.0.1:4873"```

And then you can simply `yarn version:patch`,  `yarn build` and `yarn publish:verdaccio` in the tracker monorepo.

Installation can be fiddly due to Docusaurus caches, so beware of that. I often have to run a `yarn clear` to flush those.

sdirosa (approved):
Changing target of this to next, as it's still an improvement over the current code. 

We may do more work on this in the future based on feedback (we got none so far), but as it's an opt-in feature, for now it's good enough.

PR #651, opened 2022-04-14
support nulls in DataFrame.from_pandas

KathiaBarahona:
Most of our tests for ``dropna`` and ``fillna`` contain columns with nulls, currently `DataFrame.from_pandas` raises errors when trying to convert such columns.

### DIFF
--- a/bach/tests/functional/bach/test_df_from_pandas.py
+++ b/bach/tests/functional/bach/test_df_from_pandas.py
@@ -300,3 +300,37 @@ def test_from_pandas_types_cte(pg_engine):
             convert_objects=True,
             materialization='cte'
         )
+
+
+def test_from_pandas_columns_w_nulls(engine) -> None:
+    pdf = pd.DataFrame(
+        {
+            "a": ['a', None, 'c'],
+            "b": [np.nan, 1, 2],
+            "c": [pd.NaT, pd.Timestamp("1940-04-25"), pd.NaT],
+            "d": [None, None, None],
+        }
+    )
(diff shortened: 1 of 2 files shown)


bach/tests/functional/bach/test_df_from_pandas.py
```diff
@@ -300,3 +300,37 @@ def test_from_pandas_types_cte(pg_engine):
             convert_objects=True,
             materialization='cte'
         )
+
+
+def test_from_pandas_columns_w_nulls(engine) -> None:
```
thijs-obj, line 305: Additional test:

I think this should pass but it doesn't now. Because the `if df_copy[column].dropna().empty:` check is always run, even if we don't need it.
```
def test_fails(engine):
    pdf = pd.DataFrame(
        {
            "a": ['a', None, 'c'],
            "b": [np.nan, 1, 2],
            "d": [None, None, None],
        }
    )
    pdf['d'] = pdf['d'].astype('string')

    df = DataFrame.from_pandas(
        engine=engine,
        df=pdf,
        convert_objects=True,
        materialization='table',
        name='testtest',
        if_exists='replace'
    )
    pdfx = df.to_pandas()
    print(pdfx)
    print(pdfx.dtypes)
    print(df.dtypes)

``

thijs-obj: otherwise this looks good to me

KathiaBarahona: huh true, seems gotta do the check after the first check

PR #659, opened 2022-04-19
Parallelization using pytest-xdist, engine dialect singleton

KathiaBarahona:
Decided to use [pytest-xdist ](https://pytest-xdist.readthedocs.io/en/latest/distribution.html) since it was the most efficient one that I tried. 

Currently it uses [-n auto](https://pytest-xdist.readthedocs.io/en/latest/distribution.html#:~:text=pass%20-n%20auto%20to%20use%20as%20many%20processes%20as%20your%20computer%20has%20cpu%20cores.%20this%20can%20lead%20to%20considerable%20speed%20ups%2C%20especially%20if%20your%20test%20suite%20takes%20a%20noticeable%20amount%20of%20time.), so it creates workers based on the number of CPU cores available. 

**Results of Parallelization w/ 4 workers**
image

**Results wo/ Parallelization**
image

### DIFF
--- a/bach/Makefile
+++ b/bach/Makefile
@@ -12,7 +12,7 @@ tests-bigquery:
 # against Postgres
 	mypy --check-untyped-defs bach sql_models
 	pycodestyle bach sql_models
-	pytest --big-query tests/unit tests/functional tests/
+	pytest -n 4 --dist loadgroup --big-query tests/unit tests/functional tests/
 
 
 tests-all:
@@ -21,4 +21,4 @@ tests-all:
 # run against Postgres
 	mypy --check-untyped-defs bach sql_models
 	pycodestyle bach sql_models
-	pytest --all tests/unit tests/functional tests/
+	pytest -n 4 --dist loadgroup --all tests/unit tests/functional tests/
--- a/bach/pytest.ini
+++ b/bach/pytest.ini
@@ -4,7 +4,7 @@
 # Pytest commandline parameters that we always set
 # --strict-markers: Fail if a test function has an unknown marker (e.g. is decorated with @pytest.mark.not_existing)
 # -W error::DeprecationWarning: Raise an error if we encounter a DeprecationWarning
-addopts = --strict-markers -W error::DeprecationWarning -nauto
+addopts = --strict-markers -W error::DeprecationWarning
 
 # markers defines our custom markers
 markers =
--- a/bach/pytest.ini
+++ b/bach/pytest.ini
@@ -4,7 +4,7 @@
 # Pytest commandline parameters that we always set
 # --strict-markers: Fail if a test function has an unknown marker (e.g. is decorated with @pytest.mark.not_existing)
 # -W error::DeprecationWarning: Raise an error if we encounter a DeprecationWarning
-addopts = --strict-markers -W error::DeprecationWarning
+addopts = --strict-markers -W error::DeprecationWarning -nauto
 
 # markers defines our custom markers
 markers =
--- a/bach/tests/conftest.py
+++ b/bach/tests/conftest.py
@@ -8,10 +8,12 @@ specified on the commandline, then it will (also) get a BigQuery dialect or engi
 
 Additionally we define a 'pg_engine' fixture here that always return a Postgres engine.
 """
+import logging
 import os
-from typing import NamedTuple, Optional
+from typing import NamedTuple, Optional, Dict
 
 import pytest
+from _pytest.main import Session
 from _pytest.python import Metafunc, Function
 from _pytest.config.argparsing import Parser
 import sqlalchemy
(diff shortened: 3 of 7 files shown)


bach/pytest.ini
```diff
@@ -4,7 +4,7 @@
 # Pytest commandline parameters that we always set
 # --strict-markers: Fail if a test function has an unknown marker (e.g. is decorated with @pytest.mark.not_existing)
 # -W error::DeprecationWarning: Raise an error if we encounter a DeprecationWarning
```
thijs-obj, line 6: what does -nauto do?

bach/tests/conftest.py
```diff
@@ -43,26 +45,54 @@ def pytest_addoption(parser: Parser):
...
+
+    # For more information, please see:
+    # https://pytest-xdist.readthedocs.io/en/latest/distribution.html
+    # https://docs.pytest.org/en/6.2.x/reference.html#pytest.hookspec.pytest_sessionstart
+    load_engine = True
+    if (
+        ('tests/unit' in session.config.args and len(session.config.args) == 1)
+        or session.config.getoption('markexpr') == 'db_independent'
+    ):
+        # create dialect only if we are running unit tests or db independent tests
```
thijs-obj, line 69:
```suggestion
        # create only dialect, if we are running unit tests or db independent tests
```

thijs-obj: Maybe even
```suggestion
        # create only dialect, if we are running unit tests
```

As the db-independent marker just implies that it is a unit-test.

bach/tests/conftest.py
```diff
@@ -43,26 +45,54 @@ def pytest_addoption(parser: Parser):
...
+
+ENGINE_DIALECTS: Dict[str, EngineDialect] = {}
+
+
+def pytest_sessionstart(session: Session):
+    # Initialize ENGINE_DIALECTS per pytest-xdist worker session based on current pytest running config.
+    # This way we avoid the creation of a new engine per test, as this is quite inefficient and might
+    # cause failures due to multiple clients trying to connect to db server.
+
+    # For more information, please see:
```
thijs-obj, line 61: :+1:

bach/tests/conftest.py
```diff
@@ -43,26 +45,54 @@ def pytest_addoption(parser: Parser):
...
+    # https://pytest-xdist.readthedocs.io/en/latest/distribution.html
+    # https://docs.pytest.org/en/6.2.x/reference.html#pytest.hookspec.pytest_sessionstart
+    load_engine = True
+    if (
+        ('tests/unit' in session.config.args and len(session.config.args) == 1)
```
thijs-obj, line 66: Can we not write it as this?
```suggestion
        session.config.args == ['tests/unit']
```

thijs-obj (changes requested):
Looks very epic. _much_ faster when running all tests.

One important comment tho: `-nauto` doesn't work for me. That spins up 8 workers on my laptop. I get `sqlalchemy.exc.OperationalError: (psycopg2.OperationalError) FATAL:  sorry, too many clients already` on some of the tests.
I think we should set the maximum number of connections that Postgres allows higher to solve that. For now we can also limit things to `-n4` maybe? so we can get this merged asap

thijs-obj:
One more thing, that I forgot before: not all of our tests are read-only. We should make sure that all tests that write to the DB run sequentially and/or make sure they have no race conditions. Making sure they run sequentially seems like the simplest option then?

We could mark all such tests with some decorator, or maybe use a specific fixture to mark them (e.g `engine_writable` instead of `engine`). But not sure if pytest-xdist can then make sure they run non-concurrently?

bach/pytest.ini
jansentom: agree, multithreaded testing only on demand makes sense to me too.

jansentom (commented):
Apart from comments made by Thijs, LGTM. Nice improvement that worked out of the box locally! :rocket:

bach/tests/conftest.py
KathiaBarahona: oh yes we can, duh me:)

bach/Makefile
```diff
@@ -12,7 +12,7 @@ tests-bigquery:
 # against Postgres
 	mypy --check-untyped-defs bach sql_models
 	pycodestyle bach sql_models
-	pytest --big-query tests/unit tests/functional tests/
+	pytest -n 4 --dist loadgroup --big-query tests/unit tests/functional tests/
```
thijs-obj, line 15: let's also do this on line 6

bach/tests/functional/bach/test_df_from.py
```diff
@@ -26,6 +27,7 @@ def _create_test_table(engine: Engine, table_name: str):
         conn.execute(sql)
 
 
+@pytest.mark.xdist_group(name="test_table_writers")
```
thijs-obj, line 30: Question: Is it on purpose that we have a different group name here than in `test_savepoints.py`?

thijs-obj (approved):
two comments/questions. But looks good to me generally :ok_hand:

bach/tests/functional/bach/test_df_from.py
KathiaBarahona: yes, they write on different tables. So, I don't see the purpose of waiting for tests that are unrelated if they are not writing on the same table

PR #662, opened 2022-04-21
proto type logistic regression

hendrik-obj:


bach/bach/dataframe.py
```diff
@@ -1205,21 +1205,22 @@ def __set_item_with_merge(self, key: str, value: 'Series'):
...
-
-        # align index names, this way we have all matched indexes in a single series after merge
-        # TODO: replace with value.rename(index={value_index_name: index_name})
-        #  when index renaming is supported
-        aligned_index = {index_name: value.index[value_index_name].copy_override(name=index_name)}
+        aligned_index = {}
+        for i, index_name in enumerate(self.index_columns):
+            value_index_name = list(value.index.keys())[i]
+            if self.index[index_name].dtype != value.index[value_index_name].dtype:
+                raise ValueError('dtypes of indexes at a position should be the same')
```
jansentom, line 1218:
```suggestion
                raise ValueError(f'Non matching dtypes of index series at position {i}: { self.index[index_name].dtype} != {value.index[value_index_name].dtype}')
```

modelhub/modelhub/aggregate.py
```diff
@@ -16,6 +18,84 @@
 GroupByType = Union[List[Union[str, Series]], str, Series, NotSet]
 
 
+class LogisticRegression(LogisticRegression):
```
jansentom, line 21: Wouldn't this be nicer in its own module/file?

modelhub/modelhub/aggregate.py
```diff
@@ -7,7 +7,12 @@
 from bach.series import Series
 from sql_models.constants import NotSet, not_set
 from typing import List, Union, TYPE_CHECKING
-
+import numpy as np
+from sklearn.linear_model import LogisticRegression
+from sklearn.linear_model import LogisticRegression as LogisticRegression_sklearn
+from sklearn.model_selection import StratifiedKFold
+import pandas as pd
+from sklearn.metrics import roc_curve, auc
```
hendrik-obj, line 15: still needs to be converted to bach

bach/bach/dataframe.py
hendrik-obj: did it

modelhub/modelhub/aggregate.py
hendrik-obj: see update

bach/bach/plotting.py
```diff
@@ -66,6 +66,12 @@ def hist(
         # use calculated frequencies as weights, since Pandas will try to recalculate frequencies
         hist_data = pandas.DataFrame(data={col: freq_pdf['bin_edge'] for col in numeric_columns})
         weights = freq_pdf[numeric_columns].to_numpy()
+
+        bin_edges_no_nan = bin_edges[~numpy.isnan(bin_edges)]
```
hendrik-obj, line 70: this is for the case when there is only one value in the data.

bach/bach/series/series_numeric.py
```diff
@@ -331,3 +339,10 @@ def dtype_to_expression(cls, dialect: Dialect, source_dtype: str, expression: Ex
         if source_dtype not in ['int64', 'string']:
             raise ValueError(f'cannot convert {source_dtype} to float64')
         return Expression.construct(f'cast({{}} as {cls.get_db_dtype(dialect)})', expression)
+
+    def get_random_number(self) -> "SeriesFloat64":
```
thijs-obj, line 343: Just a drive-by comment from me, as I happened to spot this. I am not actually reviewing this.


1. This is a bit similar to `bach.series.series_uuid.SeriesUuid.sql_gen_random_uuid()`, I mention that for two reasons:
    1. This function is different as this is a normal function, and `sql_gen_random_uuid()` is a class method. Would be cool to be consistent. So either make this a class method or change sql_gen_random_uuid().
    2. `sql_gen_random_uuid()` has some interesting warning about the series being non-deterministic. Might be good to have similar documentation here
2. `random()` is not a valid function on bigquery. I believe the equivalent is `RAND()` (https://cloud.google.com/bigquery/docs/reference/standard-sql/mathematical_functions#rand)
3. There should be a test for this. Now nothing fails, despite this not working on BigQuery

modelhub/modelhub/models/feature_importance.py
```diff
@@ -0,0 +1,133 @@
...
+    def get_full_results(self):
+        return self._result_dict
+
+    def _set_seed(self, data: bach.DataFrame, seed: float):
+        if not -1 <= seed <= 1:
+            raise ValueError("seed should be between -1 and 1")
+
+        from sql_models.model import SqlModelBuilder
+
+        class SetSeed(SqlModelBuilder):
```
jansentom, line 24: This is better at home in SeriesNumeric I think, leading to:
```df['seed'] = SeriesNumeric.setseed(...)```
We save a materialization that way, and it's a bit nicer imho

File: modelhub/modelhub/models/feature_importance.py [hunk 7]
```diff
@@ -0,0 +1,133 @@
...
+
+        new_base_node = bach.sql_model.BachSqlModel.from_sql_model(
+            sql_model=new_base_node_model,
+            column_expressions=data.base_node.column_expressions,
+        )
+
+        df_with_seed = data.copy_override_base_node(base_node=new_base_node)
+        return df_with_seed
+
+    def _split_data_set_stratified(self, X, y, folds, seed):
```
jansentom, line 47: This thing desperately needs docs :)
We could generalize this a slight bit more and move it to Bach as base functionality imho.

modelhub/modelhub/models/feature_importance.py
```diff
@@ -0,0 +1,133 @@
...
+    def fit(self, X, y, folds=3, seed=None, print_report=False, **kwargs):
+        if seed is None:
+            seed = np.random.random()
+        self._results_dict = {}
+
+        coef = pd.DataFrame(columns=X.data_columns)
+        for i, (X_train, X_test, y_train, y_test) in enumerate(self._get_stratified_folds(X,
+                                                                                          y,
+                                                                                          folds=folds,
+                                                                                          seed=seed)):
```
jansentom, line 91:
```suggestion
        folds = self._get_stratified_folds(X,y, folds=folds, seed=seed) 
        for i, (X_train, X_test, y_train, y_test) in enumerate(folds):
```

modelhub/modelhub/aggregate.py
```diff
@@ -169,3 +168,46 @@ def frequency(self, data: bach.DataFrame) -> bach.SeriesInt64:
         frequency = total_sessions_user.groupby(['session_id_nunique']).aggregate({'user_id': 'nunique'})
 
         return frequency.user_id_nunique
+
+    def create_feature_usage_data_set(self,
```
jansentom, line 172: Could use some docs

modelhub/modelhub/models/logistic_regression.py
```diff
@@ -0,0 +1,57 @@
+"""
+Copyright 2021 Objectiv B.V.
+"""
+from sklearn.linear_model import LogisticRegression as LogisticRegression_sk  # type: ignore
+from modelhub.metrics import Metrics
+
+
+class LogisticRegression(LogisticRegression_sk):
```
jansentom, line 8: I'm not necessarily against this inheritance, but the argument types of the functions are very unclear now. This will break terribly if you try to type the arguments (which we should imho). 
I do like how you are now Objectiv df compatible, but this is not the way I think. A normal warp would be cleaner

modelhub/modelhub/metrics.py
```diff
@@ -0,0 +1,85 @@
+"""
+Copyright 2021 Objectiv B.V.
+"""
+# todo: assumes classes are True and False
+
+
+class Metrics:
```
jansentom, line 7: I think this class here is a bit confusing. Why not do it as sklearn does with methods at toplevel??  
IMHO, the name metrics is also a bit confusing in MH context, but that's sklearn jargon so I understand where it came from.

modelhub/modelhub/aggregate.py
```diff
@@ -169,3 +168,46 @@ def frequency(self, data: bach.DataFrame) -> bach.SeriesInt64:
...
+
+        # combine
+        features_set = features_unstacked.merge(user_conversion, left_index=True, right_index=True)
+
+        y = features_set.is_converted
+        X = features_set.drop(columns=['is_converted'])
+
+        return X, y
+
+    def LogisticRegression(self, *args, **kwargs):
```
jansentom, line 203: huh?

bach/bach/series/series_numeric.py
```diff
@@ -47,6 +47,14 @@ def _arithmetic_operation(self, other, operation, fmt_str,
     def _comparator_operation(self, other, comparator, other_dtypes=('int64', 'float64')) -> 'SeriesBoolean':
         return super()._comparator_operation(other, comparator, other_dtypes)
 
+    def exp(self) -> 'SeriesAbstractNumeric':
```
jansentom, line 50: forgot to add this to docs.

bach/bach/series/series_numeric.py
jansentom, line 343: forgot to add this to docs

jansentom: imho this should be called `random`

jansentom: maybe setseed() should be added here as well.

jansentom (commented):
Review is not complete yet, but this might be a good start.

PR #668, opened 2022-04-22
Fix link to open model hub in User Intent docs

ivarpruijn:
Uses an absolute URL as a workaround for now. Might revisit this later for a proper fix.


borft (approved):
downside, is that this breaks the generated html, but I guess for now it's OK

PR #670, opened 2022-04-25
Support Window Functions for BigQuery

KathiaBarahona:
- Created `WindowFunction` enum that contains all supported window functions, this way we know which support a `window_frame_clause` based on engine. For example, all [Numbering functions](https://cloud.google.com/bigquery/docs/reference/standard-sql/numbering_functions) for BQ, don't allow this.

- Since some functions don't allow a frame clause, needed to convert `start_boundary` param as optional in `Window` class.


bach/bach/utils.py
```diff
@@ -80,3 +81,48 @@ def is_valid_column_name(dialect: Dialect, name: str) -> bool:
         prefix_ok = not any(name.startswith(prefix) for prefix in reserved_prefixes)
         return len_ok and pattern_ok and prefix_ok
     raise DatabaseNotSupportedException(dialect)
+
+
+class WindowFunction(Enum):
```
thijs-obj, line 86: There are some Windowing related enums in `partitioning.py`, any reason to not put this one there too?

bach/bach/series/series.py
```diff
@@ -1339,34 +1339,52 @@ def unique(self, partition: WrappedPartition = None, skipna: bool = True):
...
         """
         Validate that the given partition or the stored group_by is a true Window or raise an exception
         """
         from bach.partitioning import Window
-        return cast(Window, self._check_unwrap_groupby(window, isin=Window))
+
+        if not window and not self.group_by:
+            raise ValueError(f'Cannot apply {agg_function.value} function without a window clause')
+
+        if not window:
```
thijs-obj, line 1353: Is this not sort of duplicating logic of `_check_unwrap_groupby()`? I am confused on what we check there and what exactly we check here

bach/bach/series/series.py
```diff
@@ -1339,34 +1339,52 @@ def unique(self, partition: WrappedPartition = None, skipna: bool = True):
...
+        else:
+            window_cp = copy(window) if isinstance(window, Window) else window.group_by
+        if (
+            isinstance(window_cp, Window)
+            and not agg_function.supports_window_frame_clause(dialect=self.engine.dialect)
+        ):
+            # remove boundaries if the functions does not support window frame clause
+            window_cp = window_cp.set_frame_clause(start_boundary=None, end_boundary=None)
+
+        return cast(Window, self._check_unwrap_groupby(window_cp, isin=Window))
```
thijs-obj, line 1365: Should we not call `_check_unwrap_groupby()` first, and then do some of the logic in this function like, removing the bounderies from the frame clause? 

e.g. above we have a check `isinstance(window_cp, Window)`, that should always be true after calling `_check_unwrap_groupby`, or at least we won't need the cast after asserting that

thijs-obj:
Are there any functional tests that now work for BigQuery that previously didn't? Or will that come in follow-up PRs?

bach/tests/functional/bach/test_df_groupby.py
```diff
@@ -8,6 +8,7 @@
 
 from bach import Series, SeriesAbstractNumeric
 from bach.partitioning import GroupingList, GroupingSet, Rollup, Cube
+from bach.utils import WindowFunction
```
thijs-obj, line 11:
```suggestion
```

thijs-obj (commented):
Mostly looks good to me.

@jansentom perhaps you should have a look, as you originally wrote all the Window stuff

bach/bach/series/series.py
jansentom: this method being quite complex could use some tests imho.

jansentom (changes requested):
In general, LGTM, but two important things that need addressing imho

bach/bach/utils.py
KathiaBarahona: I was actually not really sure were to put this, I guess it's better to create a package just for partitioning and create an utils module there, for all enums and helpers

bach/bach/series/series.py
```diff
@@ -1353,34 +1353,52 @@ def unique(self, partition: WrappedPartition = None, skipna: bool = True):
...
+            # use series group_by attribute (e.g window.series_a.window_row_number())
+            window_cp = self.group_by
+        else:
+            window_cp = copy(window) if isinstance(window, Window) else window.group_by
+        if (
+            isinstance(window_cp, Window)
+            and not agg_function.supports_window_frame_clause(dialect=self.engine.dialect)
+        ):
+            # remove boundaries if the functions does not support window frame clause
+            window_cp = window_cp.set_frame_clause(start_boundary=None, end_boundary=None)
```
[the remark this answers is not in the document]
KathiaBarahona: same :/ but I had no other idea for doing. Do you have an alternative?

bach/bach/utils.py
```diff
@@ -80,3 +81,48 @@ def is_valid_column_name(dialect: Dialect, name: str) -> bool:
...
+    LAG = 'lag'
+
+    RANK = 'rank'
+    DENSE_RANK = 'dense_rank'
+    PERCENT_RANK = 'percent_rank'
+    CUME_DIST = 'cume_dist'
+    NTILE = 'ntile'
+    ROW_NUMBER = 'row_number'
+
+    NAVIGATION_FUNCTIONS = (
```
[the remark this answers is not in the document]
KathiaBarahona: True, will change this

bach/bach/series/series.py
jansentom: It first gets a possible window object:
from docstring: 
        1. If `wrapped` is a GroupBy, or if it contains one, use that.
        2. If it's None, check whether this Series has a group_by set and use that.
        3. If that still yields nothing, create a GroupBy([]) and use that.
after it checks whether it actually yielded one, and that it is a true Window, not a GroupBy.

Could you not call that on the top of this method, and on the resulting window check whether there are unsupported frame clauses set? That feels way more logical, but I might be missing the point of this slightly different implementation here.

bach/bach/series/series.py
KathiaBarahona: yeaah, I think I was testing other stuff and over complicated myself.

bach/bach/series/series.py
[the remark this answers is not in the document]
KathiaBarahona: wdym?

jansentom (approved):
Thanks, clean improvement!

PR #674, opened 2022-04-25
`Series.mode` support for BigQuery

KathiaBarahona:
`DescribeOperation` is dependent on `Series.mode`, but MODE aggregate function is not supported by BigQuery. Therefore, decided to use APPROX_TOP_COUNT, which yields an approximate result.

Create another issue in order to revisit this, since we should have an implementation that returns the exact value as Pandas and PG does. https://github.com/objectiv/objectiv-analytics/issues/673. Also, had some problems when all values are unique, BQ will return the last value from the list (sadly, cannot change sort behavior for this function)

### DIFF
--- a/bach/bach/series/series.py
+++ b/bach/bach/series/series.py
@@ -22,6 +22,7 @@ from bach.sql_model import BachSqlModel
 from bach.types import value_to_dtype, DtypeOrAlias, AllSupportedLiteralTypes
 from bach.utils import is_valid_column_name
 from sql_models.constants import NotSet, not_set, DBDialect
+from sql_models.util import is_bigquery
 
 if TYPE_CHECKING:
     from bach.partitioning import GroupBy, Window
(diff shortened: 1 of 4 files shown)


bach/bach/series/series.py
```diff
@@ -1311,9 +1312,16 @@ def mode(self, partition: WrappedPartition = None, skipna: bool = True):
         :param skipna: only ``skipna=True`` supported. This means NULL values are ignored.
         :returns: a new Series with the aggregation applied
         """
+        agg_expr = f'mode() within group (order by {{}})'
+        if is_bigquery(self.engine):
+            # BigQuery has no aggregate function for mode, therefore we use
```
thijs-obj, line 1317: I think we should mention something in the docstring about this.

We have a lot of things that don't work yet on BigQuery, but this works so as an end-user I would expect to get the actual mode here. If the docstring just mentions something about it being an approximation on some platforms, then at least we won't have to revisit this before unveiling the first version with BigQuery support

bach/tests/functional/bach/test_df_describe.py
```diff
@@ -7,14 +7,21 @@
...
-def test_df_categorical_describe() -> None:
-    df = get_bt_with_test_data()[['city', 'municipality']]
+def test_df_categorical_describe(engine) -> None:
+    df = get_df_with_test_data(engine)[['city', 'municipality']]
     result = df.describe()
+
+    mode_result = ['mode', 'Drylts', 'Súdwest-Fryslân']
+    if is_bigquery(engine):
+        # cannot change sorting for APPROX_TOP_COUNT
+        mode_result = ['mode', 'Snits', 'Súdwest-Fryslân']
```
thijs-obj, line 23: Are we sure this result is stable? Could it not be that next run this gives a different result?

thijs-obj (commented):
Nice to see more things working on BigQuery :ok_hand:

jansentom (commented):
LGTM, would be nice to have Thijs's docstring suggestion in the final merge. Nice to add this!

bach/tests/functional/bach/test_df_describe.py
KathiaBarahona: From my multiple tests, it always returns the last value. Sadly, I have no documentation or source to backup this (besides this [paper](https://people.cs.rutgers.edu/~farach/pubs/FrequentStream.pdf) but not like it says explicitly anything about sorting). So, my intuition is that it return the last observed value after performing the count.

PR #676, opened 2022-04-26
Bach - BQ query from other project

thijs-obj:
Two changes:
1. Corrected db-dtype of booleans for BigQuery, and updated test to test that
2. Make it possible to specific the project-id and dataset where a table lives for `from_table()` on BigQuery

### DIFF
--- a/bach/bach/from_database.py
+++ b/bach/bach/from_database.py
@@ -10,7 +10,7 @@ from bach.utils import escape_parameter_characters
 from sql_models.constants import DBDialect
 from sql_models.model import SqlModel, CustomSqlModelBuilder
 from sql_models.sql_generator import to_sql
-from sql_models.util import is_postgres, DatabaseNotSupportedException
+from sql_models.util import is_postgres, DatabaseNotSupportedException, is_bigquery
 
 
 def get_dtypes_from_model(engine: Engine, node: SqlModel) -> Dict[str, str]:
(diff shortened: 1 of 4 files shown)


jansentom (approved):
Sweet.

bach/bach/from_database.py
```diff
@@ -30,12 +30,27 @@ def get_dtypes_from_model(engine: Engine, node: SqlModel) -> Dict[str, str]:
...
     """ Query database to get dtypes of the given table. """
-    # using `INFORMATION_SCHEMA.COLUMNS` in capitals, as that way it works on both Postgres and BigQuery
+    if is_postgres(engine):
+        meta_data_table = 'INFORMATION_SCHEMA.COLUMNS'
+    elif is_bigquery(engine):
+        meta_data_table = 'INFORMATION_SCHEMA.COLUMNS'
+        if bq_dataset:
+            meta_data_table = f'{bq_dataset}.{meta_data_table}'
+        if bq_project_id:
+            meta_data_table = f'{bq_project_id}.{meta_data_table}'
```
KathiaBarahona, line 48: shouldn't we do this validation here as well? I mean this is a public method
```
if bq_project_id and not bq_dataset:
    raise ValueError('Cannot specifiy bq_project_id without setting bq_dataset.')
if bq_dataset and not is_bigquery(engine):
    raise ValueError('bq_dataset is a BigQuery-only option.')
```

thijs-obj: I did leave those checks out on purpose. My reasoning, a bit hard to write down, but is something like this: I think there is a sort of an implicit-contract here between the user of the function and the function: those values should only be set for bigquery, and never only the project-id. If only `bq_project_id` is set and `bq_dataset` is not, then the end effect is the same: the user gets an exception but the error is a bit less clear. I thought with the docstring line and parameter naming this function offers a clear abstraction for the user (us and future Bach programmers) to get the dtypes of a table.

I did add this extra validation to `DataFrame.to_table()` as that is not just a public function, but one we expect our end-users to use (it's part of the public documentation). A user specifying the wrong values has no reason to expect it to work, but as an extra service we do additional validation so he gets a clearer error and has a better user experience.

thijs-obj: Of course the 'implicit contract' I mention is very implicit. I initially didn't want to add more docstring to not tire out the user to much with boring reading :) But I agree that this is not very clear. I'll make it explicit by adding more docs and I'll add the checks.

thijs-obj:
Will merge this now. let me know what you think

PR #678, opened 2022-04-27
Support `Series.quantile` for BigQuery

KathiaBarahona:
1. `Series`.quantile now works for BQ
2. Changed/improved `quantile` query, now we use `DataFrame.stack` and generate `q` index series by using the column names:
For example:
 
   1. Generate a series per quantile (for BQ I had to rename columns into the format from below)

   quantile_df
   |   | q_0_25 | q_0_50 | q_0_75 |
   |:-:|:------:|:------:|:------:|
   |   |    x   |    y   |    z   |

   2. stack quantile df, this way we have column names (q) into one series
   
   stacked_quantile_df
   |     q   |   |
   |:------:|:-:|
   | q_0_25 | x |
   | q_0_50 | y |
   | q_0_75 | z |

   3. format `q` index (for BigQuery only). This is because we cannot name a column like `0.25`.

   final_quantile series
   |   q  |   |
   |:----:|:-:|
   | 0.25 | x |
   | 0.50 | y |
   | 0.75 | z |

### DIFF
--- a/bach/bach/quantile.py
+++ b/bach/bach/quantile.py
@@ -79,7 +79,7 @@ def calculate_quantiles(
         )
 
         # BigQuery returns quantile per row, need to apply distinct
-        df = df.materialize(distinct=True)
+        df = df.materialize(node_name='quantile', distinct=True)
 
     # q values should be numeric
     df['q'] = df['q'].astype(float)
--- a/bach/bach/quantile.py
+++ b/bach/bach/quantile.py
@@ -41,7 +41,7 @@ def calculate_quantiles(
 
         if is_bigquery(series.engine):
             # BigQuery names should start with a letter or underscore. Dots are not valid
-            q_col_name = f"q_{str(qt).replace('.', '_')}"
+            q_col_name = f"__q_{str(qt).replace('.', '_')}"
             agg_result = series.copy_override(name=q_col_name)._derived_agg_func(
                 partition=window,
                 expression=Expression.construct(f'percentile_cont({{}}, {qt})', series),
--- a/bach/bach/quantile.py
+++ b/bach/bach/quantile.py
@@ -1,9 +1,10 @@
 from typing import cast, Union, List
 
-from bach.series import SeriesAbstractNumeric, SeriesTimedelta, Series, SeriesFloat64
-from bach.expression import AggregateFunctionExpression
+from bach import SeriesString
+from bach.series import SeriesAbstractNumeric, SeriesTimedelta, Series
+from bach.expression import Expression, AggregateFunctionExpression, WindowFunctionExpression
 from bach.series.series import WrappedPartition
-from bach.operations.concat import SeriesConcatOperation
+from sql_models.util import is_bigquery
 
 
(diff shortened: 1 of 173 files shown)


bach/bach/quantile.py
```diff
@@ -18,35 +19,68 @@ def calculate_quantiles(
...
     for qt in quantiles:
         if qt < 0 or qt > 1:
             raise ValueError(f'value {qt} should be between 0 and 1.')
 
-        agg_result = cast(
-            'SeriesFloat64',
-            series._derived_agg_func(
+        if is_bigquery(series.engine):
+            # BigQuery names should start with a letter or underscore. Dots are not valid
+            q_col_name = f"q_{str(qt).replace('.', '_')}"
```
KathiaBarahona, line 44: this is a bit hacky because we have to replace back the dot, but I had no better alternative for this.

thijs-obj: Good work around this I would say :+1: 

But can we change this to:
```suggestion
            # BigQuery names should start with a letter or underscore. Dots are not valid
            q_col_name = f"__q_{str(qt).replace('.', '_')}"
```

I think it would be nice to more consistently use `__` to start internal column names. And I think that if we touch existing code or add new code, then we should go with that naming.

thijs-obj (commented):
Very nice change. Love the clear explanation in the PR description.

Left some small comments. Let me know what you think.
Generally looks good to me.

jansentom (approved):
Nice PR, Thijs' comments are spot on, for the rest LGTM!

bach/bach/series/series_boolean.py
```diff
@@ -36,7 +36,7 @@ class SeriesBoolean(Series, ABC):
     dtype_aliases = ('boolean', '?', bool)
     supported_db_dtype = {
         DBDialect.POSTGRES: 'boolean',
-        DBDialect.BIGQUERY: 'boolean',
+        DBDialect.BIGQUERY: 'BOOL',
```
thijs-obj, line 39: Why is this change in here?

bach/tests/functional/bach/test_df_describe.py
```diff
@@ -45,11 +40,10 @@ def test_df_numerical_describe(engine) -> None:
     result = df.describe(percentiles=[0.5])
     assert isinstance(result, DataFrame)
 
-    result = result.reset_index(drop=False)
-    mode_result = ['mode', 1., 3055.]
-    if is_bigquery(engine):
-        # cannot change sorting for APPROX_TOP_COUNT
-        mode_result = ['mode', 3., 93485.]
+    result = result.reset_index(drop=False).materialize()
+    # mode for PG and BQ yield different results, because we cannot change sorting for APPROX_TOP_COUNT
```
thijs-obj, line 44: Question: why do we not use `ANY` here like we do in `test_df_categorical_describe` ?

bach/bach/series/series_boolean.py
thijs-obj: aah I see, that's why it is in here

PR #679, opened 2022-04-27
Mark xdist group for `DataFrame.from_pandas(materialization='table')` tests

KathiaBarahona:
Some tests randomly fail, this is because `DataFrame.from_pandas(materialization='table')` write to db. So, we should mark such tests as a group

### DIFF
--- a/bach/tests/functional/bach/test_df_from_pandas.py
+++ b/bach/tests/functional/bach/test_df_from_pandas.py
@@ -62,6 +62,7 @@ def test_from_pandas_table(pg_engine):
     assert_equals_data(bt, expected_columns=EXPECTED_COLUMNS, expected_data=EXPECTED_DATA)
 
 
+@pytest.mark.xdist_group(name="from_pd_table")
 def test_from_pandas_table_injection(pg_engine):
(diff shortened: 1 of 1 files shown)


bach/tests/functional/bach/test_df_from_pandas.py
```diff
@@ -193,6 +196,7 @@ def test_from_pandas_index(materialization: str, pg_engine):
         expected_data=[[idx] + x[1:] for idx, x in enumerate(EXPECTED_DATA)])
 
 
+@pytest.mark.xdist_group(name="from_pd_table")
 @pytest.mark.parametrize("materialization", ['cte', 'table'])
 def test_from_pandas_types(materialization: str, pg_engine):
```
thijs-obj, line 201: Could we change `pg_engine` to something like `pg_engine_writeable` and then retract the write permissions for `pg_engine` and `engine`? Then we know for sure that we didn't forget any tests. Maybe the step with revoking the permissions is something for the future tho, when we have more time (okay never :)

KathiaBarahona: https://github.com/objectiv/objectiv-analytics/issues/692 made an issue for this

PR #682, opened 2022-04-28
Enable GH BQ testing

borft:
- enable multithreaded BQ testing


.github/workflows/bach-tests.yml
```diff
@@ -52,7 +57,7 @@ jobs:
         run: |
           cd bach
           python -m pip install --upgrade pip
-          pip install -r requirements-dev.txt -r requirements.txt
+          pip install -r requirements-dev.txt -r requirements.txt -r requirements-bigquery.txt
```
thijs-obj, line 60: very nice WIP this :+1: 

This line is a bit tricky tho. We should run all the regular non-bigquery tests without the bigquery requirements installed. Because we want to make sure that if the user only has the postgres dependencies that he can use Postgres. It is really easy to accidentally rely on something in `requirements-bigquery.txt` in a generic code-path, if we don't specifically test anymore that all postgres tests work with _only_  `pip install -r requirements-dev.txt -r requirements.txt`

So we should really have two separate tests:
1. What we have current: only install `requirements.txt` and run the postgres tests
2. A new test setup: install both `requirements.txt` and `requirements-bigquery.txt` and run the bigquery tests

.github/workflows/bach-tests.yml
```diff
@@ -34,6 +37,8 @@ jobs:
     env:
       # tell tests to use the above defined postgres service
       OBJ_DB_PG_TEST_URL: 'postgresql://objectiv:no_password_set@localhost:5432/objectiv'
+      GCP_SERVICE_ACCOUNT: ${{ secrets.GCP_SERVICE_ACCOUNT }}
+      OBJ_DB_BQ_CREDENTIALS_PATH: './sa.json'
```
jansentom, line 41:
```suggestion
      # Just the path to a temp file, the actual secret is fully contained in the env var above.
      OBJ_DB_BQ_CREDENTIALS_PATH: './sa.json'
```

jansentom (approved):
LGTM!

.github/workflows/bach-tests.yml
```diff
@@ -61,11 +68,27 @@ jobs:
         run: |
           cd bach
           pycodestyle bach sql_models
-      - name: Unit tests
+      - name: Unit tests postgres
         run: |
           cd bach
-          pytest tests/unit/ -W error::DeprecationWarning
-      - name: Functional tests
+          pytest -n 4 --dist loadgroup --postgres tests/unit
```
thijs-obj, line 74:
```suggestion
          pytest --postgres tests/unit
```
is actually slightly (0.5 seconds) faster on my machine, as the overhead of spinning up the workers is more than the time gained from the parallelization

thijs-obj: same for the unittests below

thijs-obj (approved):
Some small suggestions. Nothing super important. Fine to merge like is even I guess

thijs-obj (approved):

.github/workflows/bach-tests.yml
```diff
@@ -61,11 +68,27 @@ jobs:
...
+      - name: Functional tests postgres
         run: |
           cd bach
-          pytest tests/functional/ -W error::DeprecationWarning
+          pytest -n 4 --dist loadgroup --postgres tests/functional
```
borft, line 78:
```suggestion
          pytest --postgres tests/functional
```

thijs-obj (approved):

.github/workflows/bach-tests.yml
```diff
@@ -61,11 +68,27 @@ jobs:
         run: |
           cd bach
           pycodestyle bach sql_models
-      - name: Unit tests
+      - name: Unit tests postgres
         run: |
           cd bach
-          pytest tests/unit/ -W error::DeprecationWarning
-      - name: Functional tests
+          pytest --postgres tests/unit
```
borft, line 74:
```suggestion
          pytest -n 4 --dist loadgroup --postgres tests/unit
```

.github/workflows/bach-tests.yml
```diff
@@ -61,11 +68,27 @@ jobs:
...
+      - name: Functional tests postgres
         run: |
           cd bach
-          pytest tests/functional/ -W error::DeprecationWarning
+          pytest --postgres tests/functional
```
borft, line 78:
```suggestion
          pytest -n 4 --dist loadgroup --postgres tests/functional
```

.github/workflows/bach-tests.yml
borft: ran a test, but on GH the test went from 5mins to 6 mins (per python version). So in total 3 minutes slower, so reverted the change back to 4 processes.

thijs-obj: I think the test was flawed. What took so much longer were the functional tests, not the unit tests.

thijs-obj: But, as this is about seconds, and we have been chatting about this for many many seconds. Let's just merge this as!

borft: oh sure, was just waiting for the tests to be green after the last change!

PR #683, opened 2022-04-28
draft: Bach BQ: add tuple type

thijs-obj:
Please ignore this PR for now. It just exists as an easy way for me to track what I removed from the real PR: #700


bach/tests/conftest.py
```diff
@@ -27,9 +52,23 @@
...
+    POSTGRES = 'postgres'
+    BIGQUERY = 'bigquery'
+
+
+_ENGINE_CACHE: Dict[DB, Engine] = {}
+
+
 @pytest.fixture()
 def pg_engine() -> Engine:
-    return sqlalchemy.create_engine(DB_PG_TEST_URL)
```
thijs-obj, line 32: We use a cached connection here now. As a result I can run the tests with `-n15`, without exhausting the postgres connections.

bach/tests/functional/bach/test_series_array.py
```diff
@@ -0,0 +1,92 @@
...
+        ]
+    )
+
+
+def test_getitem(engine):
+    df = get_df_with_test_data(engine)[['skating_order']]
+    df = df.sort_index()[:1].materialize()
+    df['int_array'] = SeriesArray.from_value(base=df, value=[1, 2, 3], name='int_array', dtype=['int64'])
+    df['str_array'] = SeriesArray.from_value(base=df, value=['a', 'b', 'c'], name='str_array', dtype=['string'])
+    df = df.materialize()   # TODO: make tests pass without this materialize() call
```
thijs-obj, line 55: This is only a problem in Postgres. I will disable Arrays in Postgres for now and make these tests Big-query only.

bach/bach/series/series_tuple.py
```diff
@@ -0,0 +1,156 @@
+"""
```
thijs-obj, line 1: I think I'll remove this class from the PR. This might be useful later on, but probably best to focus on SeriesDict and SeriesArray for now.

bach/bach/series/series_dict.py
```diff
@@ -0,0 +1,176 @@
...
+from bach.types import DtypeOrAlias, get_series_type_from_dtype, StructuredDtype, Dtype, validate_dtype_value
+from sql_models.constants import DBDialect
+from sql_models.util import DatabaseNotSupportedException, is_bigquery
+
+
+if TYPE_CHECKING:
+    from bach import DataFrameOrSeries
+
+
+class SeriesDict(Series):
```
thijs-obj, line 19: Need to think about naming a bit. Now named the classes `SeriesDict` and `SeriesArray`. Perhaps `SeriesList` is more consistent instead of `Array`, or `Struct` instead of `Dict`


There is also `SeriesTuple`, but let's focus on the first two first.

bach/Makefile
```diff
@@ -12,7 +12,7 @@ tests-bigquery:
 # against Postgres
 	mypy --check-untyped-defs bach sql_models
 	pycodestyle bach sql_models
-	pytest -n 4 --dist loadgroup --big-query tests/unit tests/functional tests/
+	pytest -n 8 --dist loadgroup --big-query tests/unit tests/functional tests/
```
thijs-obj, line 15: possible now, see conftest.py. Much faster on my computer, but might not be the case for everyone?

thijs-obj:
Closing this. As it now only contains the tuple changes, and I think we won't merge that in the short term as there are higher priorities.

PR #684, opened 2022-04-29
Add version checking

borft:
To make this work, make sure your virtualenv is up to date.

Server component runs in a docker container, which can be built using:
```
make
```

Version check is done once when importing `objectiv-modelhub`.


modelhub/modelhub/__init__.py
```diff
@@ -8,3 +8,54 @@
 from modelhub.map import Map
 from modelhub.stack import *
 from modelhub.series import *
+
+import os
+
+if os.environ.get('OBJECTIV_VERSION_CHECK_DISABLE', 'false') == 'false':
+    import warnings
+    import asyncio
+    import aiohttp
```
thijs-obj, line 17: we sort of already discussed this over chat. But this doesn't work out of the box. We need to install this module. If I run 
`pytest tests_modelhub/`, all tests fail because `ModuleNotFoundError: No module named 'aiohttp'`. Now I know the PR description says to install those requirements, but we need to add that to setup.cfg and requirements.txt then too

modelhub/modelhub/__init__.py
```diff
@@ -8,3 +8,54 @@
...
+                    status = resp.status
+                    if status == 200:
+                        lines = await resp.text()
+                        for line in lines.split('\n'):
+                            items = line.split(':')
+                            if len(items) == 3:
+                                # this is a line containing package:updated:version
+                                if items[1] == 'True':
+                                    package = items[0]
+                                    version = items[2]
```
thijs-obj, line 46:
```suggestion
                            if len(items) == 3:
                                package, updated, version = items
                                if updated == 'True':
```
bit easier to read imho

modelhub/modelhub/__init__.py
```diff
@@ -8,3 +8,54 @@
...
+
+    try:
+        loop = asyncio.get_running_loop()
+        loop.create_task(check_package_version())
+    except RuntimeError as re:
+        try:
+            loop = asyncio.new_event_loop()
+            loop.run_until_complete(check_package_version())
+        except Exception as e:
+            pass
```
thijs-obj, line 61: I think this is not really async. I did a bit of reading, as I have never worked with `asyncio` before. So I am a noob, and don't quite understand why it doesn't work. BUT I did a simple experiment: I ran the version_checker locally, and added a `sleep(10)` in there. And then I ran this in an ipython terminal:
```python
import os; os.environ['OBJECTIV_VERSION_CHECK_URL'] = 'http://localhost:5000/check_version'
import time; start = time.time(); from modelhub import ModelHub; end = time.time(); print(end-start)
```
that says that importing it takes 10 seconds.

modelhub/modelhub/__init__.py
```diff
@@ -8,3 +8,54 @@
 from modelhub.map import Map
 from modelhub.stack import *
 from modelhub.series import *
+
+import os
+
+if os.environ.get('OBJECTIV_VERSION_CHECK_DISABLE', 'false') == 'false':
```
thijs-obj, line 14: Let's make sure we don't store PII like ip-address if this is opt-out

modelhub/modelhub/__init__.py
thijs-obj: I would wrap everything here in a try-except, just to be sure.

modelhub/modelhub/__init__.py
```diff
@@ -8,3 +8,54 @@
 from modelhub.map import Map
 from modelhub.stack import *
 from modelhub.series import *
+
+import os
```
thijs-obj, line 12: Would be great to have some comments here too, that explain what this section does.

modelhub/modelhub/__init__.py
borft: Yes, IP addresses are not stored

modelhub/modelhub/__init__.py
borft: you're right, when running it this way, it is run async, but blocking. And there are no other tasks in the queue. 

That said, your sleep call is most likely not async, so it would block regardless.

When running this from a script that has a running event loop (such as a Jupyter Notebook), in stead of creating a new loop, it will create a task and add that to  the pre-existing loop, and that is non blocking.

thijs-obj: > That said, your sleep call is most likely not async, so it would block regardless.

my sleep call was in the server-side component

> When running this from a script that has a running event loop (such as a Jupyter Notebook), in stead of creating a new loop, it will create a task and add that to the pre-existing loop, and that is non blocking.

Makes sense

modelhub/modelhub/__init__.py
borft: added more detailed explanation, as well as a url to the docs page

PR #686, opened 2022-05-03
abstaction of series multi level

KathiaBarahona:
# SeriesAbstractMultiLevel and SeriesNumericInterval
## Description
Databases have equivalent data structures for basic data types (integer, floats, strings, ... etc), but in some Bach functionalities we require of more complex db types (for example, `numrange` which is not supported by BigQuery). Therefore, one of the alternatives for solving these problems is by trying to simulate such types. 
Most commonly, complex db types are actually composed from other data structures, by following this pattern we can actually recreate them by defining a "multi-level" series. In other words, defining a series composed by other series, this way we resemble a similar structure from the type we want to simulate.

As an example, let's say we want to recreate Postgres [NUMRANGE](https://www.postgresql.org/docs/current/rangetypes.html) db type. 

We know that it is composed by:
* a lower-bound
* an upper-bound
* bound type (open, closed, ...)

So, we can say that this data structure is composed by other 3 datatypes.

We can translate this into a Series composed by 3 series (levels):

```python
class SeriesNumericInterval(SeriesAbstractMultiLevel)
     lower: SeriesAbstractNumeric
     upper: SeriesAbstractNumeric
     bound: SeriesString
```
By having this new data structure, Bach can have a common ground among all supported databases.

## How Multi-Level Series work?
A multi-level series treats each of its level as a normal series, its expression will always make reference to each level individually, unless it is required to build the final structure.

## Result (Final Structure)
A multi-level series should **always** construct a final structure. For example for `SeriesNumericInterval`, for each supported engine, the final data type results are the following:

| Database |            DB Type            | Pandas/Python Type |
|:--------:|:-----------------------------:|:------------------:|
| Postgres |            numrange           |   pandas.Interval  |
| BigQuery | struct <lower, upper, bounds> |   pandas.Interval  |

The only scenario were the multi-level series is actually "built" is when the user requires to fetch the data from the database, else each subquery will make reference to each level individually.

The class inheriting from `SeriesAbstractMultiLevel` should **ALWAYS** be in charge of:
* Building the column expression that constructs the final result per supported engine
* Parsing db type per engine to a common pandas/python type.


bach/bach/dataframe.py
```diff
@@ -1896,7 +1927,7 @@ def to_pandas(self, limit: Union[int, slice] = None) -> pandas.DataFrame:
 
         # Post-process any columns if needed. e.g. in BigQuery we represent UUIDs as text, so we convert
         # the strings that the query gives us into UUID objects
-        for name, series in self.data.items():
+        for name, series in self.all_series.items():
```
KathiaBarahona, line 1930: we were not parsing index values

bach/bach/dataframe.py
```diff
@@ -1985,37 +2032,47 @@ def get_current_node(
...
-            column_names = tuple(self.all_series.keys())
+                non_group_by_series = self.data
+
+        for s in non_group_by_series.values():
+            if not construct_multi_levels and isinstance(s, SeriesAbstractMultiLevel):
+                column_names += [s.name for s in s.levels.values()]
+                column_exprs += [s.get_column_expression() for s in s.levels.values()]
+            else:
+                column_names.append(s.name)
+                column_exprs.append(s.get_column_expression())
```
KathiaBarahona, line 2070: "refactor" this because felt it was a bit hard to follow

bach/bach/dataframe.py
```diff
@@ -750,48 +750,49 @@ def from_pandas(
...
-                base_node=base_node,
+        base_params = {
+            'engine': engine,
+            'base_node': base_node,
+            'group_by': group_by,
+            'sorted_ascending': None,
+            'index_sorting': [],
+        }
+        index: Dict[str, Series] = {
+            key: get_series_type_from_dtype(value).get_instance(
```
KathiaBarahona, line 777: preferred to use `get_instance`

bach/bach/partitioning.py
```diff
@@ -139,8 +139,20 @@ def __eq__(self, other):
...
+            return [g.get_column_expression() for g in self._index.values()]
+
+        from bach.series import SeriesAbstractMultiLevel
+        exprs = []
+        for g in self._index.values():
+            if not isinstance(g, SeriesAbstractMultiLevel):
+                exprs.append(g.get_column_expression())
+                continue
+
+            exprs += [level.get_column_expression() for level in g.levels.values()]
```
KathiaBarahona, line 153: this will always treat levels as individual series. We should construct multi levels only when calling a method that executes or visualizes the current node sql

bach/bach/partitioning.py
```diff
@@ -152,8 +164,22 @@ def get_group_by_column_expression(self) -> Optional[Expression]:
...
+        for g in self.index.values():
+            if not g.expression.has_multi_level_expressions:
+                exprs.append(g.expression)
+                fmt_strs.append('{}')
+                continue
+
+            # get each expression from the parent
+            multi_lvl_exprs = [expr for expr in g.expression.data if isinstance(expr, Expression)]
+            exprs.extend(multi_lvl_exprs)
+            fmt_strs.extend(['{}'] * len(multi_lvl_exprs))
```
KathiaBarahona, line 179: ALWAYS treats each level individually for groupby purposes. This avoids the problem of grouping by a struct in BQ

bach/bach/partitioning.py
```diff
@@ -422,21 +448,29 @@ def get_window_expression(self, window_func: Expression) -> Expression:
...
-            partition_fmt = 'partition by ' + ', '.join(['{}'] * len(self.index))
+            fmt_stmts = []
+            for series in self.index.values():
+                if not isinstance(series, SeriesAbstractMultiLevel):
+                    index_exprs.append(series.expression)
+                    fmt_stmts.append('{}')
+                else:
+                    index_exprs.extend([level.expression for level in series.levels.values()])
+                    fmt_stmts.extend(['{}'] * len(series.levels))
+            partition_fmt = 'partition by ' + ', '.join(fmt_stmts)
```
KathiaBarahona, line 470: same here, just treating levels individually when using the multi-level series in groupby

bach/bach/series/series.py
```diff
@@ -310,25 +313,29 @@ def expression(self) -> Expression:
...
     @classmethod
-    def get_class_instance(
-            cls,
-            base: DataFrameOrSeries,
-            name: str,
-            expression: Expression,
-            group_by: Optional['GroupBy'],
-            sorted_ascending: Optional[bool] = None,
-            index_sorting: List[bool] = None
+    def get_instance(
```
KathiaBarahona, line 316: renamed this method, for following a pattern. I found quite odd having `DataFrame.get_instance` and `Series.get_class_instance`, I rather prefer a common name for both. Please let me know if there's any problem with this. @thijs-obj @jansentom

bach/bach/expression.py
```diff
@@ -448,6 +454,20 @@ def has_aggregate_function(self) -> bool:
         return False
 
 
+class MultiLevelExpression(Expression):
+    def to_sql(self, dialect: Dialect, table_name: Optional[str] = None) -> str:
+        """
+        Compile the expression to a SQL fragment by calling to_sql() on every token or expression in data
+        :param table_name: Optional table name, if set all column-references will be compiled as
+            '"{table_name}"."{column_name}"' instead of just '"{column_name}"'.
+        :return SQL representation of the expression.
```
KathiaBarahona, line 463: need to update this

bach/bach/dataframe.py
```diff
@@ -3225,6 +3285,11 @@ def unstack(
 
         level_index = level if isinstance(level, int) else self.index_columns.index(level)
         index_to_unstack = self.index_columns[level_index]
+
+        from bach.series import SeriesAbstractMultiLevel
+        if isinstance(self.index[index_to_unstack], SeriesAbstractMultiLevel):
```
jansentom, line 3290: Wouldn't we want to unstack all series contained?

bach/bach/series/series.py
jansentom: makes sense to me

bach/bach/series/series.py
```diff
@@ -1055,6 +1072,9 @@ def apply_func(self, func: ColumnFunction, *args, **kwargs) -> List['Series']:
         .. warning::
             You should probably not use this method directly.
         """
+        if self.expression.has_multi_level_expressions:
+            raise NotImplementedError('cannot apply functions to a series with multiple levels.')
```
jansentom, line 1076: Wouldn't that mean applying it to all series in the multi level thingy?

bach/tests/functional/bach/test_df_groupby.py
```diff
@@ -688,3 +690,64 @@ def test_materialize_on_double_aggregation(engine):
...
+
+    expected = bt.to_pandas().groupby(by=['range', 'municipality']).sum()
+    result = bt.groupby(by=['range', 'municipality']).sum(numeric_only=True).sort_index()
+
+    result_pdf = result.rename(
+        columns={'inhabitants_sum': 'inhabitants', 'skating_order_sum': 'skating_order'},
+    ).to_pandas()
+    pd.testing.assert_frame_equal(expected.sort_index(), result_pdf)
+
+    if is_postgres(engine):
```
jansentom, line 726: quite nice!

bach/tests/functional/bach/test_series_multi_level.py
```diff
@@ -0,0 +1,191 @@
+import pandas as pd
```
jansentom, line 1: Awesome that these are so extensive!

bach/bach/series/series_multi_level.py
```diff
@@ -0,0 +1,494 @@
...
+            engine=self.engine,
+            base_node=self.base_node,
+            index=self.index,
+            group_by=self.group_by,
+            sorted_ascending=self.sorted_ascending,
+            index_sorting=self.index_sorting,
+        )
+
+
+def _parse_numeric_interval_value(dialect: DBDialect, value):
```
jansentom, line 373: Could use a return type

jansentom (commented):
Decent amount of work :) I like how it works, definitely a nice workaround for BQ nested type mayhem. The only thing I'm not sure I like is the many places where the individual series from the multi level one are combined using their individual expressions (partioning.py has most). Could that not be moved to SeriesAbstractMultiLevel?

bach/bach/series/series.py
KathiaBarahona: yes, but it's a bit dangerous because of  multi-typing. I think in a future we could support it but we might need more validations.

bach/bach/dataframe.py
KathiaBarahona: mmm I think I will expect just one level of the multi-series to be unstacked, so that means the entire series will be dissolved. Sounds of an odd behavior :thinking:

PR #696, opened 2022-05-09
Bach bq array struct pre work value renaming

thijs-obj:
Prepartion for later PR. This PR mainly contains some function renaming, such that the main PR will be a bit cleaner. There are no (intended) changes to functionality

### DIFF
--- a/bach/bach/series/series.py
+++ b/bach/bach/series/series.py
@@ -19,7 +19,7 @@ from bach.expression import Expression, NonAtomicExpression, ConstValueExpressio
 
 from bach.sql_model import BachSqlModel
 
-from bach.types import value_to_dtype, DtypeOrAlias, AllSupportedLiteralTypes
+from bach.types import value_to_dtype, DtypeOrAlias, AllSupportedLiteralTypes, value_to_series_type
 from bach.utils import is_valid_column_name
 from sql_models.constants import NotSet, not_set, DBDialect
 from sql_models.util import is_bigquery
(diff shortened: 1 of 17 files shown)


bach/bach/series/series.py
```diff
@@ -1671,15 +1671,15 @@ def value_counts(
...
+def value_to_series(base: DataFrameOrSeries,
                     value: Union[AllSupportedLiteralTypes, Series],
                     name: str = None) -> Series:
     """
     INTERNAL: Take a value and return a Series representing a column with that value.
 
     If value is already a Series it is returned unchanged unless it has no base_node set, in case
     it's a subquery. We create a copy and hook it to our base node in that case, so we can work with it.
-    If value is a constant then the right BuhTuh subclass is found for that type and instantiated
+    If value is a constant then the right Series subclass is found for that type and instantiated
```
jansentom, line 1682: ooh, legacy!

jansentom (approved):
Nothing to complain here. Too bad we lost a few BuhTuh's :) LGTM!

bach/bach/series/series.py
```diff
@@ -368,7 +368,7 @@ def value_to_expression(cls, dialect: Dialect, value: Optional[Any]) -> Expressi
         return cls.supported_literal_to_expression(dialect=dialect, literal=literal)
 
     @classmethod
-    def from_const(cls,
+    def from_value(cls,
                    base: DataFrameOrSeries,
                    value: Any,
```
KathiaBarahona, line 373: shouldn't it be?
```suggestion
                   value: Union[AllSupportedLiteralTypes, 'Series'],
```

KathiaBarahona (approved):
lgtm

bach/bach/series/series.py
thijs-obj: No I don't think so.
1. We don't support Series here.
2. If this was `AllSupportedLiteralTypes`, then subclasses also needed to support that. But e.g. `SeriesAbstractMultiLevel.from_value` only supports dicts. Mypy will (correctly) be angry otherwise

PR #701, opened 2022-05-10
Sphinx Builder for Docusaurus

ivarpruijn:
A Builder for Sphinx to write Docusaurus-compatible .mdx files from our source to the Modeling documentation.

To release:

1.  [x] Test.
2.  [x] Deploy both site and docs changes; recrawl.
3.  [x] Merge this PR (changes a notebook name).

Can go later:
1.  [ ] Release new modelhub package.
2.  [ ] Rebuild notebook Docker containers.


bach/docs/source/open-model-hub/version-check.rst
```diff
@@ -0,0 +1,31 @@
...
+
+The Objectiv Modelhub package comes with a built-in version checker. During import, it automatically checks to see if the loaded version is the most recent version.
+
+**Why?**
+
+We are adding new models and taxonomy operation in a rapid pace, so the goal of this version check is to notify you if when models & operations are available to use.
+
+**How?**
+
+An asynchronous http request to http://versioncheck.objectiv.io on package import.
```
borft, line 19: This should be `https://version-check.objectiv.io`

bach/docs/source/open-model-hub/version-check.rst
```diff
@@ -0,0 +1,31 @@
...
+
+Your current version of the model hub and from which python package the request is made. No cookies, IPs or any other PII is being sent.
+
+**Disable the version check**
+
+Disable the check by setting the environment variable *OBJECTIV_VERSION_CHECK_DISABLE* to *true* prior to importing objectiv-modelhub.
+
+**Source code**
+
+For more detailed information, check the source code in `__init__.py <https://github.com/objectiv/objectiv-analytics/edit/add-version-checking/modelhub/modelhub/__init__.py?>`_
```
borft, line 31:
```suggestion
For more detailed information, check the source code in `__init__.py <https://github.com/objectiv/objectiv-analytics/blob/main/modelhub/modelhub/__init__.py>`_

```

bach/docs/source/open-model-hub/version-check.rst
vincenthoogsteder: will fix

borft, line 19:
```suggestion
An asynchronous http request to https://version-check.objectiv.io on package import.

```

bach/docs/source/open-model-hub/version-check.rst
```diff
@@ -0,0 +1,31 @@
...
+
+=============
+Version check
+=============
+
+The Objectiv Modelhub package comes with a built-in version checker. During import, it automatically checks to see if the loaded version is the most recent version.
+
+**Why?**
+
+We are adding new models and taxonomy operations at a high pace, so the goal of this version check is to notify you if when models & taxonomy operations are available to use.
```
borft, line 15:
```suggestion
We are adding new models and taxonomy operations at a high pace, so the goal of this version check is to notify you when models & taxonomy operations are available to use.

```

bach/docs/source/open-model-hub/version-check.rst
```diff
@@ -0,0 +1,31 @@
...
+
+**How?**
+
+An asynchronous http request to https://version-check.objectiv.io on package import.
```
borft: Add something like:
`If an update is available, a standard Python warning will be issued`

bach/docs/source/open-model-hub/version-check.rst
```diff
@@ -0,0 +1,31 @@
...
+
+**What data is being sent?**
+
+Your current version of the model hub and from which python package the request is made. No cookies, IPs or any other PII is being sent.
```
borft: actually, the http library (aiohttp) just adds itself as the UA in the request by default. We don't explicitly include this in the request.

bach/docs/source/open-model-hub/version-check.rst
```diff
@@ -0,0 +1,31 @@
...
+
+The Objectiv Modelhub package comes with a built-in version checker. During import, it automatically checks to see if the loaded version is the most recent version.
+
+**Why?**
+
+We are adding new models and taxonomy operations at a high pace, so the goal of this version check is to notify you if when models & taxonomy operations are available to use. 
+
+**How?**
+
+An asynchronous http request to https://version-check.objectiv.io on package import. Gives a Pythong warning in case a new version is available.
```
borft, line 19:
```suggestion
An asynchronous http request to https://version-check.objectiv.io is sent on package import. A Python warning message is issued in case a new version is available. 

```

bach/docs/source/open-model-hub/version-check.rst
```diff
@@ -0,0 +1,31 @@
...
+
+=============
+Version check
+=============
+
+The Objectiv Modelhub package comes with a built-in version checker. During import, it automatically checks to see if the loaded version is the most recent version.
+
+**Why?**
+
+We are adding new models and taxonomy operations at a high pace, so the goal of this version check is to notify you if when models & taxonomy operations are available to use.
```
borft, line 15: `if when` choose either if or when

bach/docs/source/open-model-hub/version-check.rst
vincenthoogsteder: ok, let's then exclude this

bach/docs/source/open-model-hub/version-check.rst
vincenthoogsteder: haha check fixing

bach/docs/source/_ext/sphinx_docusaurus_builder/depth.py
```diff
@@ -0,0 +1,32 @@
...
+    """
+
+    depth = 0
+
+    sub_depth = {}
+
+    def get(self, name=None):
+        if name:
+            return self.sub_depth[name] if name in self.sub_depth else 0
+        return depth
```
borft, line 14: seems depth is a non existing local variable, perhaps you mean self.depth?

bach/docs/source/_ext/sphinx_docusaurus_builder/docusaurus_writer.py
```diff
@@ -0,0 +1,980 @@
+from glob import escape
```
borft, line 1: unused import

bach/docs/source/_ext/sphinx_docusaurus_builder/__init__.py
```diff
@@ -0,0 +1,40 @@
...
+from setuptools import setup
+from .docusaurus_builder import DocusaurusBuilder
+
+here = path.abspath(path.dirname(__file__))
+
+# use README.md to set a long description for the extension
+with open(path.join(here, 'README.md'), encoding='utf-8') as f:
+    long_description = f.read()
+
+# install the required libraries
```
borft, line 15: this doesn't seem to install anything, it just adds requirements to a list?

bach/docs/source/_ext/sphinx_docusaurus_builder/docusaurus_writer.py
```diff
@@ -0,0 +1,980 @@
...
+from .depth import Depth
+from .doctree2md import Translator, Writer
+from docutils import nodes
+from pydash import _
+from munch import munchify
+import os
+import yaml
+from sphinx.util.docutils import SphinxDirective, directives
+
+frontmatter = {} # holds any frontmatter forthe current doc
```
borft, line 11: do we need a global for this?

bach/docs/source/_ext/sphinx_docusaurus_builder/docusaurus_writer.py
```diff
@@ -0,0 +1,980 @@
...
+
+        """
+        Translator.__init__(self, document, builder=builder)
+        self.builder = builder
+        self.frontmatter = frontmatter
+
+
+    @property
+    def rows(self):
+        """Stores rows in a table in the document."""
```
borft, line 51: seems to return rows, but not storing them

bach/docs/source/_ext/sphinx_docusaurus_builder/__init__.py
```diff
@@ -0,0 +1,40 @@
...
+    https://www.sphinx-doc.org/en/master/extdev/index.html#extension-metadata
+
+    """
+    
+    # Register a new builder class
+    app.add_builder(DocusaurusBuilder)
+
+    return {
+        'version': '0.0.1', # identifies the extension version
+        'parallel_read_safe': True, # whether parallel reading of source files can be used
```
borft, line 38: are we sure about this? there seems to be some state that is kept in globals and/or class variables

borft:
Generally speaking, works easy, out-of-the-box (I didn't even need the readme to make it work for me). Pages look good to me. Code highlighting also seems to be working, although Docusaurus doesn't seem to understand iPython syntax. 

Nice to be able to generate all of the html of the website in one place and take advantage of all of Docusaurus niceness.

A few remarks:
- The whole process seems quite complex (from a code perspective). As this hooks into Sphinx directly, rather than using the output, I can imagine that cannot be avoided, but for me it's hard to understand what exactly is happening.
- method doc strings are inconsistent, sometime incorrect and often missing
- most methods don't specify any type hinting for either parameters or return types
- can do with some pycodestyle cleanups (missing or extra whitespace, etc)
- there seems to be some configuration, specific to Objectiv in the code in some places (like for example `/modeling`)
- there are a lot of manually created `index.rst` files, which live in the same directories as generated content
- logic for headers etc is not completely the same as Sphinx's native html writer, which means that to verify the output, one has to build the website, to observe how it looks in Docusaurus, rather than using the static html files.

bach/docs/source/_ext/sphinx_docusaurus_builder/depth.py
ivarpruijn: Looks like it, fixing

bach/docs/source/_ext/sphinx_docusaurus_builder/docusaurus_writer.py
```diff
@@ -0,0 +1,980 @@
...
+
+        # add base path and trailing slash
+        slug = '/modeling/' + slug + '/'
+        if docname == 'index' or docname[-6:] == "/index":
+            # this is an index page, so should just end in '/'
+            slug = slug[:-6]
+        return slug
+
+
+    ################################################################################
```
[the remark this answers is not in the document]
ivarpruijn: All the methods below. My own made up convention :). Want to be able to find where certain types of nodes are being covered.

bach/docs/source/_ext/sphinx_docusaurus_builder/docusaurus_writer.py
```diff
@@ -0,0 +1,980 @@
...
+                    levels = config['slug_tree_levels']
+                    # format the part after the last slash
+                    index_last_part = docname.rfind('/') + 1
+                    last_part = docname[index_last_part:]
+                    parts = last_part.split('.')
+                    last_part_formatted = '.'.join(parts[-levels:])
+                    slug = docname[0:index_last_part] + last_part_formatted
+
+        # add base path and trailing slash
+        slug = '/modeling/' + slug + '/'
```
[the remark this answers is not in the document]
ivarpruijn: Think this should be a configuration setting if we ever want to release this externally, will add a comment about it

bach/docs/source/_ext/sphinx_docusaurus_builder/docusaurus_writer.py
ivarpruijn: thanks, removing

bach/docs/source/_ext/sphinx_docusaurus_builder/__init__.py
ivarpruijn: no it's just for this, will update the comment: https://packaging.python.org/en/latest/discussions/install-requires-vs-requirements/

bach/docs/source/_ext/sphinx_docusaurus_builder/__init__.py
```diff
@@ -0,0 +1,40 @@
+"""
+Copyright 2021 Objectiv B.V.
```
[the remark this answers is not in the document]
ivarpruijn: good one, will update

bach/docs/source/_ext/sphinx_docusaurus_builder/docusaurus_writer.py
ivarpruijn: You're right, will update the doc

bach/docs/source/_ext/sphinx_docusaurus_builder/__init__.py
ivarpruijn: I'm not 100% sure. The original builder we based this on specified that it was read_safe, and I haven't seen any issues. But we could set this to False to be sure.

bach/docs/source/_ext/sphinx_docusaurus_builder/docusaurus_writer.py
```diff
@@ -0,0 +1,980 @@
...
+    def rows(self):
+        """Stores rows in a table in the document."""
+        rows = []
+        if not len(self.tables):
+            return rows
+        for node in self.tables[len(self.tables) - 1].children:
+            if isinstance(node, nodes.row):
+                rows.append(node)
+            else:
+                for node in node.children:
```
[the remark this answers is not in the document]
ivarpruijn: Good one, will update

bach/docs/source/_ext/sphinx_docusaurus_builder/docusaurus_writer.py
```diff
@@ -0,0 +1,980 @@
...
+                    last_part = docname[index_last_part:]
+                    parts = last_part.split('.')
+                    last_part_formatted = '.'.join(parts[-levels:])
+                    slug = docname[0:index_last_part] + last_part_formatted
+
+        # add base path and trailing slash
+        slug = '/modeling/' + slug + '/'
+        if docname == 'index' or docname[-6:] == "/index":
+            # this is an index page, so should just end in '/'
+            slug = slug[:-6]
```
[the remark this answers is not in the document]
ivarpruijn: Yes, this is to strip off the `index/` part at the end

ivarpruijn:
> Generally speaking, works easy, out-of-the-box (I didn't even need the readme to make it work for me). Pages look good to me. Code highlighting also seems to be working, although Docusaurus doesn't seem to understand iPython syntax.

Docusaurus supports these languages for highlighting: https://docusaurus.io/docs/markdown-features/code-blocks#supported-languages

> Nice to be able to generate all of the html of the website in one place and take advantage of all of Docusaurus niceness.

Agree! :)
 
> A few remarks:
> 
> * The whole process seems quite complex (from a code perspective). As this hooks into Sphinx directly, rather than using the output, I can imagine that cannot be avoided, but for me it's hard to understand what exactly is happening.

As far as I could find (documentation is fairly non-existent) this is the way Sphinx extensions should work. Would be great if there's a better, more simple way, but I didn't find it yet.

> * method doc strings are inconsistent, sometime incorrect and often missing
> * most methods don't specify any type hinting for either parameters or return types

I didn't focus on this right now, but we should definitely work on that if we want to release this externally.

> * can do with some pycodestyle cleanups (missing or extra whitespace, etc)

I don't have the IDE settings for this unfortunately, but maybe we can do a sweep later on.

> * there seems to be some configuration, specific to Objectiv in the code in some places (like for example `/modeling`)

I think there are two places we do this, made TODO notes about it in the code now.

> * there are a lot of manually created `index.rst` files, which live in the same directories as generated content

Deliberate choice to get more control over the exact content, but if we find we don't need it, it's easily removed.

> * logic for headers etc is not completely the same as Sphinx's native html writer, which means that to verify the output, one has to build the website, to observe how it looks in Docusaurus, rather than using the static html files.

The HTML output we had before didn't fully match with how it would look in Docusaurus, and we ran into some issues when moving from the HTML output to Docusaurus in the past. So I think it would be smart to move away from the HTML output altogether. If we need it for some reason I'm not seeing, however, I can try to make the logic the same.

PR #712, opened 2022-05-12
Test `SeriesJson` & fix DataFrame issue

thijs-obj:
Changes:
1. Change json tests: Instead of only testing `SeriesJsonB` now tests both `SeriesJsonB` and `SeriesJson`
2. Fix a bug between SeriesJson and DataFrame that showed up because of the test change


bach/bach/dataframe.py
```diff
@@ -887,13 +889,17 @@ def copy_override(
                     )
 
                 new_series[key] = series_type.get_class_instance(
```
thijs-obj, line 891: There was a bug here: **extra_params contained parameters that `Series.__init__()` didn't like, like `'series'`

The test `tests.functional.bach.test_series_json.test_json_get_single_value` broke on that. I am actually surprised that no other tests broke. Hope this change didn't break anything :crossed_fingers:

bach/bach/dataframe.py
```diff
@@ -887,13 +889,17 @@ def copy_override(
...
+                    engine=args['engine'],
```
thijs-obj, line 892: Note that I added these here explicitly again. If we put everything in `**extra_params` then mypy can't help us at all

bach/bach/dataframe.py
thijs-obj: aah actually, this is a problem in the SeriesJson, not necessarily here. Hold on to reviewing for a few seconds please, sorry about that

thijs-obj: Fixed in latest commit: https://github.com/objectiv/objectiv-analytics/pull/712/commits/b9fd36bbe7eab8de85b82825e33f59876ca678c4

But I suggest to still leave the changes here, as passing more parameters than needed to a function is not nice.

bach/tests/functional/bach/test_df_astype.py
```diff
@@ -161,12 +163,14 @@ def test_astype_to_time(engine):
         ]
     )
 
-
-def test_astype_to_json():
-    bt = get_bt_with_json_data(as_json=False)
-    bt_json_dict = bt.dict_column.astype('jsonb')
-    bt_json_list = bt.list_column.astype('jsonb')
-    bt_json_mixed = bt.mixed_column.astype('jsonb')
+@pytest.mark.parametrize('dtype', ('json', 'jsonb'))
```
jansentom, line 166: nice

jansentom (approved):

thijs-obj:
Will merge this as I have one okay. @KathiaBarahona this touches some things you worked on in DataFrame. Feel free to leave comments here and I'll process them later.

PR #713, opened 2022-05-12
set timeout to a more realistic value

borft:


modelhub/modelhub/__init__.py
```diff
@@ -43,7 +43,7 @@ async def check_package_version():
             data = '\n'.join(packages)
             try:
                 # set timeout to 0.5 seconds
```
thijs-obj, line 45: this doesn't make sense

borft: I think you are both right, fixed in https://github.com/objectiv/objectiv-analytics/pull/715

PR #718, opened 2022-05-12
Bach BQ: SeriesJson

thijs-obj:
## Changes
* `SeriesJson`:
  * All relevant functions: add code path for BigQuery
* `JsonBigQueryAccessor`: similar to `JsonPostgresAccessor` but for BigQuery
* tests: run against both Postgres and BigQuery

## To Do (will be follow-up PRs)
* Slices
* modelhub support for BigQuery Json series. Might require a bit more refactoring in the next PR

## Won't support
* The `<=` and `>=` operators are mapped to `<@` (left contains right) and `@>` (right contains left) on Postgres. We cannot (easily) support this on BigQuery

## Context
This PR is part of a series of PRs to get SeriesJson working for BigQuery:
1. Rename Series: https://github.com/objectiv/objectiv-analytics/pull/732
2. Move SeriesJson.Json class up one level: https://github.com/objectiv/objectiv-analytics/pull/733
3. **Support JSON on BigQuery:** https://github.com/objectiv/objectiv-analytics/pull/718


bach/bach/series/series_json.py
```diff
@@ -165,29 +182,68 @@ def supported_value_to_literal(
...
-        return Expression.construct(f'cast({{}} as {cls.get_db_dtype(dialect)})', expression)
-
-    def _comparator_operation(self, other, comparator, other_dtypes=('json', 'jsonb')):
-        return self._binary_operation(
+        if is_postgres(dialect):
+            if source_dtype in ('json', 'json_postgres'):
+                # SeriesJsonPostgres is a special case: SeriesJsonPostgres.expression already contains a cast
+                # to the database type 'jsonb', so we actually don't need to do any conversion
+                return expression
+            if source_dtype in ('string'):
```
KathiaBarahona, line 196: **nitpick (non-blocking):**
```suggestion
            if source_dtype == 'string':
```

bach/bach/series/series_json.py
```diff
@@ -165,29 +182,68 @@ def supported_value_to_literal(
...
+            raise ValueError(f'cannot convert {source_dtype} to json')
+        if is_bigquery(dialect):
+            if source_dtype in ('json', 'string'):
+                return expression
+            raise ValueError(f'cannot convert {source_dtype} to json')
+        raise DatabaseNotSupportedException(dialect)
+
+    def to_pandas_info(self) -> Optional['ToPandasInfo']:
+        if is_postgres(self.engine):
+            return ToPandasInfo('object', None)
```
KathiaBarahona, line 207: **question:** do we need this `ToPandasInfo('object', None)`? Why not returning just `None`?

bach/bach/series/series_json.py
```diff
@@ -165,29 +182,68 @@ def supported_value_to_literal(
...
+        if is_bigquery(self.engine):
+            # All data is stored as string, so if we actually want the objects, we need to load the string
+            # as json.
+            return ToPandasInfo('object', lambda x: json.loads(x) if x is not None else None)
```
KathiaBarahona, line 211: **question:** isn't it better to return `None` or raise a proper exception if json is not parsable? I can imagine someone callling `to_pandas` and not being sure which record is the one causing the error.

bach/bach/series/series_json.py
```diff
@@ -197,28 +253,22 @@ def max(self, partition: WrappedPartition = None, skipna: bool = True):
         """ INTERNAL: Only here to not trigger errors from describe """
         raise NotImplementedError()
 
-    def materialize(
```
KathiaBarahona, line 200: **question:** shouldn't this be in `SeriesJsonPostgres`? This is mainly if you try to materialize just the series

bach/bach/series/series_json.py
```diff
@@ -250,20 +300,76 @@ def __init__(self,
                          **kwargs)
 
 
+class JsonBigQueryAccessor:
```
KathiaBarahona, line 303: **question:** Can we have a parent class for both `JsonBigQueryAccessor` and `JsonPostgresAccessor`?

KathiaBarahona (commented):
Left a some questions for now.

bach/bach/series/series_json.py
thijs-obj: I think I too enthousiastically assumed there were no change to this in main when I merged the parent branch into this one. I'll look into what this is

bach/bach/series/series_json.py
```diff
@@ -250,20 +300,76 @@ def __init__(self,
...
 class JsonPostgresAccessor:
     """
-    class with accessor methods to json(b) type data columns.
+    class with accessor methods to SeriesJson data on Postgres.
     """
     def __init__(self, series_object: 'SeriesJson'):
         self._series_object = series_object
+        self._return_dtype: Dtype = series_object.return_dtype
 
     def __getitem__(self, key: Union[str, int, slice]):
```
KathiaBarahona, line 366: I think we should create an issue to break this function into other functions. Lot's of things are going on and it's quite hard to follow

bach/bach/series/series_json.py
```diff
@@ -250,20 +300,76 @@ def __init__(self,
...
+
+
 class JsonPostgresAccessor:
     """
-    class with accessor methods to json(b) type data columns.
+    class with accessor methods to SeriesJson data on Postgres.
     """
     def __init__(self, series_object: 'SeriesJson'):
         self._series_object = series_object
+        self._return_dtype: Dtype = series_object.return_dtype
```
KathiaBarahona, line 364: *question:* Why do we need the `return_dtype` ? For both `SeriesJson` and `SeriesJsonPostgres` is `"json"`

KathiaBarahona: nvm, but still I think is safe to just use a constant

bach/bach/series/series_json.py
thijs-obj: I thought I needed to tell pandas to use the type 'object', but it seems it does that by default for json objects. So I'll change this to None.

bach/bach/series/series_json.py
thijs-obj: `modelhub.series.series_objectiv.SeriesGlobalContexts` and `modelhub.series.series_objectiv.SeriesLocationStack` override the `return_dtype` value. In this PR I wanted to focus on the bach side, so didn't want to make any incompatible changes here

thijs-obj: `modelhub.series.series_objectiv.SeriesGlobalContexts` and `modelhub.series.series_objectiv.SeriesLocationStack` override the `return_dtype` value. In this PR I wanted to focus on the bach side, so didn't want to make any incompatible changes here

bach/bach/series/series_json.py
thijs-obj: We could. With the code we have now, what would that bring us tho?

bach/bach/series/series_json.py
thijs-obj: Changed in 864e0288

bach/bach/series/series_json.py
thijs-obj: Changed in 864e0288

bach/bach/series/series_json.py
thijs-obj: Fixed in 864e0288

bach/bach/series/series_json.py
```diff
@@ -250,20 +300,76 @@ def __init__(self,
...
+        # note: escaping is function dependent. for JSON_VALUE[1]: 'If a JSON key uses invalid JSONPath
+        # characters, then you can escape those characters using double quotes.'
+        # [1] https://cloud.google.com/bigquery/docs/reference/standard-sql/json_functions#json_value
+        assert '"' not in key
+
+        expression = Expression.construct(f'''JSON_QUERY({{}}, '$."{key}"')''', self._series_object)
+
+        if not as_str:
+            return self._series_object.copy_override(expression=expression)
+        return self._series_object.copy_override(expression=expression).copy_override_type(SeriesString)
```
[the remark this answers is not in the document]
thijs-obj: doh! Fixed in 864e0288

thijs-obj:
Made some fixes. Replied to some comments, please let me know what you think

bach/bach/series/series_json.py
KathiaBarahona: Will be easier to split `__get_item__` method, parent class should just call child methods for each case (slicing, int, str).

bach/bach/series/series_json.py
KathiaBarahona: I think eventually we need to fix this, probably when we try to support other databases some of the blocks might be reusable IMHO

thijs-obj: That makes sense. Should be natural to pick up then I think, even if no issue is present

thijs-obj:
One approval, so will merge this as then we at least have some JSON BQ support in main. @jansentom not sure if you were still going to review this, but feel free to leave comments. I'll fix them in the next branch.

jansentom:
thanks for merging. I did review, discussed the outcome IRL with you, and forgot to hit approve here, but I do approve. Thanks!

bach/bach/series/series_json.py
thijs-obj: done in https://github.com/objectiv/objectiv-analytics/pull/818

PR #719, opened 2022-05-12
Modeling docs: support targets & attributes + smaller fixes

ivarpruijn:
Several improvements found during testing.

Links to sections:
- Added support for targets (=HTML anchors).
- Added document references to example notebooks where possible, so there's no unnecessary jumping to anchors.
- Switched to direct doc reference to SeriesGlobalContexts for `get_from_context_with_type_series` method in example notebooks.

Content:
- Updates title of Modeling docs intro to 'Modeling with Objectiv'. Added a new frontmatter directive to support this.
- Fixed compact bullet list all showing up in one line in the open model hub intro.

API reference:
- Added full support for attributes like Bach's `Series.dtype`.
- No longer repeating classnames for methods & properties, e.g. no more `DateTimeOperation` in method 'DateTimeOperation.sql_format()' for Bach's `Series.AbstractDateTime.dt`.
  - Increased max depth level for the TOC, so headings like 'parameters' and 'return' are still shown. Affects other docs like in Tracking as well.
- Fixed small styling glitch in parameter list of methods.
- Smaller headings for all API References.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ Supports front-end engineers to [implement tracking instrumentation](https://www
  
 ### Open model hub
 
-A [growing collection of pre-built models](https://www.objectiv.io/docs/modeling/example_notebooks) that you run, combine or customize to quickly build in-depth analyses.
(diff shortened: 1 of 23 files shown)


ivarpruijn:
As discussed offline, we can merge this now into the `sphinx-docusaurus-builder` branch.

PR #666, opened 2022-04-21
WIP: Bach bq timedelta

thijs-obj:
NOT ready for review

PR #728, opened 2022-05-13
Support TimeDeltaSeries for BigQuery

KathiaBarahona:
Adapted most of timedelta functionalities, `dt` components are working, also did some improvements when casting from a raw string. 

Still missing making quantiles work

edit: quantiles working


bach/bach/series/series_datetime.py
```diff
@@ -361,14 +483,17 @@ class SeriesTimedelta(SeriesAbstractDateTime):
...
     supported_db_dtype = {
-        DBDialect.POSTGRES: 'interval'
+        DBDialect.POSTGRES: 'interval',
+        DBDialect.BIGQUERY: 'INTERVAL',
     }
     supported_value_types = (datetime.timedelta, numpy.timedelta64, str)
 
+    timedelta_format = {
+        DBDialect.BIGQUERY: '0-0 {days} {hours}:{minutes}:{seconds}',
+    }
```
KathiaBarahona, line 493: will remove this, ditched the idea

bach/bach/series/series_datetime.py
```diff
@@ -55,6 +74,73 @@ def sql_format(self, format_str: str) -> SeriesString:
...
+        if convert_to_seconds:
+            df = df.merge(self._get_conversion_df(), how='cross')
+
+        components_series_name = []
+        for date_part in DatePart:
+            if convert_to_seconds and date_part not in _TOTAL_SECONDS_PER_DATE_PART:
+                continue
+
+            from_interval_stmt = f'justify_interval({{}})'
+            base_fmt = f'extract({date_part.name} from {from_interval_stmt})'
```
KathiaBarahona, line 99: oh duh
```suggestion
            base_fmt = f'extract({date_part.name} from justify_interval({{}}))'
```

bach/bach/series/series_datetime.py
```diff
@@ -63,56 +149,92 @@ def components(self) -> DataFrame:
...
+            expression = Expression.construct(f'extract(epoch from {{}})', self._series)
+            return (
+                self._series
+                .copy_override_type(SeriesFloat64)
+                .copy_override(name='total_seconds', expression=expression)
+            )
+
+        # bq cannot extract epoch from interval
+        # we need to calculate it by adding all date parts normalized by seconds
+        converted_components_df = self._get_components_df(convert_to_seconds=True)
```
thijs-obj, line 223: I think it would be easier to take add the timedelta to the timestamp `1970-01-01 00:00:00` and then get the number of seconds since 1970 from the result.
Something like:
```python
expression = Expression.construct("UNIX_SECONDS(CAST('1970-01-01' AS TIMESTAMP) + {})")
```

thijs-obj: Maybe use UNIX_MICROS [1] and divide the result to get a float.

[1] https://cloud.google.com/bigquery/docs/reference/standard-sql/timestamp_functions#unix_micros

bach/bach/series/series_datetime.py
```diff
@@ -377,9 +499,28 @@ def supported_value_to_literal(
             dialect: Dialect,
             value: Union[str, numpy.timedelta64, datetime.timedelta]
     ) -> Expression:
-        value = str(value)
-        # TODO: check here already that the string has the correct format
-        return Expression.string_value(value)
+        # pandas.Timedelta checks already that the string has the correct format
+        value_td = pandas.Timedelta(value)
+
+        # interval values in iso format are allowed in SQL (both BQ and PG)
```
thijs-obj, line 505: :ok_hand: nice to use a standard format here and have pandas output that for us

bach/bach/series/series_datetime.py
```diff
@@ -63,56 +149,92 @@ def components(self) -> DataFrame:
         .. note::
             The dataframe contains only the displayed values of the timedelta.
```
thijs-obj, line 150: What does this mean?

bach/tests/functional/bach/test_series_timedelta.py
```diff
@@ -156,13 +175,14 @@ def test_timedelta_dt_properties(pg_engine) -> None:
...
     np.testing.assert_equal(
         expected_data, properties_df.sort_index().to_numpy()
     )
 
 
-def test_timedelta_dt_components(pg_engine) -> None:
+def test_timedelta_dt_components(engine) -> None:
     pdf = pd.DataFrame(
         data={
             'start_date': [
```
thijs-obj, line 188: This branch has some more interesting test cases: https://github.com/objectiv/objectiv-analytics/pull/666/files#diff-b8257690568fc0673228b1ac886f4a7d7998ef2141d30b1e791ccf239bee7f1bR166

Might not be possible to get them all working. But not sure whether for example the rounding of milliseconds is now consistent. Generally would be good to have some more cases I think

thijs-obj:
Would be nice to also convert this test: `tests.functional.bach.test_df_setitem.test_set_const_timedelta()`
Should work out of the box.

bach/bach/series/series_datetime.py
```diff
@@ -55,6 +74,73 @@ def sql_format(self, format_str: str) -> SeriesString:
 
 
 class TimedeltaOperation(DateTimeOperation):
+    def __init__(self, series: 'SeriesAbstractDateTime'):
+        super().__init__(series)
```
thijs-obj, line 78:
```suggestion
````

**Review by thijs-obj — COMMENTED** [review]
Left a few comments. Nothing huge. Tests are also not passing yet, but that's just one newline too many I believe.

File: bach/bach/series/series_datetime.py (hunk 3 above)
Comment (KathiaBarahona) [reply bach/bach/series/series_datetime.py:223]: Yeah, this is much simpler and will avoid creating another node. We MUST  avoid creating different nodes in series components from the parent.

File: bach/bach/series/series_datetime.py (hunk 5 above)
Comment (KathiaBarahona) [reply bach/bach/series/series_datetime.py]: just random words

File: bach/bach/series/series_datetime.py [hunk 8]
```diff
@@ -15,18 +16,31 @@
...
-class DatePartFormats(Enum):
-    DAYS = 'DD'
-    HOURS = 'HH24'
-    MINUTES = 'MI'
-    SECONDS = 'SS'
-    MILLISECONDS = 'MS'
-    MICROSECONDS = 'US'
+
+# conversions for date parts to seconds
+# when adjusting intervals, 30-day time periods are represented as months
```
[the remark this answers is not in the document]
KathiaBarahona: I actually left it on purpose, I spent a lot of time not understanding how `justify_interval` parses stuff. so, I prefer to have that there just in case we need it someday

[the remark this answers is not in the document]
thijs-obj: ok

bach/bach/series/series_datetime.py
```diff
@@ -463,5 +572,30 @@ def quantile(
...
-        result = calculate_quantiles(series=self.copy(), partition=partition, q=q)
-        return cast('SeriesTimedelta', result)
+
+        if not is_bigquery(self.engine):
+            return (
+                calculate_quantiles(series=self.copy(), partition=partition, q=q)
+                .copy_override_type(SeriesTimedelta)
+            )
+
+        assert isinstance(self.dt, TimedeltaOperation)  # help mypy
```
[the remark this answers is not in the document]
KathiaBarahona: huh thought mypy was going to complain for different return type, seems not. works

PR #737, opened 2022-05-18
WIP: refactor version check to be sync

borft:
- refactor version check to use requests, rather than aiohttp for mor…e predictable behaviour
- remove redundate requirements + stubs
- remove pytest config for asyncio

### DIFF
--- a/modelhub/modelhub/__init__.py
+++ b/modelhub/modelhub/__init__.py
@@ -1,7 +1,7 @@
 """
 Copyright 2021 Objectiv B.V.
 """
-__version__ = '0.0.4'
+__version__ = '0.0.5'
 
 from modelhub.modelhub import ModelHub
 from modelhub.aggregate import Aggregate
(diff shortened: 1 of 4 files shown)


modelhub/modelhub/__init__.py
```diff
@@ -26,54 +26,30 @@
...
-                pass
-    except ModuleNotFoundError:
-        # this can only happen if asyncio or aiohttp are not present
+        packages = [
+                f'objectiv-bach:{bach_version}',
+                f'objectiv-modelhub:{__version__}'
+        ]
+        data = '\n'.join(packages)
+
+        response = requests.post(CHECK_URL, data=data, timeout=5)
```
borft, line 40: @jansentom instead of `aiohttp`, this now simply uses `requests`, which was already part of the requirements

jansentom (approved):
do it

PR #744, opened 2022-05-19
Rename feature_importance.rst to feature-importance.rst

ivarpruijn:
All URLs now use a dash instead of an underscore, so let's align.


hendrik-obj (approved):
lgtm. nice improvement

PR #745, opened 2022-05-20
Explicit column names in SQL used in `from_table`

KathiaBarahona:
We should just select needed columns from SQL, selecting all columns for BigQuery might be expensive.

### DIFF
--- a/bach/bach/dataframe.py
+++ b/bach/bach/dataframe.py
@@ -588,13 +588,14 @@ class DataFrame:
 
         # use placeholders for columns in order to avoid conflicts when extracting spec references
         column_stmt = ','.join(f'{{col_{col_index}}}' for col_index in range(len(dtypes)))
-        column_place_holders = {
+        column_placeholders = {
             f'col_{col_index}': quote_identifier(engine.dialect, col_name)
             for col_index, col_name in enumerate(dtypes.keys())
         }
+        sql_params.update(column_placeholders)
 
         sql = f'SELECT {column_stmt} FROM {sql_table_name_template}'
-        model_builder = CustomSqlModelBuilder(sql=sql, name='from_table', **column_place_holders)
+        model_builder = CustomSqlModelBuilder(sql=sql, name='from_table')
         sql_model = model_builder(**sql_params)
 
         return cls._from_node(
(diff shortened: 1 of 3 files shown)


bach/bach/dataframe.py
```diff
@@ -584,7 +584,8 @@ def from_table(
             sql_table_name_template = f'{{bq_project_id}}.{sql_table_name_template}'
             sql_params['bq_project_id'] = quote_identifier(engine.dialect, bq_project_id)
 
-        sql = f'SELECT * FROM {sql_table_name_template}'
```
jansentom, line 587: The comments on line 575-576 should probably also reflect this change.

jansentom (approved):
Small comment fix requested, rest LGTM!

bach/tests/functional/bach/test_df_from_pandas.py
```diff
@@ -65,15 +65,16 @@ def test_from_pandas_table(pg_engine):
...
-        engine=pg_engine,
-        df=pdf,
-        convert_objects=True,
-        name='test_from_pd_{table}_"injection"',
-        materialization='table',
-        if_exists='replace'
-    )
-    assert_equals_data(bt, expected_columns=EXPECTED_COLUMNS_INJECTION, expected_data=EXPECTED_DATA_INJECTION)
+
+    with pytest.raises(ValueError, match="unexpected '{' in field name"):
```
thijs-obj, line 69: Question:
1) Why did we change this behaviour?
2) If this is the behaviour we want, should we not test this for both BigQuery and Postgres?

bach/bach/dataframe.py
thijs-obj: agree

thijs-obj (commented):
Added a comment

bach/tests/functional/bach/test_df_from_pandas.py
KathiaBarahona: doesn't make sense to me to allow it if user is not able to select the series with those type of names :woman_shrugging:  It is actually tested for both

bach/bach/dataframe.py
KathiaBarahona: when did those comments appeared haha Will change them

thijs-obj (commented):
lgtm

PR #750, opened 2022-05-23
Remove obsolete copright notices from READMEs

ivarpruijn:
The READMEs currently contain copyright messages. As every file has a copyright message now, as well as our License files, this is redundant. This PR removes them from the READMEs.


jansenbob (approved):
LGTM

PR #751, opened 2022-05-23
Do not aggregate the index in a DataFrame

vard-obj:
The change was done in `DataFrame._groupby_to_frame` method.
`DataFrame._groupby_to_frame` was converting indexes (those which were not included in `groupby`) into data series. In pandas when you do groupby it doesn't insert index into dataframe columns.
So with this change, `groupby` in Bach also doesn't insert indexes into the dataframe columns -> hence it is eliminating the problem of aggregation of indexes in the dataframe.


thijs-obj (approved):
looks good to me

PR #743, opened 2022-05-19
Modeling Docs: changes that don't affect URLs

thijs-obj:
* Update docstring of Series classes: always list the  underlying database type
* Assorted small fixes/improvements

This is a subset of the original changes. Original PR: https://github.com/objectiv/objectiv-analytics/pull/742 There were some discussions already there

PR #752, opened 2022-05-23
fix iff

thijs-obj:
Fix for comment https://github.com/objectiv/objectiv-analytics/pull/743#discussion_r879519618


thijs-obj:
will just merge this. FYI @jansentom

PR #754, opened 2022-05-24
`DataFrame.merge()` docstring improvement

thijs-obj:


bach/bach/dataframe.py
```diff
@@ -2139,8 +2139,10 @@ def merge(
         :param right: DataFrame or Series to join on self
         :param how: supported values: {‘left’, ‘right’, ‘outer’, ‘inner’, ‘cross’}
         :param on: optional, column(s) or condition(s) to join left and right on.
-            If *on* is based on a condition (``SeriesBoolean``), the condition MUST make reference to both
-            nodes involved in the merge.
+            If *on* is based on a condition (:class:`SeriesBoolean`), then self and right must have different
```
KathiaBarahona, line 2142: This statement : ``then self and right must have different
            base_nodes`` is not entirely true, both nodes must be referenced by the SeriesBoolean base node, but we don't care if they are the same. https://github.com/objectiv/objectiv-analytics/blob/2e6bd763deae4cadaa701b2054d9b138b26337f2/bach/bach/merge.py#L98-L102

Tho, a SeriesBoolean referencing twice to the same node is a very extreme edge case, can only imagine this happening if it was done manually

KathiaBarahona (approved):
lgtm, just a small comment

bach/bach/dataframe.py
thijs-obj: A few lines above we raise an exception if `left.base_node == right.base_node` https://github.com/objectiv/objectiv-analytics/blob/2e6bd763deae4cadaa701b2054d9b138b26337f2/bach/bach/merge.py#L95-L96

thijs-obj: I understood that correct, right? @KathiaBarahona

KathiaBarahona: Oh, then entirely ignore me haha

PR #759, opened 2022-05-24
Fix tests

thijs-obj:


KathiaBarahona (approved):

thijs-obj:
Fixed in https://github.com/objectiv/objectiv-analytics/pull/681

PR #761, opened 2022-05-24
use instance_dtype instead of dtype

KathiaBarahona:
Currently failing when dataframe has a StructureDType series


thijs-obj (approved):
nice fix :+1:

ivarpruijn:
Should we have a new release for this?

KathiaBarahona:
> Should we have a new release for this?
yup, not high priority but we should

jansentom:
I'll roll a new release for this EOD

PR #767, opened 2022-05-25
Slightly change behaviour of json slice

thijs-obj:
Changes:
1. If a slice on a json_series does not return any data, then return an empty array instead of null.
2. Fix an edge case for json-slices: `[:]`

The first one is the important change. Changed this while implementing this for BigQuery, as this seemed more logical. This PR isolates just the postgres changes, so we can review just that. Let me know if you think this is better or worse @hendrik-obj

### DIFF
--- a/bach/bach/series/series_json.py
+++ b/bach/bach/series/series_json.py
@@ -23,7 +23,8 @@ class SeriesJson(Series):
     """
     A Series that represents the JSON type and its specific operations.
 
-    Depending on the database this Series is backed by different database types:
+
+    **Database support and types**
 
     * On Postgres this utilizes the native 'jsonb' database type.
     * On BigQuery this utilizes the generic 'STRING' database type.
@@ -429,16 +430,21 @@ class JsonPostgresAccessor:
                 else:
                     where = f'>= {start}'
             else:
-                where = f'<= {stop}'
+                if key.stop is not None:
+                    where = f'<= {stop}'
+                else:
+                    # no start and stop: we want to select all elements.
+                    where = 'is not null'  # should be true for all ordinalities.
             combined_expression = f"""(select jsonb_agg(x.value)
             from jsonb_array_elements({{}}) with ordinality x
             where ordinality - 1 {where})"""
             expression_references += 1
+            non_null_expression = f"coalesce({combined_expression}, '[]'::jsonb)"
             return self._series_object\
                 .copy_override_dtype(dtype=self._return_dtype)\
                 .copy_override(
                     expression=Expression.construct(
-                        combined_expression,
+                        non_null_expression,
                         *([self._series_object] * expression_references)
                     )
                 )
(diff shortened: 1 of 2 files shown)


thijs-obj:
Okay, just noticed that this completely breaks modelhub, as that actually checks for null.

Still think the behaviour from this PR is more logical tho, as that matches what python does (e.g. `[1, 2, 3][5:5]` gives an empty array not None). But I'll work under the assumption that you'll say this is a bad idea :)

hendrik-obj (changes requested):
The change looks good to me. But indeed, modelhub assumes None, so this needs to be adjusted.

bach/bach/series/series_json.py
```diff
@@ -473,3 +492,16 @@ def get_value(self, key: str, as_str: bool = False):
         return self._series_object\
             .copy_override_dtype(dtype=return_dtype)\
             .copy_override(expression=expression)
+
+    def get_array_length(self) -> 'SeriesInt64':
```
thijs-obj, line 496: I thought about implementing this as `len()`, such that we could do `len(series_object.json)`. But it was quite hard to make that work for both arrays and objects. Therefore decided to make it an array specific function instead.

bach/bach/series/series_json.py
```diff
@@ -473,3 +492,16 @@ def get_value(self, key: str, as_str: bool = False):
...
+        """
+        Get the length of the toplevel array.
```
thijs-obj, line 498: Might do something here with a base class in a later PR. Just so that we have a place to put the docstring, without having to repeat it each time.

modelhub/modelhub/modelhub.py
```diff
@@ -48,7 +48,7 @@ def __init__(self,
         """
 
         self._time_aggregation = time_aggregation
-        self._conversion_events = {}  # type: ignore
+        self._conversion_events = cast(Dict[str, ConversionEventDefinitionType], {})
```
thijs-obj, line 51: Just something I added to help mypy help me out with the changes in map.py

thijs-obj:
@hendrik-obj I made some changes, and now all modelhub tests pass. Do you think there is anything else that's affected by this?

bach/bach/series/series_json.py
```diff
@@ -377,6 +378,19 @@ def get_value(self, key: str, as_str: bool = False) -> Union['SeriesString', 'Se
         from bach.series import SeriesString
         return self._series_object.copy_override(expression=expression).copy_override_type(SeriesString)
 
+    def get_array_length(self) -> 'SeriesInt64':
```
jansentom, line 381: Would it not be more pythonic/pandas-like if this was called `array_len()`?

jansentom (approved):
LGTM

hendrik-obj (approved):
lgtm

bach/bach/series/series_json.py
thijs-obj: Added comment in f13d24be482a4f08ceffed3964d3963ec49a7d55

thijs-obj:
processed review comments. Thus merging

PR #768, opened 2022-05-25
Bach bq json slices

thijs-obj:


bach/bach/series/series_json.py
```diff
@@ -323,11 +330,16 @@ def materialize(
...
-class JsonBigQueryAccessor:
+TSeriesJson = TypeVar('TSeriesJson', bound='SeriesJson')
+
+
+class JsonAccessor(Generic[TSeriesJson]):
     """
-    class with accessor methods to JSON type data columns on BigQuery.
+    Abstract class with accessor methods to JSON type data.
+
+    Database specific subclasses implement the abstract functions.
```
KathiaBarahona, line 340: could you list them? I think you did not marked them as abstract

bach/bach/series/series_json.py
```diff
@@ -339,84 +351,184 @@ def __getitem__(self, key: Union[str, int, slice]):
...
+
+        Assumes that self._series_object is an array!
+
+        :param value: slice.start or slice.stop
+        :param start: whether value is the slice.start (True) or slice.stop (False) value
+        :return: Expression that will evaluate to an integer that can be used to compare against positions
+                    in the array.
+        """
+        if value is not None and not isinstance(value, int):
+            raise TypeError(f'Slice value must be None or an integer, value: {value}')
```
KathiaBarahona, line 460: Not sure if we have any test coverage for this

bach/bach/series/series_json.py
```diff
@@ -339,84 +351,184 @@ def __getitem__(self, key: Union[str, int, slice]):
...
+            self._series_object, start_expression, stop_expression
+        )
+        json_str_expression = Expression.construct(
+            "'[' || ARRAY_TO_STRING(ARRAY({}), ', ') || ']'",
+            values_expression
+        )
+        return self._series_object\
+            .copy_override(expression=json_str_expression)
+
+    def _get_slice_partial_expr(self, value: Optional[int], start: bool) -> Expression:
```
KathiaBarahona, line 448: **nitpick** I think `is_start` hints more that it is a boolean
```suggestion
    def _get_slice_partial_expr(self, value: Optional[int], is_start: bool) -> Expression:
```

bach/bach/series/series_json.py
```diff
@@ -339,84 +351,184 @@ def __getitem__(self, key: Union[str, int, slice]):
...
+                    in the array.
+        """
+        if value is not None and not isinstance(value, int):
+            raise TypeError(f'Slice value must be None or an integer, value: {value}')
+
+        if value is None:
+            if start:
+                return Expression.construct('0')
+            else:
+                max_int = 2 ** 63 - 1
```
KathiaBarahona, line 466: I think it's better to have a constant for this and explain why.

bach/bach/series/series_json.py
```diff
@@ -339,84 +351,184 @@ def __getitem__(self, key: Union[str, int, slice]):
...
+        """ For documentation, see implementation in parent class :class:`JsonAccessor` """
         if key >= 0:
             expression = Expression.construct(f'''JSON_QUERY({{}}, '$[{key}]')''', self._series_object)
             return self._series_object.copy_override(expression=expression)
         # case key <= 0
         # BigQuery doesn't (yet) natively support this, so we emulate this.
         array_len = self.get_array_length()
         expr_offset = Expression.construct(f'OFFSET({{}} {key})', array_len)
-        expression = Expression.construct('JSON_EXTRACT_ARRAY({})[{}]', self._series_object, expr_offset)
+        expression = Expression.construct('JSON_QUERY_ARRAY({})[{}]', self._series_object, expr_offset)
```
KathiaBarahona, line 483: **question** Why using `JSON_QUERY_ARRAY` instead of `JSON_EXTRACT_ARRAY`?

bach/bach/series/series_json.py
```diff
@@ -339,84 +351,184 @@ def __getitem__(self, key: Union[str, int, slice]):
...
+        # index selector [1] (e.g. `$['key with special characters']`). But BigQuery only supports the dot
+        # selector [2][3]. However BigQuery does allow us to quote the key when using the dot selector [4],
+        # which is not in the jsonpath draft spec. So we quote the key, and just raise an exception if
+        # anybody tries to use a key that contains a quote.
+        # [1] https://www.ietf.org/archive/id/draft-ietf-jsonpath-base-05.html#name-index-selector
+        # [2] https://www.ietf.org/archive/id/draft-ietf-jsonpath-base-05.html#name-dot-selector
+        # [3] https://cloud.google.com/bigquery/docs/reference/standard-sql/json_functions#JSONPath_format
+        # [4] https://cloud.google.com/bigquery/docs/reference/standard-sql/json_functions
+        if '"' in key:
+            raise ValueError(f'key values containing double quotes are not supported. key: {key}')
```
KathiaBarahona, line 503: Add test for this please.

KathiaBarahona (commented):
Left some comments, I like that now we have a super class for the accessors. Now the code looks quite clean :rocket: . Tho I have a question, why do we have the following methods as public?:
- `get_array_slice`
- `get_array_item`
- `get_dict_item`

IMHO, it feels a bit counter-intuitive since as a user I can relate a json series with a dict, so for example having something like:

```python
x = [1, 2, 3]
x_series = SeriesJson.from_value(x, ...)
first_item = x_series.json.get_array_item(0) # looks weird
first_item_pythonic = x_series.json[0]  # more intuitive
```
I guess you are trying to follow something like `dict.get` pattern, but I don't think we should.
I think its best to keep those methods as private and just rely on `__get_item__`, (and I think same for `get_value`, but I guess the `as_str` param was a hacky way to deal with Postgres)

Also, I'm not sure if you are planning to open another PR for tests. As there are some scenarios that have no coverage.

bach/bach/series/series_json.py
thijs-obj: `JSON_EXTRACT_ARRAY` is a legacy function according to the docs, `JSON_QUERY_ARRAY` is recommended.
https://cloud.google.com/bigquery/docs/reference/standard-sql/json_functions

thijs-obj:
> Also, I'm not sure if you are planning to open another PR for tests. As there are some scenarios that have no coverage.
I have no PRs planned for testing more functionality, just trying to make the existing tests pass with BigQuery and added some small things. So if there are cases you think should be tested, please comment them here.

bach/bach/series/series_json.py
thijs-obj: Made them abstract in eacb15a6231a11a8955f2cbfc9cc3f35e4180e7a

bach/bach/series/series_json.py
thijs-obj: Changed in eacb15a6231a11a8955f2cbfc9cc3f35e4180e7a

bach/bach/series/series_json.py
thijs-obj: Add explanation in eacb15a6231a11a8955f2cbfc9cc3f35e4180e7a

Why do you think it is better to have a constant? (defined at the top of the file, or in the class?)

bach/bach/series/series_json.py
thijs-obj: Added in eacb15a6231a11a8955f2cbfc9cc3f35e4180e7a

thijs-obj:
I added 

> Left some comments, I like that now we have a super class for the accessors. Now the code looks quite clean rocket . Tho I have a question, why do we have the following methods as public?:
> 
> * `get_array_slice`
> * `get_array_item`
> * `get_dict_item`
> 
> IMHO, it feels a bit counter-intuitive since as a user I can relate a json series with a dict, so for example having something like:
> 
> ```python
> x = [1, 2, 3]
> x_series = SeriesJson.from_value(x, ...)
> first_item = x_series.json.get_array_item(0) # looks weird
> first_item_pythonic = x_series.json[0]  # more intuitive
> ```
> 
> I guess you are trying to follow something like `dict.get` pattern, but I don't think we should. I think its best to keep those methods as private and just rely on `__get_item__`, (and I think same for `get_value`, but I guess the `as_str` param was a hacky way to deal with Postgres)

I just thought it was nice to have simple functions that do a single thing and have them available to the user. No particular reason other than that. Made them all private in fb0ac971

bach/bach/series/series_json.py
KathiaBarahona: mainly for everyone to understand that the variable should never be changed. I think top of the file is ok, maybe we end up using it for other engines.

KathiaBarahona:
> > Also, I'm not sure if you are planning to open another PR for tests. As there are some scenarios that have no coverage.
> 
> I have no PRs planned for testing more functionality, just trying to make the existing tests pass with BigQuery and added some small things. So if there are cases you think should be tested, please comment them here.

For tests it will be nice to have
unit tests for:
* `_get_slice_partial_expr`

functional tests for:
* `get_value` with `as_str=True` 
* `get_array_length`
* and tests with NULL values. There are no nullables in any of the columns from df returned by `get_df_with_json_data`. Will be nice to see that BQ its okay with those scenarios

bach/bach/series/series_json.py
thijs-obj: I don't agree with this reasoning.

Now I fully agree that this is a constant, but defining it at the top just makes this worse to read imho. 

Bit I think anyone changing any variable should always understand its usage; we can't just change random variables and expect things to still work.

How it's now, I read it and see 'we are returning a big number so we are past the end of the array, and that big number is the max value of a 64 bit integer. that looks correct' If I define it at the top of the file than reading this fragment becomes 'we are returning some number, let me check the definition to see if it actually is a valid big number; *scroll up*; that looks correct; *scroll back*'. That just adds indirection.

If we end up re-using this constant for other cases, then I think that would be a great moment to move its definition. IMHO, we should always optimize code for understandability now, not for a hypothetical future.

bach/bach/series/series_json.py
KathiaBarahona: We already do type checks for other functionalities, so its just too keep the formality

bach/bach/series/series_json.py
KathiaBarahona: Alright, to be honest I have no concern where its defined, just gave my opinion about where I thought it was better.

KathiaBarahona (approved):
lgtm

KathiaBarahona:
One small thing before I forget, we should update the docstring for `SeriesJson`. https://github.com/objectiv/objectiv-analytics/blob/76455d4b61bc4382d1d2e4fc7da07a836651a306/bach/bach/series/series_json.py#L107-L132 , we still don't support that slicing case for BQ

thijs-obj:
> One small thing before I forget, we should update the docstring for `SeriesJson`.
> 
> https://github.com/objectiv/objectiv-analytics/blob/76455d4b61bc4382d1d2e4fc7da07a836651a306/bach/bach/series/series_json.py#L107-L132
> 
> , we still don't support that slicing case for BQ

Yes, but the tests for that are also still marked as `TODO: BigQuery`. So in a later PR we can either make this work (challenging), or we can mark the tests as skip, at which point we need to update the docs.

Perhaps unclear, but generally I have the two strategies for tracking what still need to be done:
1. Leave tests to only work with postgres. This means that the code still needs to be fixed
2. Create specific issues to track work on things.

As at the start _everything_ still needed to be done, strategy 1 was a lot easier. Slowly we get to the part where we can switch to strategy 2. But this is still strategy 1

thijs-obj:
Talked with Hendrik. He said he doesn't need to review

thijs-obj:
I think all comments have been addressed, so will merge. Let me know if you think anything still needs changing or clarifying @KathiaBarahona, will do so in next PR

Issue #692, opened 2022-05-09
Fixture for writable Postgres engine to separate database-writable tests


KathiaBarahona:
When running parallelized tests, we need to keep in mind that database-writable tests might cause conflicts with other tests that are reading/writing the same table. Therefore, we have to separate them into a different `xdist` group (this happens only for Postgres). It would be nice to change the main Postgres engine fixture to read-only, and create a separate fixture for a writable engine.

thijs-obj:
doh. I duplicated this issue in #771. Closing that one, just copy pasting the description from there here so we don't lose the ideas mentioned there.

#### text from 771:
Situation:
* Read-only tests can safely be parallelized.
* Tests that write to a database table need to be grouped, such that they are run consecutively.
  * We do this using the `@pytest.mark.xdist_group` annotation. But it is easy to forget a test

Idea:
Use a read-only account for tests that are not marked as writing tests. That way they'll consistently fail instead of failing sometimes.


Future ideas:
* Split github tasks for writing and read-only tasks. Run read-only tests for all python versions in parallel
* Make more tests read-only, by pre-creating tables in database

KathiaBarahona:
This was already fixed when introducing unique_table_test_name fixture. https://github.com/objectiv/objectiv-analytics/blob/66dc7b7981b92f01652e77d38c3d11a6ab761e29/bach/tests/conftest.py#L100

PR #770, opened 2022-05-25
group a writing test

thijs-obj:
Actually had a test fail locally because of a race-condition.

Also created a low prio issue to fix this better in the future https://github.com/objectiv/objectiv-analytics/issues/771


thijs-obj:
> LGTM, tho duplicated issue: #692 :)

good point. closed the new issue

PR #773, opened 2022-05-26
Added columns attribute for DataFrame

vard-obj:
In Pandas, we have the DataFrame.columns attribute, which returns the column labels of the DataFrame, so I added it as well for Bach DataFrame, it is an alias to the existing Bach DataFrame.data_columns.

### DIFF
--- a/bach/bach/dataframe.py
+++ b/bach/bach/dataframe.py
@@ -299,6 +299,9 @@ class DataFrame:
         """
         return list(self.data.keys())
 
+    # alias for getting data names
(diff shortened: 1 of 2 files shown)


KathiaBarahona (approved):
lgtm. Just don't forget to add it in [DataFrame/index.rst](https://github.com/objectiv/objectiv-analytics/blob/main/bach/docs/source/bach/api-reference/DataFrame/index.rst)

thijs-obj (approved):

PR #775, opened 2022-05-27
Convert UUID values in assert_equals_data

KathiaBarahona:
Just moved this functionality to `assert_equals_data` as we will need this in Modelhub tests for BQ.


bach/tests/functional/bach/test_data_and_utils.py
```diff
@@ -242,6 +257,7 @@ def assert_equals_data(
     use_to_pandas: bool = False,
     round_decimals: bool = False,
     decimal=4,
+    convert_uuid: bool = False,
```
thijs-obj, line 260: question: How do you see this with other datatypes. E.g. with json that we also store as string on BigQuery, will we then also have an `convert_json` option here? Seems a bit suboptimal to keep adding parameters to this functions

jansentom: Lets make sure we don't put behaviour in this method that should actually be tested properly to match pandas. As such,  if we add any default behaviour based on engine, it should not be here but in Bach instead?

jansentom (approved):

bach/tests/functional/bach/test_data_and_utils.py
KathiaBarahona: https://github.com/objectiv/objectiv-analytics/issues/791 Here, for the future

Issue #776, opened 2022-05-27
BQ: `window_first_row` and `window_last_row` raising error when executing query


KathiaBarahona:
Currently we allow creating window clauses without an order by clause (this is ok). But for BigQuery's `FIRST_ROW` and `LAST_ROW`, the order by clause MUST be mandatory. Therefore, we should raise an exception when calling `Series.window_first_row` and `Series.window_last_row` when trying to apply a window without sorting.

PR #777, opened 2022-05-27
BQ: Remove window frame for `LAG` and `LEAD`

KathiaBarahona:
Seems both `lag` and `lead` do not allow window frames. 

Created a new issue https://github.com/objectiv/objectiv-analytics/issues/776 , some tests were failing when I was trying to convert them into multi-db


bach/tests/functional/bach/test_df_window.py
```diff
@@ -4,11 +4,12 @@
...
 from bach.partitioning import WindowFrameMode, WindowFrameBoundary
-from tests.functional.bach.test_data_and_utils import assert_equals_data, get_bt_with_test_data
+from tests.functional.bach.test_data_and_utils import assert_equals_data, get_bt_with_test_data, get_df_with_test_data
 
 
-def test_windowing_frame_clause():
-    bt = get_bt_with_test_data(full_data_set=True)
+def test_windowing_frame_clause(pg_engine):
+    engine = pg_engine  # TODO: BigQuery
+    bt = get_df_with_test_data(engine, full_data_set=True)
```
KathiaBarahona, line 12: currently failing for BQ because we are testing with a window without order by

bach/bach/partitioning.py
```diff
@@ -53,7 +53,11 @@ class WindowFunction(Enum):
 
     def supports_window_frame_clause(self, dialect: Dialect) -> bool:
         if is_bigquery(dialect):
-            return self.value not in WindowFunction._NUMBERING_FUNCTIONS.value
+            unsupported_nav_func = (WindowFunction.LAG.value, WindowFunction.LEAD.value,)
```
thijs-obj, line 56: instead of `WindowFunction._NUMBERING_FUNCTIONS` should we not just have a `WindowFunction._FUNTIONS_THAT_DO_NOT_SUPPORT_WINDOW_FRAME_CLAUSE` or something like that?

Now we have two constants defined `_NUMBERING_FUNCTIONS` and `_NAVIGATION_FUNCTIONS`. We only use those constants in this function, but then in this function we have two exceptions defined. That doesn't make much sense

thijs-obj (commented):
tests look good. But comment about implementation

Issue #781, opened 2022-05-27
BQ: Fix SQL Format for DateTimeOperation


KathiaBarahona:
Accessor class `DateTimeOperation` uses: `to_char` which doesn't exist in BQ.
https://github.com/objectiv/objectiv-analytics/blob/1a14f9234f664bfcf4fa38e07f70cb2bc7d3676e/bach/bach/series/series_datetime.py#L50-L68

Use:
`FORMAT_TIMESTAMP` or `FORMAT_DATE` instead

Issue #780, opened 2022-05-27
ModelHub: Support BQ `ModelHub.map.is_new_user`


KathiaBarahona:
`Map.is_new_user` is generating an error when evaluating the following line https://github.com/objectiv/objectiv-analytics/blob/1a14f9234f664bfcf4fa38e07f70cb2bc7d3676e/modelhub/modelhub/map.py#L80

This probably is caused because both objects have different `base_node`, since we need to apply materialization when calling `groupby` with BQ.

PR #788, opened 2022-05-30
ModelHub Pipelines: ExtractedContextsPipeline and SessionizedDataPipeline (Support BigQuery) 

KathiaBarahona:
**Description**
(Messed up the previous branch :crying_cat_face: )
Created 3 new classes:
- `BaseDataPipeline`: Abstract class in charge of processing and validating the data generated by the pipeline. Mainly based on:
    - `_get_pipeline_result`, the function in charge of creating bach dataframe.
    - `validate_pipeline_result`, the function in charge of verifying that the bach dataframe was correctly generated, this way we always ensure the pipeline is working as expected.
 **Both functions MUST always be defined by child class**

- `ExtractedContextsPipeline`: Pipeline in charge of extracting event information from taxonomy column based on engine. Main reason of creating this is to avoid writing a raw SQL per supported database, as the entire logic behind the context extraction is performed by the pipeline and leaving sql-related tasks to ``bach``.
- `SessionizedDataPipeline`: Pipeline in charge of calculating `session_id` and `session_hit_number`. This pipeline is dependent to `ExtractedContextsPipeline`.

Also:
* moved https://github.com/objectiv/objectiv-analytics/blob/main/modelhub/tests_modelhub/functional/modelhub/data_and_utils.py to new package containing testing data in separate files.
* Seems we were dropping and creating again the same table every time we called `get_objectiv_dataframe_test`. Now table is created once only when running PG tests.
* Similar as in bach, for running tests for both engines use `make tests-all`, only for BQ: `make tests-bigquery`
* For skipping bq in test: use `@pytest.mark.skip_bigquery`, skipping pg: use `@pytest.mark.skip_postgres`
* Parallelized tests are still not implemented, but running tests for all engines should not be an issue

@hendrik-obj @vard-obj  I barely changed any existent functional test, but still please check if results from
`ExtractedContextsPipeline` and `SessionizedDataPipeline` are the expected. I tried to resemble the raw queries as accurate
as possible.

**Checklist**
- [X] Create engine based on ``db_url`` in ModelHub class
- [X] Fetching data from snowplow
- [X] extracted contexts and sessionized data working for both BQ and PG
- [X] docstrings for pipeline classes
- [X] running pytest for both engines
- [X] functional and unit tests for context data pipeline
- [X] functional tests for sessionized data pipeline

**Dependencies**
- https://github.com/objectiv/objectiv-analytics/pull/777
- https://github.com/objectiv/objectiv-analytics/pull/775
- https://github.com/objectiv/objectiv-analytics/pull/784
- https://github.com/objectiv/objectiv-analytics/pull/782

This PR just gets the based data for BQ and PG. Other modelhub functionalities still don't work with BQ, since we need to
change a couple of things in ModelHub and Bach.
Pending Issues:
- https://github.com/objectiv/objectiv-analytics/issues/780
- Most functions are failing because `ModelHub.time_agg` calls `DateTimeOperation.to_sql` (failing for BQ). https://github.com/objectiv/objectiv-analytics/issues/781


modelhub/modelhub/stack/util.py
```diff
@@ -1,53 +1,128 @@
...
+        if not provided, all expected objectiv columns will be used instead.
+    :param check_index: if true, will check if dataframe has expected index series
+    :param check_dtypes: if true, will check if each series has expected dtypes
+    :param with_md_dtypes: if true, will check if series has expected modelhub dtype
+    """
+    columns = columns_to_check if columns_to_check else ObjectivSupportedColumns.get_all_columns()
+    supported_dtypes = get_supported_dtypes_per_objectiv_column(with_md_dtypes=with_md_dtypes)
+
+    for col in columns:
+        supported_col = ObjectivSupportedColumns(col)
```
thijs-obj, line 114: The only way that we use `supported_col` in this for-loop is as `supported_col.value`, which is actually the same value as `col`. So I think it would be nicer if we just use `col` everywhere, or alternatively, us `supported_col.value` everywhere and never `col`

modelhub/modelhub/modelhub.py
```diff
@@ -69,31 +72,39 @@ def conversion_events(self):
...
+            return create_engine(db_url, credentials_path=bq_credentials_path)
+
+        import os
+        db_url = db_url or os.environ.get('DSN', 'postgresql://objectiv:@localhost:5432/objectiv')
+        return create_engine(db_url)
+
+    def get_objectiv_dataframe(
+        self,
+        db_url: str,
+        table_name: str,
```
KathiaBarahona, line 102: I think I removed the default value
```suggestion
        table_name: str = 'data',
```

modelhub/modelhub/stack/extracted_contexts.py
```diff
@@ -1,26 +1,261 @@
...
+        # rename series to objectiv supported
+        df_cp = df_cp.rename(
+            columns={
+                'cookie_id': ObjectivSupportedColumns.USER_ID.value,
+                '_type': ObjectivSupportedColumns.EVENT_TYPE.value,
+                '_types': ObjectivSupportedColumns.STACK_EVENT_TYPES.value,
+            },
+        )
+
+        return df_cp.materialize(node_name='extracted_taxonomy')
```
KathiaBarahona, line 167: Mmm this one is not necessary for PG, will consider only for BQ

modelhub/modelhub/stack/extracted_contexts.py
```diff
@@ -1,26 +1,261 @@
...
+        Returns a bach DataFrame containing each extracted field as series
+        """
+        df_cp = df.copy()
+        taxonomy_series = df_cp[self._taxonomy_column.name]
+        dtypes = _DEFAULT_TAXONOMY_DTYPE_PER_FIELD
+
+        if is_bigquery(df.engine):
+            # taxonomy column is a list, we just need to get the first value (LIST IS ALWAYS A SINGLE EVENT)
+            taxonomy_series = taxonomy_series.elements[0]
+            dtypes = self._taxonomy_column.dtype[0]
```
KathiaBarahona, line 145:
```suggestion
            # same definition as _DEFAULT_TAXONOMY_DTYPE_PER_FIELD with some extra fields
            dtypes = self._taxonomy_column.dtype[0]
```

modelhub/modelhub/stack/extracted_contexts.py
```diff
@@ -1,26 +1,261 @@
...
+        """
+        df_cp = df.copy()
+        objectiv_dtypes = get_supported_dtypes_per_objectiv_column()
+        for col in ObjectivSupportedColumns.get_extracted_context_columns():
+            if col not in df_cp.data:
+                continue
+
+            df_cp[col] = df_cp[col].astype(objectiv_dtypes[col])
+
+        return df_cp
```
KathiaBarahona, line 205: will move this to `BasePipeline`, `SessionizedDataPipeline` does the same but with sessionized columns.

modelhub/modelhub/stack/sessionized_data.py
```diff
@@ -1,50 +1,269 @@
...
+    ObjectivSupportedColumns, get_supported_dtypes_per_objectiv_column, check_objectiv_dataframe
+)
+from modelhub.stack.base_pipeline import BaseDataPipeline
+from modelhub.stack.extracted_contexts import get_extracted_contexts_df, ExtractedContextsPipeline
+
+
+class _BaseCalculatedSessionSeries(Enum):
+    IS_START_OF_SESSION = 'is_start_of_session'
+    SESSION_START_ID = 'session_start_id'
+    SESSION_COUNT = 'is_one_session'
```
KathiaBarahona, line 19: :P just using the same column name from previous raw query, the key just make more sense to me. idk

modelhub/modelhub/stack/sessionized_data.py
```diff
@@ -1,50 +1,269 @@
...
+
+        # make sure the result has AT LEAST the following series
+        self._validate_data_columns(
+            expected_columns=[
+                ObjectivSupportedColumns.EVENT_ID.value,
+                ObjectivSupportedColumns.USER_ID.value,
+                ObjectivSupportedColumns.MOMENT.value,
+            ],
+            current_columns=context_df.data_columns,
+        )
```
KathiaBarahona, line 42: Prooobably it looks unnecessary (`get_extracted_contexts_df` already validates those columns), but I think it's nice to explicitly say what the pipeline expects and if anything changes in ECP, then SDP should complain in case its requires something.

modelhub/modelhub/stack/util.py
KathiaBarahona: I agree, I guess I did it more in case `col` is not a member of `ObjectivSupportedColumns`. But I guess I can validate first if it exists in `supported_dtypes`

modelhub/tests_modelhub/data_and_utils/utils.py
```diff
@@ -1,13 +1,23 @@
...
+from sqlalchemy import create_engine
 from sqlalchemy.engine import Engine
 from tests.functional.bach.test_data_and_utils import get_bt, run_query
 
 from modelhub import ModelHub
 from tests_modelhub.data_and_utils.data_json_real import TEST_DATA_JSON_REAL, JSON_COLUMNS_REAL
+from tests_modelhub.data_and_utils.data_objectiv import TEST_DATA_OBJECTIV
+
+
+def convert_moment_to_utc_time(moment: str) -> int:
```
KathiaBarahona, line 17: I think its safe to define it as private, we just use it in `get_parsed_objectiv_data`
```suggestion
def _convert_moment_to_utc_time(moment: str) -> int:
```

modelhub/tests_modelhub/functional/modelhub/test_model_hub.py
```diff
@@ -126,7 +124,8 @@ def test_is_new_user(db_params):
     )
 
 
-def test_add_conversion_event(db_params):
```
KathiaBarahona, line 129: oops, I think I removed this test by mistake.

modelhub/modelhub/stack/util.py
```diff
@@ -1,53 +1,128 @@
...
+    _EXTRACTED_CONTEXT_COLUMNS = (
+        EVENT_ID, DAY, MOMENT, USER_ID, GLOBAL_CONTEXTS, LOCATION_STACK, EVENT_TYPE, STACK_EVENT_TYPES,
+    )
+
+    _SESSIONIZED_COLUMNS = (
+        SESSION_ID, SESSION_HIT_NUMBER,
     )
 
+    @classmethod
+    def get_extracted_context_columns(cls) -> Tuple[str, ...]:
```
KathiaBarahona, line 41: I think I'll return lists instead, I keep casting the returns for all this functions
```suggestion
    def get_extracted_context_columns(cls) -> List[str]:
```

modelhub/modelhub/stack/base_pipeline.py
```diff
@@ -0,0 +1,66 @@
...
+    @property
+    def result_series_dtypes(self) -> Dict[str, str]:
+        """
+        Mapping between series generated by pipeline with its respective dtype.
+        """
+        raise NotImplementedError()
+
+    def _convert_dtypes(self, df: bach.DataFrame) -> bach.DataFrame:
+        """
+        Helper function that converts each context series to its correct dtype,
```
KathiaBarahona, line 46:
```suggestion
        Helper function that converts each result series to its correct dtype,
```

modelhub/modelhub/stack/extracted_contexts.py
```diff
@@ -1,26 +1,273 @@
...
+
+    def _apply_extra_processing(self, df: bach.DataFrame) -> bach.DataFrame:
+        """
+        Applies more operations to the dataframe (if needed).
+
+        Returns a bach DataFrame
+        """
+        df_cp = df.copy()
+        if not is_bigquery(self._engine):
```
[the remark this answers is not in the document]
KathiaBarahona: sure

thijs-obj (commented):
Haven't quite finished reviewing. But have to go pickup the little one now, so submitting this now, and will leave some more comments later.

modelhub/modelhub/stack/sessionized_data.py
```diff
@@ -1,50 +1,272 @@
...
+        df_cp = df.copy()
+        df_cp[result_series_name] = bach.SeriesInt64.from_value(
+            base=df_cp, value=None, name=result_series_name,
+        )
+        # consider only events that are start of session
+        df_cp.loc[is_start_session_series, result_series_name] = session_start_id
+        return df_cp[result_series_name]
+
+    @staticmethod
+    def _calculate_session_count(df: bach.DataFrame) -> bach.SeriesInt64:
```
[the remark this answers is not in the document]
jansentom: That is a very nice remark. If at all possible, we should do this imho.

modelhub/modelhub/stack/extracted_contexts.py
```diff
@@ -1,26 +1,273 @@
...
+    def __init__(self, engine: Engine, table_name: str):
+        super().__init__(engine, table_name)
+        self._taxonomy_column = _get_taxonomy_column_definition(engine)
+
+        # check if table has all required columns for pipeline
+        dtypes = bach.from_database.get_dtypes_from_table(
+            engine=self._engine,
+            table_name=self._table_name,
+        )
+        self._validate_data_columns(
```
thijs-obj, line 107: Here we only validate that the right columns are present. Should we not validate the dtypes too? Would be quite simple to check I think, `assert dtypes.items() <= self._get_base_dtypes().items()` (but then with a proper error) should do the trick

modelhub/modelhub/stack/extracted_contexts.py
```diff
@@ -1,26 +1,273 @@
...
+        DBDialect.POSTGRES: {
+            'event_id': bach.SeriesUuid.dtype,
+            'day': bach.SeriesDate.dtype,
+            'moment': bach.SeriesTimestamp.dtype,
+            'cookie_id': bach.SeriesUuid.dtype,
+        },
+        DBDialect.BIGQUERY: {},
+    }
+
+    def __init__(self, engine: Engine, table_name: str):
```
thijs-obj, line 98: question: Is there any specific reason that we do validation in __init__ here, and in the other DataPipeline class we do it in `_get_pipeline_result`?

modelhub/modelhub/stack/basic_features.py
```diff
@@ -1,24 +0,0 @@
-"""
```
thijs-obj, line 1: nice cleanup :+1:

modelhub/modelhub/stack/extracted_contexts.py
```diff
@@ -1,26 +1,273 @@
...
+    if is_postgres(engine):
+        return _PG_TAXONOMY_COLUMN
+
+    if is_bigquery(engine):
+        return _BQ_TAXONOMY_COLUMN
+
+    raise Exception('define taxonomy definition for engine')
+
+
+class ExtractedContextsPipeline(BaseDataPipeline):
```
thijs-obj, line 62: I like the comments, the validation how the pipeline is cut up in small steps that is each performed by a separate function. And it looks correct to me. But I do have to wonder whether we need everything here.

In the past this was a 10 line query. Now this is a 190 line class. Of course there is stuff here that wasn't in the old version. But that's a lot of code, and it's not all generic, as we still have places where we check `is_postgres`, `is_bigquery` or `DBDialect`. I am thinking that having two queries, one for BigQuery and one for Postgres, and just call `DataFrame.from_model()` on the result would be simpler, and it would isolate the db-specific stuff better.

thijs-obj (commented):
Left a bunch of questions and comments.

One final question: On Postgres the `event_id` is the primary key, and thus unique. On BigQuery an `event_id` might appear multiple times. Is this something that gets smoothed over by the group-bys etc. we do here, or is that a concern that still needs to be tackled?

modelhub/modelhub/stack/sessionized_data.py
```diff
@@ -1,50 +1,272 @@
...
+    Gets context and sessionized data from pipeline.
+    :param engine: db connection
+    :param table_name: table from where to extract data
+    :param set_index: set index series for final dataframe
+
+    returns a bach DataFrame
+    """
+    pipeline = SessionizedDataPipeline(engine=engine, table_name=table_name)
+    result = pipeline(**kwargs)
+    if set_index:
```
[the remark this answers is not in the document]
KathiaBarahona: True, tho I was just following the same definition with `get_extracted_contexts_df`. I found it a bit weird (maybe hackish) that one function supports it and the other one not.

modelhub/modelhub/stack/extracted_contexts.py
KathiaBarahona: Yes, IMHO we should raise any errors between the expected dtypes and from what we get from the db's table as soon as possible (creating an instance).

I think that its better to have (dumb example), 
this:
``` python
    pipeline = ExtractedContextsPipeline(bq_engine, table='events') # takes 2 seconds on checking dtypes
    result = pipeline()  
    result2 = pipeline() 
```
than:
``` python
    pipeline = ExtractedContextsPipeline(bq_engine, table='events')
    result = pipeline()  # takes 2 seconds on checking dtypes
    result2 = pipeline() # takes 2 seconds on checking dtypes
```

KathiaBarahona:
> Left a bunch of questions and comments.
> 
> One final question: On Postgres the `event_id` is the primary key, and thus unique. On BigQuery an `event_id` might appear multiple times. Is this something that gets smoothed over by the group-bys etc. we do here, or is that a concern that still needs to be tackled?


I think it's a bit out of the scope of the branch, so I think we should tackle that separately.

modelhub/modelhub/stack/extracted_contexts.py
KathiaBarahona: Yes, but that it is just delaying an issue we will have with future engines. Since, we might end up with a bunch of raw queries that practically do the same, and sounds a bit counterproductive IMHO. 
Anyways, I think that the `is_postgres`, `is_bigquery`, ... `is_x`, `DBDialect` can be removed if ALL columns are in the taxonomy json and not as separate columns :woman_shrugging: , but that's something else to discuss ig.

modelhub/modelhub/stack/extracted_contexts.py
KathiaBarahona: tho hah!, will raise errors for the json types. Since `bach.from_database.get_dtypes_from_table` returns `string` for those

modelhub/modelhub/stack/sessionized_data.py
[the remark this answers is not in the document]
KathiaBarahona: @hendrik-obj @vard-obj what do you think about this?

modelhub/modelhub/stack/extracted_contexts.py
thijs-obj: 1. Why would this be an issue if we have more engines? Even if we support 5 engines, I think I would still find it less complex if we had 5 short queries vs one big class that handles all 5 engines in one.
2. Why would it be bad to delay an issue? Writing the SQL now would be super minimal investment in time compared to building this class, and at this moment time is very scarce.

Main point tho is that at this time we don't support more engines, and at this time the taxonomy json doesn't contain all columns on all engines. I think we should evaluate whether this makes sense at _this_ time.  And at this time I find this much more complex to read, review, and maintain. Of course that's a totally subjective opinion, but it's how I feel about it.

thijs-obj:
> > Left a bunch of questions and comments.
> > One final question: On Postgres the `event_id` is the primary key, and thus unique. On BigQuery an `event_id` might appear multiple times. Is this something that gets smoothed over by the group-bys etc. we do here, or is that a concern that still needs to be tackled?
> 
> I think it's a bit out of the scope of the branch, so I think we should tackle that separately.

That sounds good to me. Let's make sure we don't move the issue of the board then when we merge this PR tho.

modelhub/tests_modelhub/functional/modelhub/test_model_hub.py
```diff
@@ -57,6 +55,7 @@ def test_is_first_session(db_params):
     )
 
 
+@pytest.mark.skip_bigquery
 def test_is_new_user(db_params):
```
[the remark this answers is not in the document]
thijs-obj: OK

modelhub/modelhub/stack/extracted_contexts.py
thijs-obj: ah yes, I might have oversimplified this a bit :speak_no_evil: Still good to build a better check tho

modelhub/modelhub/stack/sessionized_data.py
```diff
@@ -1,50 +1,272 @@
...
+
+        # create lag series with correct expression
+        # this way we avoid materializing when
+        # doing an arithmetic operation with window function result
+        previous_moment_series = moment_series.copy_override(
+            expression=moment_series.window_lag(window=window).expression,
+        )
+        previous_moment_series = previous_moment_series.copy_override_type(bach.SeriesTimestamp)
+
+        event_lapsed_time = (moment_series - previous_moment_series).copy_override_type(bach.SeriesTimedelta)
```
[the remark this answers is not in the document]
thijs-obj: Can we add an inline comment to explain this? or add an assert
```python
assert isinstance(event_lapsed_time, bach.SeriesTimedelta)  # help mypy
```

Otherwise I doubt I'll find this comment thread again if I am wondering about this in the future

KathiaBarahona:
Issue for BigQuery and non-unique event-ids
https://github.com/objectiv/objectiv-analytics/issues/812

modelhub/modelhub/stack/sessionized_data.py
```diff
@@ -1,50 +1,272 @@
...
+        return df_cp[result_series_name]
+
+    @staticmethod
+    def _calculate_session_count(df: bach.DataFrame) -> bach.SeriesInt64:
+        """
+        generates a unique number for each session, but not in the right order,
+        by calculating the amount of observed session starts before current session start.
+        For example:
+           event_id  is_start_of_session   is_one_session
+              1            True                1
```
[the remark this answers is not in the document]
thijs-obj: ok makes sense to follow the original query. And the real main reason makes sense too :)

modelhub/modelhub/stack/sessionized_data.py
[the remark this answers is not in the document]
KathiaBarahona: sure

modelhub/modelhub/stack/extracted_contexts.py
KathiaBarahona: Done here https://github.com/objectiv/objectiv-analytics/pull/788/commits/6ca89551c323992df83b91d334b94149384f9a2d

modelhub/modelhub/stack/sessionized_data.py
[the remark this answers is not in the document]
KathiaBarahona: Done here https://github.com/objectiv/objectiv-analytics/pull/788/commits/7361e790f2ca54bc6c56dd2e974d9d2c9b6d0d38

modelhub/modelhub/stack/extracted_contexts.py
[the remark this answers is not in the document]
KathiaBarahona: Done here https://github.com/objectiv/objectiv-analytics/pull/788/commits/7361e790f2ca54bc6c56dd2e974d9d2c9b6d0d38

modelhub/modelhub/stack/sessionized_data.py
```diff
@@ -1,50 +1,272 @@
...
+        previous_moment_series = moment_series.copy_override(
+            expression=moment_series.window_lag(window=window).expression,
+        )
+        previous_moment_series = previous_moment_series.copy_override_type(bach.SeriesTimestamp)
+
+        event_lapsed_time = (moment_series - previous_moment_series).copy_override_type(bach.SeriesTimedelta)
+
+        result_series_name = _BaseCalculatedSessionSeries.IS_START_OF_SESSION.value
+        df_cp = df.copy()
+        df_cp[result_series_name] = True
```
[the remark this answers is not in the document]
KathiaBarahona: Done here https://github.com/objectiv/objectiv-analytics/pull/788/commits/0d069cf4ff684ef2031499b596883644e1f46524
Also explaining why we use None instead of False

modelhub/modelhub/stack/base_pipeline.py
```diff
@@ -0,0 +1,94 @@
...
+
+        for expected_key, expected_dtype in expected_dtypes.items():
+            current_dtype = current_dtypes[expected_key]
+
+            expected_series_type = get_series_type_from_dtype(expected_dtype)
+            current_series_type = get_series_type_from_dtype(current_dtype)
+            if (
+                expected_series_type != current_series_type
+                and not issubclass(current_series_type, expected_series_type)
+            ):
```
thijs-obj, line 91: unimportant nitpick that can be ignored, but I believe this is the same as. As a class is a subclass of itself
```suggestion
            if not issubclass(current_series_type, expected_series_type):
```

KathiaBarahona:
Functional Tests PR: https://github.com/objectiv/objectiv-analytics/pull/792
Unit Tests PR: https://github.com/objectiv/objectiv-analytics/pull/789

PR #793, opened 2022-05-30
Make sure to disable version check in CI

borft:


jansentom (changes requested):
Discussed offline: Think this should not be done here, but in pytest setup

PR #796, opened 2022-05-30
move version checker server into modelhub dir (as that is what it's for)

borft:


jansentom (approved):
LGTM, thanks for the quick change.

PR #797, opened 2022-05-30
Bach: Support C-Standard Date Codes in `DateTimeOperation.sql_format`

KathiaBarahona:
Since BigQuery uses standard date codes, seems it's easier for `bach` to support them by default. 
In case for PG codes, the function will try to parse the format from standard to PG, for example: `%Y-%m-%d` will be parsed into `YYYY-MM-DD`

Things done:
-  `strftime` method in `DateTimeOperation` for formatting datetime values
-  deprecation of `sql_format` in `DateTimeOperation` class. 
-  List of supported c-codes, mapping of c-codes with respective Postgres code
-  Formatting of c-code date format string to postgres date format
-  Warnings for unsupported c-codes

Btw, here just to have an idea of how the regex works: https://regex101.com/r/zo3bqo/1


bach/bach/series/utils/datetime_formats.py
```diff
@@ -0,0 +1,213 @@
...
+    'TIME_ZONE_NAME': 'TZ',
+
+    # time formats
+    'HOUR_MINUTE': 'HH24:MI',
+    'HOUR_MINUTE_SECOND': 'HH24:MI:SS',
+
+    # locale based formats
+    'LOCALE_DATE_TIME': 'Dy Mon  FMDD HH24:MI:SS YYYY',
+    'LOCALE_DATE': 'MM/DD/YY',
+    'LOCALE_TIME': 'HH24:MI:SS',
```
KathiaBarahona, line 140: ofc, this depends on the locale. Not sure if we should keep this :woman_shrugging:

thijs-obj (changes requested):
I think we are trying to do too much here. The backwards compatibility seems unnecessary to me. I suggest we add a `strftime` function, and remove the old `sql_format()` (either now or in the future). Then we don't need to have the extra parameter that indicates whether we use a database specific format string or a strftime format string.

We don't have any unittests for the translation of format strings from one format to the other, and I am sure there are edge cases that don't work[1]. Making those edge cases work doesn't have a lot of priority I believe, but having some basic unittests would be nice.


[1] Example edge case: `'%g%y'` becomes `'IYYY'`, but `to_char('IYYY')` on postgres gives a  different value than `strftime('%g%y')` on python. It seems the correct result would be `'IY""YY'`

bach/bach/series/series_datetime.py
```diff
@@ -47,13 +48,17 @@ class DateTimeOperation:
...
 
         :param format_str: The format to apply to the date/time column.
-            Currently, this uses Postgres' data format string syntax:
-            https://www.postgresql.org/docs/14/functions-formatting.html
+            Currently, this uses  1989 C standard format codes:
+            https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes
+
+        :param parse_format_str: If True, it will assume that format_str is in standard format and will
+            parse each format code into its equivalent based on engine, otherwise the format passed will be
+            used without any changes.
```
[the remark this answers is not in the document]
KathiaBarahona: same thing, didn't want to repeat myself. But will change since it might cause confusion

bach/bach/series/utils/datetime_formats.py
```diff
@@ -0,0 +1,213 @@
...
+        codes_to_replace.
+    :param date_format: string to be replaced
+
+    returns the string obtained by replacing old codes with new codes.
+    """
+    to_repl_codes_tokens = re.findall(pattern='|'.join(codes_to_replace.values()), string=date_format)
+
+    new_date_format = date_format
+    for to_repl_code in to_repl_codes_tokens:
+        code_name = [name for name, code in codes_to_replace.items() if code == to_repl_code][0]
```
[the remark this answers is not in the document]
KathiaBarahona: I actually started with an Enum (that's why all keys are in upper case). But I think I eventually felt it was just too much for something really simple. :woman_shrugging:  Do you have another idea?

bach/bach/series/series_datetime.py
```diff
@@ -47,13 +48,17 @@ class DateTimeOperation:
     def __init__(self, series: 'SeriesAbstractDateTime'):
         self._series = series
 
-    def sql_format(self, format_str: str) -> SeriesString:
+    def sql_format(self, format_str: str, parse_format_str: bool = True) -> SeriesString:
```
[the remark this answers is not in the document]
hendrik-obj: maybe indeed new function will be nicer and a deprecation warning to the old one. discussed that also with Vard and Vincent. sorry, this pr is here just to quick, so I didn't mention that yet..

bach/tests/functional/bach/test_series_date.py
```diff
@@ -77,27 +79,108 @@ def test_date_comparator(asstring: bool, engine):
...
+            supported_codes.append(code_name)
+
+    result = mt.sort_index()
+    expected = pmt[supported_codes].sort_index()
+
+    if is_postgres(engine):
+        # PG starts counting from 1
+        expected['WEEKDAY_NUMBER'] = (expected['WEEKDAY_NUMBER'].astype(int) + 1).astype(str)
+        expected['CENTURY'] = '21'  # datetime returns year / 100 and truncates integral part
+    pd.testing.assert_frame_equal(expected, result[supported_codes].to_pandas())
```
hendrik-obj, line 132: could `AM_OR_PM_LOWER` somehow also depend on locale settings? I am getting `P` for expected data in the test and therefore the test fails (postgres)

bach/tests/functional/bach/test_series_date.py
```diff
@@ -77,27 +79,108 @@ def test_date_comparator(asstring: bool, engine):
...
 
     assert mt['moment'].dtype == 'timestamp'
-    assert mt['date'].dtype == 'date'
-
     assert mt['moment'].dt.sql_format('YYYY').dtype == 'string'
 
-    mt['fyyyy'] = mt['moment'].dt.sql_format('YYYY')
-    mt['fday'] = mt['date'].dt.sql_format('Day')
+    if is_bigquery(engine):
+        non_supported_codes = [
```
hendrik-obj, line 89: can we inform users that they are using unsupported codes?

bach/bach/series/series_datetime.py
```diff
@@ -47,13 +48,17 @@ class DateTimeOperation:
...
         """
         Allow formatting of this Series (to a string type).
 
         :param format_str: The format to apply to the date/time column.
-            Currently, this uses Postgres' data format string syntax:
-            https://www.postgresql.org/docs/14/functions-formatting.html
+            Currently, this uses  1989 C standard format codes:
```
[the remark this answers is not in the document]
KathiaBarahona: Done here https://github.com/objectiv/objectiv-analytics/pull/797/commits/79789748fcc902daaaa086e962702548ebea7457

bach/bach/series/series_datetime.py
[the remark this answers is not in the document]
KathiaBarahona: Removed here https://github.com/objectiv/objectiv-analytics/pull/797/commits/79789748fcc902daaaa086e962702548ebea7457

bach/bach/series/utils/datetime_formats.py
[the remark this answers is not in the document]
KathiaBarahona: Done here: https://github.com/objectiv/objectiv-analytics/pull/797/commits/79789748fcc902daaaa086e962702548ebea7457

bach/tests/functional/bach/test_series_date.py
KathiaBarahona: Yes, done here: https://github.com/objectiv/objectiv-analytics/pull/797/commits/79789748fcc902daaaa086e962702548ebea7457

bach/tests/functional/bach/test_series_date.py
KathiaBarahona: I removed this one

KathiaBarahona: Could you  try with this one https://github.com/objectiv/objectiv-analytics/blob/ceda97581efc56c074c77cc2d5d77ada26941177/bach/tests/functional/bach/test_series_date.py#L148-L165 (added UTC timezone, might help)

KathiaBarahona:
**BigQuery doesn't support ``%f`` Code:**

BigQuery supports most of C Code formats, but `%f` is not one of them. **Main Problem** is that ModelHub relays on seconds fractional part (microseconds/milliseconds):
https://github.com/objectiv/objectiv-analytics/blob/3f2039e51c3863f1679b7fb4fdeeddd68d7a41d3/modelhub/modelhub/modelhub.py#L23

Now, for example:

```python
from datetime import datetime
dt = datetime(2022, 6, 3, 9, 53, 30, 123456)

print(dt.strftime("%Y-%m-%d %H:%M:%S.%f"))
#  >> 2022-06-03 09:53:30.123456
```

Will actually yield  `2022-06-03 09:53:30.%f` in BigQuery, now if you want the format to work in BigQuery, you should use:
`%Y-%m-%d %H:%M:%E6S` (`E6S` include seconds with 6 digits of fractional precision).

Example:
```sql
SELECT 
FORMAT_DATE('%Y-%m-%d %H:%M:%S.%f', CURRENT_DATETIME()) as incorrect_format_microseconds,
FORMAT_DATE('%Y-%m-%d %H:%M:%E6S', CURRENT_DATETIME()) as correct_format_microseconds, 
FORMAT_DATE('%Y-%m-%d %H:%M:%E3S', CURRENT_DATETIME()) as correct_format_milliseconds
```

So, I have a couple of questions:
1. How should we treat this? Just replacing `%S.%f` with `%E6S` might solve the problem, but still if user just wants microseconds? Do we just ignore this case and show `%f`? (Not sure if is worth to raise a warning, or maybe yes..?)

2. There's not code for milliseconds, but we currently use them in ModelHub default format
      If we currently want Bach, to generate that: 
              for Postgres: `time_series.dt.strftime("'%Y-%m-%d %H:%M:%S.MS)`
              for BigQuery: `time_series.dt.strftime("'%Y-%m-%d %H:%M:%E3S")`
     which is a bit nasty.
     So, should we create a new code for milliseconds? or just use microseconds?


**Conclusion**
After discussing with @jansentom 
1. replacing `%S.%f` with `%E6S`, and throwing exception/user warning if we encounter a `%f` after that.
2. Modelhub should use microseconds instead.

thijs-obj (commented):
Just committing my review, so my previous review won't block merging after this is reviewed. Didn't review again, but don't think all changes are here either?

thijs-obj (approved):
ow, so that didn't actually work. I'll approve then just to please github. But as I said before, I haven't actually reviewed it again after what we discussed yesterday.

KathiaBarahona:
> Just committing my review, so my previous review won't block merging after this is reviewed. Didn't review again, but don't think all changes are here either?

I'm just missing the substitution of `%S.%f` for BigQuery and removing `%p` code to avoid issues with locale. Thanks for unblocking!

bach/bach/series/utils/datetime_formats.py
```diff
@@ -0,0 +1,169 @@
...
+
+def parse_c_standard_code_to_postgres_code(date_format: str) -> str:
+    """
+    Parses a date format string from standard codes to Postgres date codes.
+    If c-standard code has no equivalent, it will remain in resultant string.
+
+    Steps to follow:
+        date_format = '%Y%m-%Y%m-%Y%m%d-%d'
+
+        Step 1: Get all unique groups of continuous c-codes,
```
thijs-obj, line 109: Would a while loop that runs over all characters not simpler? Keep a result string, and just keep adding characters to that, when running into a code look that up in the dictionary

KathiaBarahona: Yes, but I think it returns a kinda ugly string or maybe something I wouldn't write when doing the query manually. For example:
`%Y-%m-%d` will return: `YYYY""-MM""-DD""` instead of `YYYY-MM-DD`. Since by just keep adding characters to the result string we are not aware when we already parsed a code.

bach/tests/functional/bach/test_series_date.py
KathiaBarahona: Removed that code in https://github.com/objectiv/objectiv-analytics/pull/797/commits/f17680f8cac2f65f440cdd1437dd002eb765162b

bach/tests/functional/bach/test_series_date.py
```diff
@@ -77,30 +81,79 @@ def test_date_comparator(asstring: bool, engine):
...
+    for idx, fmt in enumerate(all_formats):
+        df[f'date_f{idx}'] = df['date_series'].dt.strftime(fmt)
+        df[f'timestamp_f{idx}'] = df['timestamp_series'].dt.strftime(fmt)
 
-    mt['fyyyy'] = mt['moment'].dt.sql_format('YYYY')
-    mt['fday'] = mt['date'].dt.sql_format('Day')
+    expected_columns = df.columns[2:]
+
+    if is_postgres(engine):
+        percentage_format_date = '2022-%%01-01'  # %% is not supported for pg
```
thijs-obj, line 105: can we not just add `"%%": "%"` to `_C_STANDARD_CODES_X_POSTGRES_DATE_CODES`?

bach/bach/series/utils/datetime_formats.py
```diff
@@ -0,0 +1,192 @@
...
+            continuous date codes yield the correct value as intended
+            For example having '%y%Y':
+              TO_CHAR(cast('2022-01-01' as date), 'YY""YYYY') will generate '222022' (correct)
+              vs.
+              TO_CHAR(cast('2022-01-01' as date), 'YYYYYY') will generate '202222' (incorrect)
+
+    """
+
+    codes_base_pattern = '|'.join(_SUPPORTED_C_STANDARD_CODES)
+    grouped_codes_matches = re.findall(pattern=rf"(?P<codes>({codes_base_pattern})+)", string=date_format)
```
thijs-obj, line 131:
```suggestion
    grouped_codes_matches = re.findall(pattern=rf"(?P<codes>(?:{codes_base_pattern})+)", string=date_format)
```
Then we only get a single group from the regex. and we don't need to discard the second one (line 137 and 140)

bach/bach/series/utils/datetime_formats.py
```diff
@@ -0,0 +1,192 @@
...
+        )
+
+    return ''.join(new_date_format_tokens)
+
+
+def parse_c_code_to_bigquery_code(date_format: str) -> str:
+    """
+    Replaces all '%S.%f' with '%E6S', since BigQuery does not support microseconds c-code.
+    """
+    if '%S.%f' in date_format:
```
thijs-obj, line 184: There are probably some edge cases here, like `%%S.%f`. but that's not a really big deal I think

bach/tests/unit/bach/test_series_utils_datetime_formats.py
```diff
@@ -0,0 +1,56 @@
...
+    # continuous c-codes in single string
+    assert parse_c_standard_code_to_postgres_code('%Y%Y%Y') == 'YYYY""YYYY""YYYY'
+
+    # nested groups
+    assert parse_c_standard_code_to_postgres_code('%Y%m-%Y%m%d-%Y%m-%m%d-%d') == (
+        'YYYY""MM"-"YYYY""MM""DD"-"YYYY""MM"-"MM""DD"-"DD'
+    )
+
+    # regular postgres format
+    assert parse_c_standard_code_to_postgres_code('YYYYMMDD') == '"YYYYMMDD"'
```
thijs-obj, line 36: :+1:

thijs-obj (approved):
Left some questions/comments.

At some point we need to move on from this tho. So I am also fine with merging this as is. it's all very contained.

bach/bach/series/utils/datetime_formats.py
KathiaBarahona: hah worked! Thanks!

Issue #815, opened 2022-06-01
Improved logging for Plugins


sdirosa:
Tracker Plugins have logic to determine if they can run in the target environment. Some of them can disable themselves if certain conditions are not met. Others skip their enrich phase if certain conditions are not met, again we don't log any information when this occurs.

It would be much better for debugging if we logged messages when these occur. E.g. in headless browsers.

PR #816, opened 2022-06-01
Improved logging for plugins

sdirosa:
Cleanup and improvements to Plugin logging and initialization.

Tracker Plugins have logic to determine if they can run in the target environment. 

Some of them can disable themselves if certain conditions are not met. Others skip their enrich phase. In either case we don't log any information.

It would be much better for debugging if we logged messages when these occur. E.g. in headless browsers.

Plugins:

- [x] ApplicationContextPlugin
- [x] OpenTaxonomyValidationPlugin
- [x] HttpContext
- [x] PathContextFromURL
- [x] RootLocationContextFromURL

As an extra I've also made sure all errors and warnings are properly colored in console.

See also: https://github.com/objectiv/objectiv-analytics/issues/815


tracker/verdaccio/config.yaml
```diff
@@ -28,7 +28,6 @@ packages:
     access: $all
     publish: $all
     unpublish: $all
-    proxy: npmjs
```
sdirosa, line 31: Unrelated tweak: we always publish/republish all of our packages at once. There's no need for this setting to be enabled. It also solves the EPUBLISHCONFLICT error one would get when publishing to Verdaccio without patching the packages first.

PR #824, opened 2022-06-02
 support requiresContext for context

borft:
- support requiresContext for context
- add/update tests

PR #825, opened 2022-06-02
Add required contexts to AbstractLocationContext

borft:
Add required contexts to AbstractLocationContext. This will enforce a  RootLocationContext and PathContext be present, when at least 1 instance of AbstractLocationContext is present. Additionally, as an InteractiveEvent requires an AbstractLocationContext, this will make sure all InteractiveEvents will have a RootLocationContext and a PathContext.

This depends on https://github.com/objectiv/objectiv-analytics/pull/824 for validation to work properly


sdirosa (approved):
Looks exactly as we discussed. I'll test this out thoroughly 👍🏼

PR #830, opened 2022-06-03
[ModelHub]: Fix import for NamedTuple

KathiaBarahona:
Imported `NamedTuple` from `mypy.build` module instead of `typing`


borft (approved):
LGTM

PR #836, opened 2022-06-07
Prepare tests

thijs-obj:


KathiaBarahona (approved):
lgtm, is there an issue for converting the rest of tests?

thijs-obj:
> lgtm, is there an issue for converting the rest of tests?

No. I added todos to all tests to track that. I'll copy paste here from another reply related to this (https://github.com/objectiv/objectiv-analytics/pull/768#issuecomment-1148507776):

generally I have the two strategies for tracking what still need to be done:
1. Leave tests to only work with postgres. This means that the code still needs to be fixed
2. Create specific issues to track work on things.

As at the start _everything_ still needed to be done, strategy 1 was a lot easier. Slowly we get to the part where we can switch to strategy 2. But this is still strategy 1

PR #846, opened 2022-06-08
[ModelHub: BQ] Support BigQuery in Aggregate.session_duration

KathiaBarahona:
Just changed the condition used in `session_duration` method, comparing an interval with a string is not supported in BigQuery. So use `dt.total_seconds` instead.

**Nano-precision Problem**

While I was changing test, I got into a weird problem with intervals containing nano-precision. For example, since modelhub can aggregate session durations by `avg` aggregation function, it is possible to generate something like this:
`0-0 0 00:00:00.123456789`. The main problem is if user tries to convert this into a Timestamp or do `dt.total_seconds`. 
Bach will generate the following expression: `UNIX_MICROS(CAST('1970-01-01' AS TIMESTAMP) + interval_column)` <- this will raise an error when executing a query, since BQ Timestamps support till microseconds precision. I'm still not sure how this can be solved, there's no way to date format an interval, and other idea is extracting each date part from interval and rebuild it (already tried justifying it but does not work)

Here for test:

```sql
with `example_session_duration_nano` as (
  select * from 
    UNNEST([STRUCT<`_index__session_id` INT64, `session_duration` INTERVAL> 
    (2, cast("""P0DT0H0M0S""" as INTERVAL)),
    (4, cast("""P0DT0H0M0.003S""" as INTERVAL)),
    (1, cast("""P0DT0H0M0.001S""" as INTERVAL)),
    (3, cast("""P0DT0H0M0.004S""" as INTERVAL)),
    (5, cast("""P0DT0H0M0S""" as INTERVAL)),
    (6, cast("""P0DT0H0M0S""" as INTERVAL)),
    (7, cast("""P0DT0H0M0S""" as INTERVAL))
  ])
)
select 
avg(`session_duration`) as `average_session_duration`,
CAST('1970-01-01' AS TIMESTAMP) + avg(`session_duration`)  # raises exception: Operation results in TIMESTAMP with nanoseconds precision
from `example_session_duration_nano` 
 
 
```

### DIFF
--- a/modelhub/modelhub/aggregate.py
+++ b/modelhub/modelhub/aggregate.py
@@ -142,10 +142,10 @@ class Aggregate:
 
         gdata = self._check_groupby(data=data, groupby=new_groupby)
         session_duration = gdata.aggregate({'moment': ['min', 'max']})
-        session_duration['session_duration'] = session_duration['moment_max']-session_duration['moment_min']
+        session_duration['session_duration'] = session_duration['moment_max'] - session_duration['moment_min']
 
         if exclude_bounces:
-            session_duration = session_duration[(session_duration['session_duration'] > '0')]
+            session_duration = session_duration[(session_duration['session_duration'].dt.total_seconds > 0)]
 
         if method not in ['sum', 'mean']:
             raise ValueError("only 'sum and 'mean' are supported for `method`")
--- a/modelhub/tests_modelhub/functional/modelhub/test_model_hub_logistic_regression.py
+++ b/modelhub/tests_modelhub/functional/modelhub/test_model_hub_logistic_regression.py
@@ -9,6 +9,7 @@ import pytest
 from tests.functional.bach.test_data_and_utils import get_df_with_test_data
 from tests_modelhub.functional.modelhub.logistic_regression_test_utils import TestLR
 from tests_modelhub.data_and_utils.utils import create_engine_from_db_params
+pytestmark = [pytest.mark.skip_bigquery]  # TODO: BigQuery
 
 
 def test_fitted_model(db_params):
--- a/modelhub/tests_modelhub/functional/modelhub/test_model_hub_session_duration.py
+++ b/modelhub/tests_modelhub/functional/modelhub/test_model_hub_session_duration.py
@@ -3,11 +3,9 @@ Copyright 2021 Objectiv B.V.
 """
 
 # Any import from modelhub initializes all the types, do not remove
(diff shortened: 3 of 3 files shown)


modelhub/tests_modelhub/functional/modelhub/test_model_hub_logistic_regression.py
```diff
@@ -9,6 +9,7 @@
 from tests.functional.bach.test_data_and_utils import get_df_with_test_data
 from tests_modelhub.functional.modelhub.logistic_regression_test_utils import TestLR
 from tests_modelhub.data_and_utils.utils import create_engine_from_db_params
+pytestmark = [pytest.mark.skip_bigquery]  # TODO: BigQuery
```
thijs-obj, line 12: Question: why do the changes in this PR make it so that we now need to skip these tests?

thijs-obj (approved):
left a question, but lgtm

modelhub/modelhub/aggregate.py
```diff
@@ -142,10 +142,10 @@ def session_duration(self,
 
         gdata = self._check_groupby(data=data, groupby=new_groupby)
         session_duration = gdata.aggregate({'moment': ['min', 'max']})
-        session_duration['session_duration'] = session_duration['moment_max']-session_duration['moment_min']
+        session_duration['session_duration'] = session_duration['moment_max'] - session_duration['moment_min']
 
         if exclude_bounces:
-            session_duration = session_duration[(session_duration['session_duration'] > '0')]
```
vard-obj, line 148: The exclude_bounces param is missing a description in the docstrings

vard-obj: the rest lgtm

vard-obj (approved):

modelhub/modelhub/aggregate.py
KathiaBarahona: what do you suggest for a description?

vard-obj: something like:
`:param exclude_bounces: if True filtering for session duration > 0 secs` ?

modelhub/tests_modelhub/functional/modelhub/test_model_hub_logistic_regression.py
thijs-obj: aah ok clear. These tests never worked for BigQuery. It's unrelated to the other changes here

PR #852, opened 2022-06-08
Bach BQ: SeriesJson slicing with dicts

KathiaBarahona:
Followed this approach: https://docs.google.com/document/d/1yPt4ML5liYgqiRbbJmhW_j3Co5PkSA0539nKocb25xg

### DIFF
--- a/bach/bach/series/series_json.py
+++ b/bach/bach/series/series_json.py
@@ -2,6 +2,8 @@
 Copyright 2021 Objectiv B.V.
 """
 import json
+import operator
+from functools import reduce
 from abc import abstractmethod
 from typing import Dict, Union, TYPE_CHECKING, Tuple, cast, Optional, List, Any, TypeVar, Generic
 
@@ -451,7 +453,11 @@ class JsonBigQueryAccessor(JsonAccessor, Generic[TSeriesJson]):
         return self._series_object\
             .copy_override(expression=json_str_expression)
 
-    def _get_slice_partial_expr(self, value: Optional[int], is_start: bool) -> Expression:
+    def _get_slice_partial_expr(
+        self,
+        value: Optional[Union[Dict[str, str], int]],
+        is_start: bool,
+    ) -> Expression:
         """
         Return expression for either the lower bound or upper bound of a slice.
 
@@ -462,6 +468,9 @@ class JsonBigQueryAccessor(JsonAccessor, Generic[TSeriesJson]):
         :return: Expression that will evaluate to an integer that can be used to compare against positions
                     in the array.
         """
+        if isinstance(value, dict):
+            return self._find_in_json_list(value, is_start)
+
         if value is not None and not isinstance(value, int):
             raise TypeError(f'Slice value must be None or an integer, value: {value}')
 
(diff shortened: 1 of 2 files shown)


bach/tests/functional/bach/test_series_json.py
```diff
@@ -261,34 +261,39 @@ def test_json_getitem_query(pg_engine, dtype):
             [4, []]
         ]
     )
+
+@pytest.mark.skip_postgres
+def test_json_same_start_end_slicing(engine, dtype):
```
KathiaBarahona, line 266: https://github.com/objectiv/objectiv-analytics/issues/853

bach/bach/series/series_json.py
```diff
@@ -479,6 +488,36 @@ def _get_slice_partial_expr(self, value: Optional[int], is_start: bool) -> Expre
...
+
+        :param filtering_slice: dictionary containing the keys and values to use on filter.
+        :param is_start: whether value is the slice.start (True) or slice.stop (False) value. If true,
+            the first element containing the values is returned, otherwise the last.
+        :return: Expression that gets the position of the element based on the slicing filters.
+        """
+        if not isinstance(filtering_slice, dict):
+            raise TypeError(f'key should be a dict, actual type: {type(filtering_slice)}')
+
+        # this way we can reuse code from the accesor without duplicating
```
thijs-obj, line 505: clever

jansentom, line 505: I was a bit confused about what `element` mean, until I discovered it in the query template below. Maybe this helps for others suffering from local blindness :)

```suggestion
        # this way we can reuse code from the accesor without duplicating.
        # 'element' will filled by iterating over the json array in the query below.
```

bach/tests/functional/bach/test_series_json.py
jansentom: Hmm, does @hendrik-obj rely on this behaviour?

jansentom: The pydoc in the accessor also states the "broken" behaviour if I understand you correctly. Should we follow `.loc` behaviour here or invent our own? WDYT?

jansentom (commented):
Very simple solution actually and useful to have! Let's check what to do with in/exclusive end of slice behaviour. @hendrik-obj @vard-obj Please shine your light as well, thanks!

thijs-obj (approved):
lgtm

I do think that we should not 'abuse' the slice operation for this. Don't think mypy likes that either. But that's a separate issue that we need to discuss with everyone involved.

bach/bach/series/series_json.py
KathiaBarahona: Done in https://github.com/objectiv/objectiv-analytics/pull/852/commits/10397788d40f0ef62c196d0f1bf6e8e049624847

bach/tests/functional/bach/test_series_json.py
KathiaBarahona: Mmmm if you added the correct pydoc, it seems that's a seperate issue. Regards, the inclusive issue.. seems that modelhub uses only dicts for start slice, which I only found in tests and here https://github.com/objectiv/objectiv-analytics/blob/4a6cd2479387512cf52fc89630b56bb54da9e228/modelhub/modelhub/series/series_objectiv.py#L171

hendrik-obj: sorry for late to this party. 
Yes, the inspiration for doing it as it was (including the value) was that it is the same as `.loc`. 
Another reason (or maybe the reason for the loc implementation) is that it is in many cases, easier. For example, it is easy to get from the slice that includes the matching end-dict, to the matching dict minus the last object. The other way around however, is difficult.

hendrik-obj: So, it is not 'broken' as https://github.com/objectiv/objectiv-analytics/issues/853 suggests.

bach/bach/series/series_json.py
```diff
@@ -465,6 +474,37 @@ def _get_slice_partial_expr(self, value: Optional[int], is_start: bool) -> Expre
...
+            expression=Expression.construct('element'),
+        )
+
+        all_filters = [
+            element_json.json.get_value(filter_key, as_str=True) == str(filter_value)
+            for filter_key, filter_value in filtering_slice.items()
+        ]
+        slicing_mask = reduce(operator.and_, all_filters)
+        fmt = (
+            f"(select {'min' if is_start else 'max'}(case when {{}} then pos end) "
```
hendrik-obj, line 503:
```suggestion
            f"(select {'min' if is_start else '1 + max'}(case when {{}} then pos end) "
```
not fully tested, but something like this gives the expected `.loc`-like behaviour

hendrik-obj (changes requested):
imo we should keep the postgres behaviour. see comments.

bach/tests/functional/bach/test_series_json.py
thijs-obj: I don't know `.loc` well. How does that handle 'normal' slices with integers? Does that handle it in the same way we do, or should we change that too?

hendrik-obj: But seeing the discussion, there might be something that is more clear for everybody? finding the position of a matching object in an array as a separate function for example? downside is that the syntax will get quite big if you then want to do a slices.

hendrik-obj:
this test should pass in this pr for bq now: `test_objectiv_stack_type3` in `test_series_objectiv.py`. good to keep track of this somehow.

PR #855, opened 2022-06-08
Modelhub bq locationstack

thijs-obj:
Changes:
1. Make `Locationstack.feature_stack` work on BigQuery
2. Temporary work around for `LocationStack.nice_name`. Actual implementation will follow in later PR
3. Use 'events' as default table when running on BigQuery.

Results
With these changes a few of the steps in the `basic-product-analytics.ipynb` notebook work. To test that out simply add something like this to the top of that file:
```python
import os
os.environ['GOOGLE_APPLICATION_CREDENTIALS'] = 'even_the_path_is_a_secret/objectiv-production--bigquery-read-only.json'
os.environ['DSN'] = 'bigquery://objectiv-production/snowplow'
```

### DIFF
--- a/modelhub/modelhub/series/series_objectiv.py
+++ b/modelhub/modelhub/series/series_objectiv.py
@@ -252,13 +252,16 @@ class SeriesLocationStack(SeriesJson):
             return series
 
         @property
-        def nice_name(self) -> 'SeriesJson':
+        def nice_name(self) -> 'SeriesString':
             """
             .. _ls_nice_name:
 
             Returns a nice name for the location stack. This is a human readable name for the data in the
             feature stack.
             """
+            if is_bigquery(self._series_object.engine):
+                # TODO: This is temporary, just so that we have something
+                return self._series_object.json[0].json.get_value('_type', as_str=True) + ' TODO'
             expression = Expression.construct(
                 f"""(
                 select array_to_string(
--- a/modelhub/modelhub/series/series_objectiv.py
+++ b/modelhub/modelhub/series/series_objectiv.py
@@ -4,19 +4,26 @@ Copyright 2021 Objectiv B.V.
 import json
 import os
 
+from typing import List, Generic, TypeVar
 # added for metabase export
 import requests
 
-from bach import DataFrame
+from bach import DataFrame, Series, SeriesString, SeriesJson
 from bach.expression import Expression, quote_string, quote_identifier
-from bach.series import Series
-from bach.series import SeriesJson
 from bach.series.series_json import JsonAccessor
 from bach.types import register_dtype
 from sql_models.util import is_postgres, DatabaseNotSupportedException, is_bigquery
 
 
-class ObjectivStack(JsonAccessor):
+TSeriesJson = TypeVar('TSeriesJson', bound='SeriesJson')
+
+
+class ObjectivStack(JsonAccessor, Generic[TSeriesJson]):
+    """
+    Specialized JsonAccessor that has functions to work on SeriesJsons whose data consists of an array
+    of objects.
+    """
+
     def get_from_context_with_type_series(self, type: str, key: str, dtype='string'):
         """
         .. _get_from_context_with_type_series:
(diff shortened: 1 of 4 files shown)


modelhub/modelhub/series/series_objectiv.py
```diff
@@ -63,6 +70,68 @@ def _bigquery_get_from_context_with_type_series(self, type: str, key: str, dtype
         value_series = ctx_series.json.get_value(key=key, as_str=as_str)
         return value_series.copy_override_dtype(dtype)
 
+    def filter_keys_of_dicts(self, keys: List[str]) -> 'TSeriesJson':
```
thijs-obj, line 73: We might want to move functions like this to SeriesJson.

For now I suggest we just make things work, and after we have clear what interfaces we need we can make this part of Bach, if we want that.

modelhub/modelhub/series/series_objectiv.py
```diff
@@ -63,6 +70,68 @@ def _bigquery_get_from_context_with_type_series(self, type: str, key: str, dtype
...
+            comma_expr = Expression.string_value(', ')
+            json_object_str_expressions.append(key_expr)
+            json_object_str_expressions.append(colon_expr)
+            json_object_str_expressions.append(val_expr)
+            if i < (len(keys) - 1):
+                json_object_str_expressions.append(comma_expr)
+        json_object_key_values_expr = Expression.construct(
+            ', '.join('{}' for _ in json_object_str_expressions),
+            *json_object_str_expressions
+        )
```
KathiaBarahona, line 121: I felt that was a bit difficult to follow since we have a lot of expressions and we kinda need to know when to add a comma.  Maybe building **key: value** expressions first and later join them all?
Something like:
```suggestion
        engine = self._series_object.engine
        key_template_expr = f"{{}} || {quote_string(engine, ':')} "
        unnested_item_helper_series = self._series_object.copy_override(expression=Expression.construct('item'))

        # creates expressions in key: value format
        sub_items_expr = [
            join_expressions(
                [
                    Expression.construct(key_template_expr, Expression.string_value(json.dumps(key))),
                    unnested_item_helper_series.json.get_value(key).expression,
                ],
                join_str=" || ",
            )
            for key in keys
        ]

        # add comma separator between each key-value
        json_object_key_values_expr = join_expressions(sub_items_expr, join_str=f" || {quote_string(engine, ',')} || ")
```

thijs-obj: I get what you are saying, this is difficult to follow. However I found your version even harder to follow.

I have made a new attempt, used some of your improvements (e.g. `join_expressions()`), and added more clarifying comments, but it is essentially the same algorithm and gives the exact same output. What do you think..

This is the commit with the changes: https://github.com/objectiv/objectiv-analytics/pull/855/commits/88e748acea61e48e0ee501d5b302fc0ffc7531fc

thijs-obj:
Please feel free to merge this if __all__ reviewers approve, as I am off tomorrow.

modelhub/modelhub/series/series_objectiv.py
KathiaBarahona: Yes it is, but still. Looks odd to check when to add a comma, as you could just use `join_str` parameter for separating each key-value expression by: `"|| {quote_string(',')} ||"`. I think that my main problem is just having too many individual  expressions that can be grouped into one, I guess that's why it was difficult for me to understand as I had to keep track of the position of each of them.  But its a subjective matter ig, I have no problem to merge it as it is as this is more of a provisional solution and it'll  be moved to bach (soonish, I hope).

hendrik-obj: I think we might want to use `to_json_string`. Then we don't need to bother about the commas etc.

I didn't test it with expressions etc, but the query for `feature_stack` could then look like change from (only the part that creates the array, so not the last part of creating a string from the array):
```sql
ARRAY(select concat('{', """\"_type\"""", """: """, JSON_QUERY(item, '$."_type"'), """, """, """\"id\"""", """: """, JSON_QUERY(item, '$."id"'), '}') from unnest(json_query_array(`location_stack`, '$')) as item)
```
to:
```sql
ARRAY(select to_json_string(struct(JSON_EXTRACT_SCALAR(item, "$._type") as _type, JSON_EXTRACT_SCALAR(item, "$.id") as id)) from unnest(json_query_array(`location_stack`, '$')) as item)
```

hendrik-obj (commented):
looks good, it works. Please have a look at the suggestion in the comment. Could that simplify both the code and the resulting sql?

modelhub/modelhub/series/series_objectiv.py
thijs-obj: I think this is a very good suggestion, and much simpler.

Only concern I'd have is with the `JSON_EXTRACT_SCALAR`, it will only work if the fields are string, numeric, or boolean. If the context contains a nested json object it won't work. But looking at the schema we don't currently have that, and I guess we won't have that? @hendrik-obj 

I'll go ahead and implement your suggestion then

thijs-obj:
Discussion is still a bit open. Will merge this now, as the lack of clarity what works and what queries we generate in which notebook makes it a bit unclear to me what still needs to be tackled and what is done.

Might change the implementation later.

modelhub/modelhub/series/series_objectiv.py
thijs-obj: follow-up issue: https://github.com/objectiv/objectiv-analytics/issues/909

PR #856, opened 2022-06-08
Bach unittest, convert tests to multi-database

thijs-obj:
Changes to bach functional tests:
* Replaced calls to `get_bt_with_test_data()` with calls to `get_df_with_test_data()`, where possible.
* Replaced calls to `get_bt_with_railway_data()` with calls to `get_df_with_railway_data()` ,where possible.
* Got rid of `get_bt_with_food_data()` and replaced calls to it with calls to `get_df_with_food_data()`.

### DIFF
--- a/bach/tests/functional/bach/test_df_setitem.py
+++ b/bach/tests/functional/bach/test_df_setitem.py
@@ -9,9 +9,9 @@ import numpy as np
 
 from bach import SeriesInt64, SeriesString, SeriesFloat64, SeriesDate, SeriesTimestamp, \
     SeriesTime, SeriesTimedelta, Series, SeriesJson, SeriesBoolean
-from tests.functional.bach.test_data_and_utils import get_bt_with_test_data, assert_postgres_type, \
-    assert_equals_data, CITIES_INDEX_AND_COLUMNS, get_bt_with_railway_data, get_df_with_test_data, \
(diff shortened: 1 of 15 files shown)


bach/tests/functional/bach/test_df_setitem.py
```diff
@@ -155,12 +157,16 @@ def test_set_series_column():
     )
     assert bt.duplicated_column == bt['duplicated_column']
 
-    bt['spaces in column'] = bt['founding']
+    spaced_col_name = 'spaces in column'
+    if is_bigquery(engine):
+        # BigQuery doesn't support spaces in column names. So we don't test that functionality on that engine
+        spaced_col_name = 'spaces_in_column'
```
KathiaBarahona, line 163: **question** isn't it better to just do a separate test for this and skip for bq?

KathiaBarahona (approved):
lgtm, just one small question. Nice we are getting rid of those functions

bach/tests/functional/bach/test_df_setitem.py
thijs-obj: I think you are right. Fixed in 3dd06e59830cba9cd4aacd4f935a5c0c465e0567

thijs-obj:
Discussed over chat. Was ok to merge after last change

PR #860, opened 2022-06-09
Make session_gap_seconds a more prominent and required setting

jansentom:
- remove the default in the pipeline
- Add a default value of 1800s back to modelhub

### DIFF
--- a/modelhub/modelhub/modelhub.py
+++ b/modelhub/modelhub/modelhub.py
@@ -135,6 +135,7 @@ class ModelHub:
             start_date=start_date,
             end_date=end_date,
             table_name=table_name,
+            session_gap_seconds=1800,
         )
 
         # sessionized data returns both series as bach.SeriesJson.
--- a/modelhub/modelhub/modelhub.py
+++ b/modelhub/modelhub/modelhub.py
@@ -22,6 +22,7 @@ GroupByType = Union[List[Union[str, bach.Series]], str, bach.Series, NotSet]
 ConversionEventDefinitionType = Tuple[Optional['SeriesLocationStack'], Optional[str]]
 
 TIME_DEFAULT_FORMAT = '%Y-%m-%d %H:%M:%S.%f'
+SESSION_GAP_DEFAULT_SECONDS = 1800
 
 
 class ModelHub:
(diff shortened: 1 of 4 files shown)


modelhub/modelhub/stack/sessionized_data.py
```diff
@@ -40,7 +40,12 @@ class SessionizedDataPipeline(BaseDataPipeline):
         - all context and sessionized series defined in ObjectivSupportedColumns
         - correct dtypes for both context and sessionized series
     """
-    def _get_pipeline_result(self, session_gap_seconds=180, **kwargs) -> bach.DataFrame:
```
jansentom, line 43: So this 180 was a bug, it should have been 1800s by default.

modelhub/modelhub/stack/sessionized_data.py
```diff
@@ -40,7 +40,12 @@ class SessionizedDataPipeline(BaseDataPipeline):
...
+
+    def __init__(self, engine: Engine, table_name: str, session_gap_seconds: int):
```
jansentom, line 44: Put this here as a mandatory param. Could not add it to _get_pipeline_result, because that would break the interface with base ofc.

modelhub/modelhub/modelhub.py
```diff
@@ -135,6 +135,7 @@ def get_objectiv_dataframe(
             start_date=start_date,
             end_date=end_date,
             table_name=table_name,
+            session_gap_seconds=1800,
```
KathiaBarahona, line 138: **suggestion** maybe a constant for it? like we do for time aggregation

KathiaBarahona (approved):
just couple of questions. good catch! lgtm

modelhub/modelhub/modelhub.py
jansentom: makes sense, done.

modelhub/modelhub/stack/sessionized_data.py
```diff
@@ -40,7 +40,12 @@ class SessionizedDataPipeline(BaseDataPipeline):
...
+        super().__init__(engine, table_name)
+        self.session_gap_seconds = session_gap_seconds
```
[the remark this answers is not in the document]
jansentom: not really, fixed.

PR #861, opened 2022-06-09
adjust notebooks to new json slicing

hendrik-obj:
fillna can't be used after json slicing anymore. An empty array is returned instead of None since https://github.com/objectiv/objectiv-analytics/pull/767

### DIFF
--- a/bach/docs/source/example-notebooks/product-analytics.rst
+++ b/bach/docs/source/example-notebooks/product-analytics.rst
@@ -147,9 +147,12 @@ First we define a conversion event in the Objectiv DataFrame.
 
 .. code-block:: python
 
-    modelhub.add_conversion_event(location_stack=df.location_stack.json[{'id': 'objectiv-on-github',
-                                                                         '_type': 'LinkContext'}:].fillna(
-                                                 df.location_stack.json[{'id': 'github', '_type': 'LinkContext'}:]),
+    # Create a columns that extract all location stacks that lead to our github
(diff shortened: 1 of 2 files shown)


bach/docs/source/example-notebooks/product-analytics.rst
```diff
@@ -147,9 +147,12 @@ First we define a conversion event in the Objectiv DataFrame.
 
 .. code-block:: python
 
-    modelhub.add_conversion_event(location_stack=df.location_stack.json[{'id': 'objectiv-on-github',
-                                                                         '_type': 'LinkContext'}:].fillna(
-                                                 df.location_stack.json[{'id': 'github', '_type': 'LinkContext'}:]),
+    # Create a columns that extract all location stacks that lead to our github
+    df['github_press'] = df.location_stack.json[{'id': 'objectiv-on-github', '_type': 'LinkContext'}:]
+    df.loc[df.location_stack.json[{'id': 'github', '_type': 'LinkContext'}:]!=[],'github_press'] = df.location_stack
```
thijs-obj, line 152: Am I understanding correctly that if either of these two is not-empty, then the location stack depicts a location that lead to our github? 1) `df.location_stack.json[{'id': 'objectiv-on-github', '_type': 'LinkContext'}:]` 2) `f.location_stack.json[{'id': 'github', '_type': 'LinkContext'}:]`

hendrik-obj: yes, correct, this works, but we'd like to make this simpler soon

PR #875, opened 2022-06-10
Remove obsolete time format

vard-obj:


modelhub/modelhub/modelhub.py
```diff
@@ -119,9 +119,9 @@ def get_objectiv_dataframe(
             'postgresql://objectiv:@localhost:5432/objectiv' will be used.
         :param table_name: the name of the sql table where the data is stored.
         :param start_date: first date for which data is loaded to the DataFrame. If None, data is loaded from
-            the first date in the sql table. Format as 'YYYY-MM-DD'.
```
hendrik-obj, line 122: I think we can leave this as it was. This is about how you have to format the date when you pass a startdate as an argument. So for example '2022-02-02' (as done in the notebooks)

hendrik-obj (changes requested):
let's leave modelhub.py as it is, then we can merge

hendrik-obj (approved):

PR #876, opened 2022-06-10
Missing description in the doc page

vard-obj:


bach/docs/source/example-notebooks/logistic-regression.rst
```diff
@@ -21,6 +21,17 @@ At first we have to install the open model hub and instantiate the Objectiv Data
 
 Creating a feature set
```
vincenthoogsteder, line 22: @vard-obj small think but shall we also say Creating a feature set to predict user behavior, like in the actual NB?

hendrik-obj (commented):
lgtm

PR #883, opened 2022-06-13
Bach BQ: Remove hack from DataFrame.quantile

KathiaBarahona:
We no longer need the hack for adding the quantile label since the function is no longer generating complex merge subqueries anymore. Tho we do need to add a materialization for BigQuery because it will return duplicated quantiles

### DIFF
--- a/bach/bach/dataframe.py
+++ b/bach/bach/dataframe.py
@@ -2476,21 +2476,31 @@ class DataFrame:
                 index={},
                 group_by=initial_series.group_by,
             )
-            if len(quantiles) == 1:
-                return quantile_df
 
-            # a hack in order to avoid calling quantile_df.materialized().
-            # Currently doing quantile['quantile'] = qt
-            # will raise some errors since the expression is not an instance of AggregateFunctionExpression
-            quantile_df['quantile'] = initial_series\
-                .copy_override_dtype(dtype='float64')\
-                .copy_override(expression=AggregateFunctionExpression.construct(fmt=f'{qt}'))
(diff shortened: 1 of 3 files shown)


bach/bach/operations/describe.py
```diff
@@ -236,7 +236,4 @@ def _calculate_percentiles(self) -> Optional[DataFrame]:
             )
         )
 
-        if is_bigquery(percentile_df.engine):
-            # BigQuery returns quantile per row, need to apply distinct
-            percentile_df = percentile_df.materialize(node_name='describe_quantile', distinct=True)
```
KathiaBarahona, line 241: removed this from describe, DataFrame.quantile should be responsible for this.

bach/bach/dataframe.py
```diff
@@ -2476,21 +2476,31 @@ def quantile(
...
-        from bach.operations.concat import DataFrameConcatOperation
-        result = DataFrameConcatOperation(objects=all_quantile_dfs, ignore_index=True)()
-        # q column should be in the index when calculating multiple quantiles
-        return result.set_index('quantile')
+        final_index = 'quantile'
+
+        if len(quantiles) == 1:
+            # if only one quantile was calculated, then we don't need to add the label on the index
+            result = all_quantile_dfs[0]
+            result = result[[col for col in result.data_columns if col != final_index]]
```
KathiaBarahona, line 2492: we don't return early for this case, since we also need to materialize for BQ

thijs-obj (approved):
lgtm

Issue #884, opened 2022-06-13
[Bach] Implement SeriesJson.json.array_contains for Postgres


KathiaBarahona:
Json accessor has `array_contains` method, implement it for Postgres JsonAccessorImplementation

PR #885, opened 2022-06-13
Bach BQ: Support `SeriesJson.__ge__` for BigQuery

KathiaBarahona:
Created `array_contains` in json accessor. (Still missing PG implementation https://github.com/objectiv/objectiv-analytics/issues/884). Therefore '>=' operator is now supported, still missing '<=' but not sure how yet.

 **Important**: If we want to know if a dict is contained by the array, the order of the keys matters and should match all keys and values, this is mainly because current implementation is based on string comparison (was the fastest way to implement ig). 

Also, IMHO we should limit both operators when the top-level type of the json is a list, since it's not really pythonic the current implementation for Postgres (can find if a sub path is part of the main json path). 
For example:
```
list_a = ['a', 'b', 'c']
list_b = ['b', 'c']

list_a  >= list_b  # True

dict_a = {'a': 'b', 'c': 'd'}
dict_b = {'c': 'd'}
dict_a >= dict_b # will raise a TypeError
```
Therefore, we should create a separate function for knowing if a dict contains another dict. `SeriesJson.json.json_contains_dict` maybe?

### DIFF
--- a/bach/bach/series/series_json.py
+++ b/bach/bach/series/series_json.py
@@ -8,7 +8,7 @@ from typing import Dict, Union, TYPE_CHECKING, Tuple, cast, Optional, List, Any,
 from sqlalchemy.engine import Dialect
 
 from bach.series import Series
-from bach.expression import Expression
+from bach.expression import Expression, join_expressions
 from bach.series.series import WrappedPartition, ToPandasInfo
 from bach.sql_model import BachSqlModel
 from bach.types import DtypeOrAlias, StructuredDtype, AllSupportedLiteralTypes, Dtype
@@ -243,6 +243,10 @@ class SeriesJson(Series):
     def __ge__(self, other: Union['Series', AllSupportedLiteralTypes]) -> 'SeriesBoolean':
         if is_postgres(self.engine):
             return self._comparator_operation(other, "@>")
+        if is_bigquery(self.engine):
+            # there is no operator for "@>", it needs to be simulated
+            return self.json.array_contains(other)
+
         message_override = f'Operator >= is not supported for type json on database {self.engine.name}'
         raise DatabaseNotSupportedException(self.engine, message_override=message_override)
 
@@ -406,6 +410,19 @@ class JsonAccessor(Generic[TSeriesJson]):
         # So for now we have a dedicated len function for arrays.
         return self._implementation.get_array_length()
 
+    def array_contains(self, item: Union['Series', AllSupportedLiteralTypes]) -> 'SeriesBoolean':
+        """
+        Find if an object or list of objects are contained in the array.
+
+        :param item: items to be verified if it is contained by the array.
+
+        :returns: boolean series indicating if the array contains all elements.
+
+        This assumes the top-level item in the json is an array. Will result in an exception (later on) if
+        that's not the case!
+        """
+        return self._implementation.array_contains(item)
+
 
 class JsonBigQueryAccessorImpl(Generic[TSeriesJson]):
     """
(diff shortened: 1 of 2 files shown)


thijs-obj:
Looking at the usage of `>=`. I can only find two usages:
1. https://github.com/objectiv/objectiv-analytics/blob/main/modelhub/modelhub/aggregate.py#L204
2. https://github.com/objectiv/objectiv-analytics/blob/main/modelhub/modelhub/aggregate.py#L255

Both usages check whether the `stack_event_types` column, which is an array of strings, contains one specific string.

**Suggestions:**
* Limit `array_contains()` to just that functionality: checking whether a string appears in a string-array. Perhaps rename it `array_contains_string()` or something like that.
* Call `array_contains()` directly, instead of using the `>=` operator. As far as I know, what we currently have is different from what the `>=` operator does on lists in plain python, so to me it is confusing anyway. 

What do you think?

KathiaBarahona:
> Looking at the usage of `>=`. I can only find two usages:
> 
> 1. https://github.com/objectiv/objectiv-analytics/blob/main/modelhub/modelhub/aggregate.py#L204
> 2. https://github.com/objectiv/objectiv-analytics/blob/main/modelhub/modelhub/aggregate.py#L255
> 
> Both usages check whether the `stack_event_types` column, which is an array of strings, contains one specific string.
> 
> **Suggestions:**
> 
> * Limit `array_contains()` to just that functionality: checking whether a string appears in a string-array. Perhaps rename it `array_contains_string()` or something like that.
> * Call `array_contains()` directly, instead of using the `>=` operator. As far as I know, what we currently have is different from what the `>=` operator does on lists in plain python, so to me it is confusing anyway.
> 
> What do you think?

Totally agree, but we should open a separate PR for migrating and deprecating that behavior. That's why we can't do it on this one

thijs-obj:
> 
> Totally agree, but we should open a separate PR for migrating and deprecating that behavior. That's why we can't do it on this one

okay, I figured since there are only two usages. We can just replace it right now, and be done with it. But I think your approach is good too.

@hendrik-obj or @vard-obj perhaps you can shine a light on how you would like to have this work ultimately?

bach/bach/series/series_json.py
```diff
@@ -524,6 +541,47 @@ def get_array_length(self) -> 'SeriesInt64':
             .copy_override_type(SeriesInt64) \
             .copy_override(expression=expression)
 
+    def array_contains(self, item: Union['Series', AllSupportedLiteralTypes]) -> 'SeriesBoolean':
+        """ For documentation, see implementation in parent class :class:`JsonAccessor` """
```
thijs-obj, line 545:
```suggestion
        """ For documentation, see implementation in class :class:`JsonAccessor` """
```
I realize this is a copy-paste from the wrong docstring that I typed above, still good to have it correct

vard-obj:
> okay, I figured since there are only two usages. We can just replace it right now, and be done with it. But I think your approach is good too.
> 
> @hendrik-obj or @vard-obj perhaps you can shine a light on how you would like to have this work ultimately?

I like the idea of calling `array_contains()` instead of using  `>=` operator.

hendrik-obj:
I think it is good to kill the `>=` and `<=` behaviour as it currently exists. In this pr, the result is that it works quite differnent for postgres than for bq. 
For example, if you look at test `test_json_compare`, it could be written as 
```python
bts = bt.mixed_column >= {"a": "b"}
```
This would pass the same test for postgres still, but gives different results for bigquery even though it doesn't raise errors.

So maybe we should stop supporting completely __le__ and __ge__? Then it is also more pythonic.
because the example in the description of this pr only works for sets anyway, not lists.

Another thing: in pg you should always pass a list when comparing with `>=` in bq that is not needed.

Note that if we do unsupport __le__ and __ge__, we need to make some changes in modelhub and a notebook

PR #887, opened 2022-06-13
Bach BQ: Remove `COALESCE` from `JsonBigQueryAccessorImpl.get_value`

KathiaBarahona:
The coalesce is generating big query expressions as sometimes we might have expressions for nested jsons.

### DIFF
--- a/bach/bach/series/series_json.py
+++ b/bach/bach/series/series_json.py
@@ -505,8 +505,8 @@ class JsonBigQueryAccessorImpl(Generic[TSeriesJson]):
         if not as_str:
             return self._series_object.copy_override(expression=json_expression)
         from bach.series import SeriesString
-        # if accessed value is a json, this will return NULL since JSON_VALUE
-        # is used only for scalars
+        # JSON_QUERY will return a string containing valid json. i.e. if the queried value is a string, then
+        # it will be quoted. Here we remove the quotes
         value_expression = Expression.construct(
             f'REGEXP_REPLACE({{}}, {{}}, {{}})',
             json_expression,
--- a/bach/bach/series/series_json.py
+++ b/bach/bach/series/series_json.py
@@ -505,13 +505,17 @@ class JsonBigQueryAccessorImpl(Generic[TSeriesJson]):
         if not as_str:
             return self._series_object.copy_override(expression=json_expression)
         from bach.series import SeriesString
+        # if accessed value is a json, this will return NULL since JSON_VALUE
+        # is used only for scalars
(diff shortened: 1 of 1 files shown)


bach/bach/series/series_json.py
```diff
@@ -505,13 +505,17 @@ def get_value(self, key: str, as_str: bool = False) -> Union['SeriesString', 'TS
...
+        # if accessed value is a json, this will return NULL since JSON_VALUE
+        # is used only for scalars
         value_expression = Expression.construct(
-            '''JSON_VALUE({}, '$."{}"')''',
-            self._series_object,
-            Expression.raw(key)
+            f'REGEXP_REPLACE({{}}, {{}}, {{}})',
+            json_expression,
+            Expression.raw('r\'(^"|"$)\''),
+            Expression.string_value(''),
```
KathiaBarahona, line 514: @thijs-obj Decided to just replace the start/end quotes just for having the same behavior as we have with Postgres. If we just return NULL when the value to be extracted is a json, then we need to add a condition on the tests... so I'm not sure if there's another way to go. Also tried `FORMAT('%t', ....)` and `TRIM(...,  '"')` but didn't worked for some edge cases (when having nested quotes).

bach/bach/series/series_json.py
```diff
@@ -505,13 +505,17 @@ def get_value(self, key: str, as_str: bool = False) -> Union['SeriesString', 'TS
         if not as_str:
             return self._series_object.copy_override(expression=json_expression)
         from bach.series import SeriesString
+        # if accessed value is a json, this will return NULL since JSON_VALUE
```
thijs-obj, line 508: Comment doesn't make sense, as we don't use JSON_VALUE anymore

bach/bach/series/series_json.py
```diff
@@ -505,13 +505,17 @@ def get_value(self, key: str, as_str: bool = False) -> Union['SeriesString', 'TS
...
+        # is used only for scalars
```
thijs-obj, line 509:
```suggestion
        # JSON_QUERY will return a string containing valid json. i.e. if the queried value is a string, then
        # it will be quoted. Here we remove the quotes
```

thijs-obj (approved):
Left a suggestion for a comment, please have a look to see what you think.

Code is okay tho :+1:

bach/bach/series/series_json.py
KathiaBarahona: yeh, was doing the opposite. Forgot to remove that

PR #889, opened 2022-06-14
Show more results for better demonstration logistic regression

hendrik-obj:
`head(20)`


vincenthoogsteder (approved):
LGTM, very nice demo results like this

PR #895, opened 2022-06-14
[ModelHub] Decorator for limiting series usage in Map and Aggregate methods

KathiaBarahona:
Created `use_only_required_objectiv_series` decorator. The main purpose is to limit the series to be used in the function, this way we avoid getting expressions from unused series in the result's basenode. I'm still missing unit tests, but seems to work.

### DIFF
--- a/modelhub/modelhub/aggregate.py
+++ b/modelhub/modelhub/aggregate.py
@@ -9,7 +9,7 @@ from bach.series import Series
 from sql_models.constants import NotSet, not_set
 from typing import List, Union, TYPE_CHECKING
 
-from modelhub.utils import use_only_required_objectiv_series
+from modelhub.util import use_only_required_objectiv_series
 
 if TYPE_CHECKING:
     from modelhub import ModelHub
@@ -58,8 +58,6 @@ class Aggregate:
                              groupby: Union[List[Union[str, Series]], str, Series],
                              column: str,
                              name: str):
-
-        self._mh._check_data_is_objectiv_data(data)
 
         data = self._check_groupby(data=data,
                                    groupby=groupby,
--- a/modelhub/modelhub/aggregate.py
+++ b/modelhub/modelhub/aggregate.py
@@ -1,6 +1,7 @@
 """
 Copyright 2021 Objectiv B.V.
 """
+from enum import Enum
 from typing import List, Union
 
 import bach
@@ -8,6 +9,7 @@ from bach.series import Series
 from sql_models.constants import NotSet, not_set
 from typing import List, Union, TYPE_CHECKING
 
+from modelhub.utils import use_only_required_objectiv_series
 
 if TYPE_CHECKING:
     from modelhub import ModelHub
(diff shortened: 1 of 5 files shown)


thijs-obj (commented):
Had a first look

I like that we don't just call `_check_data_is_objectiv_data()` to check the whole DataFrame everytime. But instead check that the actual columns that we need for that specific function are in the DataFrame.

I am not a fan of declaring what columns each function expects in `util.py` tho. I think that information should be kept wit the function that it applies to. If in each function we did something like `data = data[['session_id', 'user_id', 'moment']]` we would more or less achieve the same thing. And I think that would be much simpler.

Most important question I think is: did we test whether this improves performance? I can imagine that sometimes removing columns from a DataFrame and then materializing it, also leads to dataframes having different base_nodes down the line and thus needing more merges to bring data back together again?

Also good to sync with @hendrik-obj on this. As his branch https://github.com/objectiv/objectiv-analytics/pull/904 does something related

KathiaBarahona:
> I am not a fan of declaring what columns each function expects in `util.py` tho.

 I agree,  ended up removing the dictionary and passing the required series as a param. Here https://github.com/objectiv/objectiv-analytics/pull/895/commits/ac4cfc1ccf7bf153a9a1cd82b6bc3cb67a830528


> Most important question I think is: did we test whether this improves performance? I can imagine that sometimes removing columns from a DataFrame and then materializing it, also leads to dataframes having different base_nodes down the line and thus needing more merges to bring data back together again?

Improvements are in terms of query complexity, let's say we have a bunch of expressions using location stack functionalities. Such expressions are not required anymore if user needs only the calculated series. In terms of merge, well I noticed that the original column_expressions from the base_node used before calculating the series actually remain. The improvement here are in child nodes, the amount of column expressions is reduced. Test comparing done here: https://github.com/objectiv/objectiv-analytics/pull/895/commits/5892008580271fdd2ae551100fd63d9f84a8265c 
But in overall, this is not a solution for avoiding merges, merges will be always performed because calculations involve lots of windowing functions and materializations at the end of the function. 

> Also good to sync with @hendrik-obj on this. As his branch #904 does something related

Will take a look to this.

vard-obj (approved):
lgtm

modelhub/modelhub/decorators.py
```diff
@@ -0,0 +1,123 @@
...
+                columns_to_check=required_series,
+                check_index=True,
+                check_dtypes=True,
+                with_md_dtypes=True,
+            )
+
+            extra_series = _get_extra_series_to_include_from_params(
+                func, data, include_series_from_params, *args, **kwargs,
+            )
+            series_to_include = list(set(required_series or {}) | set(extra_series))
```
thijs-obj, line 67: Didn't test this, but I think this should be
```suggestion
            series_to_include = set(required_series or {}) | set(extra_series)
```
The next line is `data = data[series_to_include]`, and in `DataFrame.__getitem__()` we use the order of the columns in a list, but if we pass a set we respect the original order. I think it would be weird if this decorator changes the order of the columns

thijs-obj (commented):
Left a single comment, nothing further

PR #900, opened 2022-06-16
Deprecate le and ge for json Series

hendrik-obj:


bach/tests/functional/bach/test_series_json.py
```diff
@@ -69,26 +69,29 @@ def test_json_get_single_value(engine, dtype):
     assert a == {'a': 'b', 'c': {'a': 'c'}}
 
 
-# TODO: __le__ for BigQuery
-def test_json_compare_lists__ge__(engine, dtype):
+def test_json_array_contains(engine, dtype):
     bt = get_df_with_json_data(engine=engine, dtype=dtype)
-    bts = bt.list_column >= [{"c": "d"}]
+
+    # for bigquery order of keys should be the same as in the data
```
thijs-obj, line 75: Question: should we not just limit support to strings, boolean and numbers? These dicts in bigquery are really tricky. If the key order is different it fails. It might also fail if there is a space or a tab somewhere (e.g. for nice formatting). Just seems really tricky

thijs-obj (approved):
I think the change is good, but I am wondering whether we should not simplify this further. Hence the question I left.

thijs-obj (approved):

bach/tests/functional/bach/test_series_json.py
hendrik-obj: agreed. changed in 1935cfb7ed97ce139e701fe49f1a6229eadab048

PR #901, opened 2022-06-16
Support temporary tables for bigquery

thijs-obj:
Changes:
* Support the TEMP_TABLE materialization for BigQuery
* Add parameter `materialization` to materialize(): We allow 'CTE' (default) and 'TEMP_TABLE'.
  * Allowing real tables and views might be tricky. Running `head()` or `to_pandas()` would then have permanent side effects.


bach/sql_models/model.py
```diff
@@ -59,6 +60,13 @@ def is_statement(self) -> bool:
     def is_cte(self) -> bool:
         return self.value.is_cte
 
+    @property
+    def has_lasting_effect(self) -> bool:
+        """ Return true if the materialization is a permanent table or view."""
+        # TODO: make these functions consistent, e.g. have a field in value for this, or remove those from
```
thijs-obj, line 66: I will do this in a follow-up PR, IF this PR gets merged. Because changing this will change the hashes of SqlModels, which will change quite some test results.

thijs-obj:
As this might prove helpful for complex analysis. I am submitting this for review.

Please have a look @hendrik-obj @KathiaBarahona @jansentom

bach/tests/functional/bach/test_df_materialize.py
```diff
@@ -1,19 +1,23 @@
...
 from sql_models.graph_operations import get_graph_nodes_info
+from sql_models.model import Materialization
 from sql_models.util import is_bigquery, is_postgres
 from tests.functional.bach.test_data_and_utils import assert_equals_data, get_df_with_test_data, \
     get_df_with_json_data
 
 
 @pytest.mark.parametrize("inplace", [False, True])
-def test_materialize(inplace: bool, engine):
+@pytest.mark.parametrize("materialization", [Materialization.CTE, 'temp_table'])
```
jansentom, line 19: Bit strange to mix constant and string notation here unless you really want to test both here, but in that case you should do all variations?
```suggestion
@pytest.mark.parametrize("materialization", [Materialization.CTE,Materialization.TEMP_TABLE])
```

jansentom (approved):
Nice, LGTM

bach/bach/dataframe.py
```diff
@@ -990,6 +990,7 @@ def materialize(
         inplace=False,
         limit: Any = None,
         distinct: bool = False,
+        materialization: Union[Materialization, str] = Materialization.CTE
```
KathiaBarahona, line 993: mmm isn't it easier just to accept string values? I mean it's a public method, so not sure if users will be aware from which module they can get the Materialization Enum

bach/bach/dataframe.py
```diff
@@ -2123,7 +2134,13 @@ def view_sql(self, limit: Union[int, slice] = None) -> str:
             variable_values=self.variables
         )
         model = update_placeholders_in_graph(start_node=model, placeholder_values=placeholder_values)
-        return to_sql(dialect=self.engine.dialect, model=model)
+        sql_statements = to_sql_materialized_nodes(
+            dialect=self.engine.dialect, start_node=model
+        )
+        sql_statements = [sql_stat for sql_stat in sql_statements
+                          if not sql_stat.materialization.has_lasting_effect]
```
KathiaBarahona, line 2141: maybe?
```suggestion
        sql_statements = to_sql_materialized_nodes(
            dialect=self.engine.dialect, start_node=model, with_lasting_effect=False,
        )
```

bach/sql_models/sql_generator.py
```diff
@@ -110,7 +123,10 @@ def _materialize(dialect: Dialect, sql_query: str, model: SqlModel) -> str:
     if materialization == Materialization.TABLE:
         return f'create table {quoted_name} as {sql_query}'
     if materialization == Materialization.TEMP_TABLE:
-        return f'create temporary table {quoted_name} on commit drop as {sql_query}'
+        if is_postgres(dialect):
+            return f'create temporary table {quoted_name} on commit drop as {sql_query}'
+        if is_bigquery(dialect):
+            return f'CREATE TEMP TABLE {quoted_name} as {sql_query}'
```
KathiaBarahona, line 129:
```suggestion
        if is_postgres(dialect):
            return f'create temporary table {quoted_name} on commit drop as {sql_query}'
        if is_bigquery(dialect):
            return f'CREATE TEMP TABLE {quoted_name} as {sql_query}'
        raise DatabaseNotSupportedException(dialect, f'Cannot create temporary table for {dialect.value}')
        
```

KathiaBarahona (commented):
Left a couple of comments just code-related.

KathiaBarahona (commented):
**question** Are we running the create temp table query even if we already created it or tell BigQuery not to run the script again? For example a normal pattern is:
```
df.materialize(inplace=True, materialization='temp_table', node_name='initial_data')
df.head() # Runs query for creating temp table for initial data
# more stuff not changing the materialized node
.... 
df.head() # Runs again query for creating temp table for initial data
```

bach/bach/savepoints.py
```diff
@@ -225,15 +225,15 @@ def get_create_statements(self, dialect: Dialect) -> Dict[str, str]:
...
+            if sql_stat.materialization in (Materialization.TABLE, Materialization.VIEW)
         }
 
-    def to_sql(self, dialect: Dialect) -> Dict[str, str]:
+    def to_sql(self, dialect: Dialect) -> List[GeneratedSqlStatement]:
         """
         Generate the sql for all save-points
         :param dialect: SQL Dialect
-        :return: dictionary mapping the name of each savepoint to the sql for that savepoint.
+        :return: List of GeneratedSql statements, each representing one savepoint
```
KathiaBarahona, line 236:
```suggestion
        :return: List of GeneratedSqlStatement, each representing one savepoint
```

bach/sql_models/model.py
```diff
@@ -47,6 +47,7 @@ class Materialization(Enum):
     QUERY = MaterializationType('query', True, True)
     VIEW = MaterializationType('view', True, False)
     TABLE = MaterializationType('table', True, False)
+    """ TEMP TABLE is a temporary table that is limited to the current session. """
```
KathiaBarahona, line 50: what do you mean with `limited to the current session`? I mean BigQuery temp tables are created on a separate dataset and are available for 24hrs

bach/bach/dataframe.py
KathiaBarahona: also, shouldn't `to_sql_materialized_nodes` return a str instead of a list of statements? I think this function should have the same return type as `to_sql`

thijs-obj:
> **question** Are we running the create temp table query even if we already created it or tell BigQuery not to run the script again? For example a normal pattern is:
> 
> ```
> df.materialize(inplace=True, materialization='temp_table', node_name='initial_data')
> df.head() # Runs query for creating temp table for initial data
> # more stuff not changing the materialized node
> .... 
> df.head() # Runs again query for creating temp table for initial data
> ```

In this case we will create two temporary tables. The input data might have changed between those two calls, so the tables might be different. BigQuery will automatically clean-up the temporary tables after 24 hours[1], so from our point of view creating those temporary tables doesn't really have side effects, other than the query we are running right now. Bach/SqlModels forgets it created a temporary table directly after `head()` or `to_pandas()` finishes, and that is fine.

If we want to properly reuse tables, we should create permanent tables or materialized views. But then things become more complex, because I don't think we want to do that as part of `head()`. I think in that case the user needs to take explicit control of that, e.g. through a function `create_table()`  or something like that*. I was typing a whole essay here about why that shouldn't be part of `head()`/`to_pandas()`, but maybe we should discuss that separately. 

[1] https://cloud.google.com/bigquery/docs/multi-statement-queries#temporary_tables
[*] We built something called savepoints for this, but it's not really great, so I think we need to change the ideas a bit there.

bach/tests/functional/bach/test_df_materialize.py
thijs-obj: Yeah I wanted to test both cte and temp-table, and both string and enum parameters. I figured this is a nice trade-off between test coverage and not completely exploding the number of tests

bach/bach/dataframe.py
thijs-obj: Maybe. We have another function in DataFrame that also accepts both, so that's why I did it like this: consistency. I am pretty sure there was a discussion about it back then too :thinking: 

Myself, I actually really like using the `Enum` parameter type when working in e.g. PyCharm, but at the same time I dislike it when working in a notebook without auto-import. This added complexity is a bit shitty, but very localized. In the end I just copied what we did in the `set_savepoint` function.

bach/bach/dataframe.py
```diff
@@ -1015,6 +1023,9 @@ def materialize(
         index_dtypes = {k: v.instance_dtype for k, v in self.index.items()}
         series_dtypes = {k: v.instance_dtype for k, v in self.data.items()}
         node = self.get_current_node(name=node_name, limit=limit, distinct=distinct)
+        materialization = Materialization.normalize(materialization)
+        assert materialization in (Materialization.CTE, Materialization.TEMP_TABLE)
```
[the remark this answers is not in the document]
thijs-obj: `Materialization` is in the `sql_models` sub-project, and I think it's not nice to have functionality in that project to describe what Bach's `DataFrame.materialize()` supports. For a 'pure' dependency relation, Bach should use sql-models, but sql-models shouldn't have knowledge about Bach.

Of course we could have a function in sql-models that's not related to Bach, that happens to execute this check. But in the end, I think this check is here, because here we decide that we want to support these two and e.g. not `TABLE`. We could easily support `TABLE` too, but it just doesn't make sense for Bach `DataFrame.materialize()` to support that, I think at least

bach/bach/dataframe.py
thijs-obj: `to_sql_materialized_nodes()` is also used in another place, where we use the meta data. If we returned a string we would have to change a lot of things

bach/bach/savepoints.py
thijs-obj: Done in a8d085b4

bach/sql_models/model.py
thijs-obj: Good point, changed this in a8d085b4 to:

```python
    """
    TEMP TABLE is a temporary table that is limited to the current session, or a table that is guaranteed to
    be cleaned up by the database at some later time.
    """
```

bach/bach/dataframe.py
thijs-obj: I did think about adding an `with_lasting_effect` parameter or something like that. But I didn't see a really convincing reason for that to be better than do the filtering here, so I went with this :man_shrugging:

bach/sql_models/model.py
thijs-obj: https://github.com/objectiv/objectiv-analytics/pull/944

hendrik-obj (commented):
lgtm. nice addition. see comment, maybe do the notebook separately?

notebooks/basic-product-analytics.ipynb
```diff
@@ -136,7 +136,13 @@
...
     "df['utm_campaign'] = df.global_contexts.gc.get_from_context_with_type_series(type='MarketingContext', key='campaign')\n",
     "df['utm_content'] = df.global_contexts.gc.get_from_context_with_type_series(type='MarketingContext', key='content')\n",
-    "df['utm_term'] = df.global_contexts.gc.get_from_context_with_type_series(type='MarketingContext', key='term')"
+    "df['utm_term'] = df.global_contexts.gc.get_from_context_with_type_series(type='MarketingContext', key='term')\n",
+    "\n",
+    "# We'll do a lot of operation on this data. To make this easier for the database (especially BigQuery),\n",
+    "# we tell Bach to materialize the current DataFrame as temporary table. This has no effect now, but any later\n",
+    "# queries that build on this DataFrame will consists of two queries: one to create a temporary table, and one that\n",
+    "# queries that table and does operations on that\n",
+    "df = df.materialize(materialization='temp_table')"
```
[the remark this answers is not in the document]
thijs-obj: Sounds like a good plan. I'll take out this change.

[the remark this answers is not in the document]
thijs-obj: Done in 367b1a7

PR #903, opened 2022-06-17
ModelHub: Calculate dependent conversion series at once

hendrik-obj:
reuse add columns to data for multiple functions in the model hub. now for all internal model hub functions joins shouldn't be necessary anymore.


KathiaBarahona:
## Description:

Created `_get_calculated_conversion_df` method in `Map` class. It is in charged of calculating all series needed for any of: `is_conversion_event`, `conversions_in_time`, `conversions_counter` and `pre_conversion_hit_number`. For example, if we need to calculate `pre_conversion_hit_number`, it will return a dataframe including:

`__is_conversion_event`, `__conversions_in_time`,` __pre_conversion_hit_number`. The caller is in charge of getting the correct series, casting to right type and assigning final name.

modelhub/modelhub/aggregate.py
```diff
@@ -250,21 +250,38 @@ def top_product_features_before_conversion(self,
...
+
+        # get conversion information
+        # conversions_counter, is_conversion_event, conversions_in_time
+        conversions_df = self._mh.map._get_calculated_conversion_df(
+            series_to_calculate=_CalculatedConversionSeries.CONVERSIONS_COUNTER,
+            data=events_with_user_interactions,
+            name=name,
+            partition='session_id',
+        )
+
```
KathiaBarahona, line 270: Not sure how to avoid this, making `_get_calculated_conversion_df` public is a bit nasty as its just an internal helper and  not intended to be exposed...

modelhub/modelhub/aggregate.py
```diff
@@ -250,23 +250,45 @@ def top_product_features_before_conversion(self,
...
+        data = data.materialize(node_name='extract_application_and_feature_nice_name')
+
+        # reduce the amount of data to be considered
+        groupby_col = ['__application', '__feature_nice_name', 'event_type']
+        data = data.dropna(subset=groupby_col)
+
+        # select only user interactions
+        events_with_user_interactions = data[
+            data.stack_event_types.json.array_contains(event_type)
+        ]
```
KathiaBarahona, line 262: @hendrik-obj should consider this filter till the end right? Probably this will yield different results from expected

modelhub/modelhub/aggregate.py
```diff
@@ -250,23 +250,45 @@ def top_product_features_before_conversion(self,
         else:
             data['__feature_nice_name'] = data.location_stack.ls.nice_name
 
+        data = data.materialize(node_name='extract_application_and_feature_nice_name')
+
+        # reduce the amount of data to be considered
+        groupby_col = ['__application', '__feature_nice_name', 'event_type']
+        data = data.dropna(subset=groupby_col)
```
KathiaBarahona, line 257: actually didn't help

modelhub/modelhub/aggregate.py
```diff
@@ -243,29 +243,45 @@ def top_product_features_before_conversion(self,
...
+        if location_stack is not None:
+            converted_users_filtered['__feature_nice_name'] = location_stack.ls.nice_name
+        else:
+            converted_users_filtered['__feature_nice_name'] = (
+                converted_users_filtered.location_stack.ls.nice_name
+            )
+
+        converted_users_filtered.materialize(
+            node_name='extract_application_and_feature_nice_name', inplace=True,
+        )
```
KathiaBarahona, line 282: moved this after all conversion calculation. I think it's better to do heavy stuff after we have filtered the data

modelhub/modelhub/aggregate.py
```diff
@@ -243,28 +243,43 @@ def top_product_features_before_conversion(self,
...
-            converted_users.stack_event_types.json.array_contains(event_type)
-        ]
+        data = data[data.stack_event_types.json.array_contains(event_type)]
+
+        # merge with filtered conversion interactive events
+        converted_users_filtered = data.merge(conversions_df, on='event_id')
+        converted_users_filtered['__application'] = converted_users_filtered.global_contexts.gc.application
+
+        if location_stack is not None:
+            converted_users_filtered['__feature_nice_name'] = location_stack.ls.nice_name
```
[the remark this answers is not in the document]
KathiaBarahona: naaaah, df will do it

modelhub/modelhub/map.py
```diff
@@ -193,34 +221,173 @@ def pre_conversion_hit_number(self,
...
+        """
+        Generates a dataframe containing all required conversion series based on series_to_calculate param.
+        All new series will contain private names.
+        """
         self._mh._check_data_is_objectiv_data(data)
 
-        data = data.copy()
-        data['__conversions'] = self._mh.map.conversions_in_time(data, name=name)
+        # consider only required series for calculations
+        all_required_series = ['event_type', 'session_id', 'moment', 'session_id', 'session_hit_number']
```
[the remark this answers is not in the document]
hendrik-obj: also, if we not do this suggestion, 'session_id' is in there twice, so 1 can be removed.

hendrik-obj (commented):
nice improvements, one comment for a possible optimization

modelhub/modelhub/map.py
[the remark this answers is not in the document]
hendrik-obj: mm i tested it with a clean df, and only setting the `add_conversion_event`. In that case it seems to be faster. However, if you run the entire notebook (where the df contains a number of additional columns), the query of the last cell doesn't execute anymore because of the complexity. So it is a bit of a trade off, but seems like your initial option is the way to go for now.

modelhub/modelhub/map.py
```diff
@@ -179,34 +213,173 @@ def pre_conversion_hit_number(self,
...
+        """ For documentation, see function :meth:`Map.conversions_in_time()` """
+        series_to_calculate = _CalculatedConversionSeries.CONVERSIONS_IN_TIME
+        self._check_conversion_dependencies(data, series_to_calculate)
+
+        is_conversion_event_column = _CalculatedConversionSeries.IS_CONVERSION_EVENT.private_name
+        column_name_to_add_conversion_counter = '__conversion_counter'
+        data[column_name_to_add_conversion_counter] = 0
+        data.loc[data[is_conversion_event_column], column_name_to_add_conversion_counter] = 1
+
+        if is_bigquery(data.engine):
```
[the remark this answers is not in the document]
KathiaBarahona: Fixed here https://github.com/objectiv/objectiv-analytics/pull/903/commits/55aa1a969503c6b5bcc83aacb013d715dc35bb3b

modelhub/modelhub/aggregate.py
```diff
@@ -243,28 +243,43 @@ def top_product_features_before_conversion(self,
...
             raise ValueError('Conversion event label is not provided.')
 
-        data['__application'] = data.global_contexts.gc.application
-
-        if location_stack is not None:
-            data['__feature_nice_name'] = location_stack.ls.nice_name
-        else:
-            data['__feature_nice_name'] = data.location_stack.ls.nice_name
+        # temporary workaround, we should not call private constants and methods from Map
+        # replace this after having a better solution
```
thijs-obj, line 247: Question: It says temporary here, and `replace this after having a better solution`. Do we have a plan for a better solution already?

modelhub/modelhub/map.py
```diff
@@ -85,25 +117,24 @@ def is_conversion_event(self, data: bach.DataFrame, name: str) -> bach.SeriesBoo
...
-            series = data.event_type == conversion_event
-        elif conversion_event is None:
-            series = conversion_stack.json.get_array_length() > 0
-        else:
-            series = ((conversion_stack.json.get_array_length() > 0) & (data.event_type == conversion_event))
-        return series.copy_override(name='is_conversion_event')
+        series = conversion_df[series_to_calculate.private_name]
+        return (
+            series
+            .copy_override(name=series_to_calculate.public_name, index=data.index)
```
thijs-obj, line 130: Here we set `index=data.index`, we do the same in `conversions_counter`, but not in `conversions_in_time`

Question: Does the df returned by `_get_calculated_conversion_df()` not already have this set correctly?. If so it would be nicer to not set it again here, now it's just confusing me me :)

modelhub/modelhub/map.py
```diff
@@ -85,25 +117,24 @@ def is_conversion_event(self, data: bach.DataFrame, name: str) -> bach.SeriesBoo
...
+            .copy_override_type(bach.SeriesBoolean)
```
thijs-obj, line 131: Question: Is this just to help mypy, or is this actually needed runtime too?

thijs-obj (commented):
Left some questions. No big comments. Fine to merge as is, but please look at questions and answer

modelhub/modelhub/aggregate.py
KathiaBarahona: Not yet, I don't have a better idea yet. I just feel  that having that method public might cause confusions on users :woman_shrugging:

modelhub/modelhub/map.py
KathiaBarahona: I probably forgot to remove this when doing it, `_get_calculated_conversion_df` must return the correct index. Will fix it

modelhub/modelhub/map.py
KathiaBarahona: help mypy. There is no way for the package to know that that specific series is a series boolean as the typing of `_get_calculated_conversion_df` is just a bach.DataFrame

modelhub/modelhub/map.py
KathiaBarahona: Removed here https://github.com/objectiv/objectiv-analytics/pull/903/commits/936bbec95cbbd9e7087e6e469945ef59f77196e1

PR #916, opened 2022-06-20
Modelhub location stack nice name

### LINKED ISSUE #910: Implement nice_name for BQ

thijs-obj:
**Change:** Implement `LocationStack.nice_name` for BigQuery.

Other changes: Small fixes to naming and such.

This resolves #910


hendrik-obj (approved):
lgtm, postgres implementation is a bit different now from bigquery. they could be made more similar in structure. something for a followup pr?

thijs-obj:
> They could be made more similar in structure. something for a followup pr?

Do you think this is something we should do? I sort of figured, both implementations work, and they do the same so that's good for now and the foreseeable future

PR #917, opened 2022-06-20
Bach fix for `.json[negative index]` and other small improvements

thijs-obj:
Changes:
* Bugfix: previously `.json[-5]` would give an error on BigQuery if `-5` was not a valid index. Whereas on Postgres, and with positive indexes it gives `NULL`. Now with non-existing negative indexes it also gives NULL
* Slightly changed implementation for negative indexes: reverse the array instead of using array-length to calculate positive index. This way the json-series is only referenced once in the expression instead of twice, which is an improvement if the expression is non-trivial.
* Optimized test: single query vs three queries previously

Of course changing both the tests and the implementation in a single PR is not a best practice. But I figured since I only changed one of the implementations (Postgres is unchanged) and the change is small, it would be easier to have a in single PR.

### DIFF
--- a/bach/bach/series/series_json.py
+++ b/bach/bach/series/series_json.py
@@ -525,14 +525,16 @@ class JsonBigQueryAccessorImpl(Generic[TSeriesJson]):
 
     def get_array_item(self, key: int) -> 'TSeriesJson':
         """ For documentation, see implementation in parent class :class:`JsonAccessor` """
-        if key >= 0:
-            expression = Expression.construct(f'''JSON_QUERY({{}}, '$[{key}]')''', self._series_object)
-            return self._series_object.copy_override(expression=expression)
-        # case key < 0
-        # BigQuery doesn't (yet) natively support this, so we emulate this.
-        array_len = self.get_array_length()
-        expr_offset = Expression.construct(f'OFFSET({{}} {key})', array_len)
-        expression = Expression.construct('JSON_QUERY_ARRAY({})[{}]', self._series_object, expr_offset)
+        if key < 0:
+            # BigQuery doesn't (yet) natively support this, so we emulate this by reversing the array
+            key = -key - 1
+            array_expr = Expression.construct(
+                "'[' || ARRAY_TO_STRING(ARRAY_REVERSE(JSON_QUERY_ARRAY({})), ', ') || ']'",
+                self._series_object
+            )
+        else:
+            array_expr = self._series_object.expression
+        expression = Expression.construct(f'''JSON_QUERY({{}}, '$[{key}]')''', array_expr)
         return self._series_object.copy_override(expression=expression)
 
     def get_dict_item(self, key: str) -> 'TSeriesJson':
--- a/bach/bach/series/series_json.py
+++ b/bach/bach/series/series_json.py
@@ -532,14 +532,13 @@ class JsonBigQueryAccessorImpl(Generic[TSeriesJson]):
         """
         if key < 0:
             # BigQuery doesn't (yet) natively support this, so we emulate this by reversing the array
-            key = -key - 1
-            array_expr = Expression.construct(
-                "'[' || ARRAY_TO_STRING(ARRAY_REVERSE(JSON_QUERY_ARRAY({})), ', ') || ']'",
+            key = abs(key)
+            expression = Expression.construct(
+                f'ARRAY_REVERSE(JSON_QUERY_ARRAY({{}}))[SAFE_ORDINAL({key})]',
                 self._series_object
             )
         else:
-            array_expr = self._series_object.expression
-        expression = Expression.construct(f'''JSON_QUERY({{}}, '$[{key}]')''', array_expr)
+            expression = Expression.construct(f'''JSON_QUERY({{}}, '$[{key}]')''', self._series_object)
         return self._series_object.copy_override(expression=expression)
 
     def get_dict_item(self, key: str) -> 'TSeriesJson':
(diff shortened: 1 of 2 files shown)


bach/tests/functional/bach/test_series_json.py
```diff
@@ -134,45 +134,24 @@ def test_json_compare(engine, dtype):
...
-            [2, None],
-            [3, {"_type": "SectionContext", "id": "top-10"}],
-            [4, None]
-        ]
-    )
-    bts = bt.mixed_column.json["a"]
+    bt = bt[['mixed_column']]
+    bt['get_0'] = bt.mixed_column.json[0]
+    bt['get_min2'] = bt.mixed_column.json[-3]
+    bt['get_min5'] = bt.mixed_column.json[-5]  # -5 doesn't exist, we expect to get `None`
```
thijs-obj, line 141: This gave an error previously. Something I saw while working on the `nice_name` PR yesterday

bach/tests/functional/bach/test_series_json.py
```diff
@@ -134,45 +134,24 @@ def test_json_compare(engine, dtype):
...
+    bt['get_a'] = bt.mixed_column.json["a"]
+    bt = bt.drop(columns=['mixed_column'])
+    print(bt.view_sql())
```
vard-obj, line 144: I guess this print is not needed.

bach/bach/series/series_json.py
```diff
@@ -524,15 +524,22 @@ def _find_in_json_list(self, filtering_slice: Dict[str, str], is_start: bool):
...
+        if key < 0:
+            # BigQuery doesn't (yet) natively support this, so we emulate this by reversing the array
+            key = -key - 1
+            array_expr = Expression.construct(
+                "'[' || ARRAY_TO_STRING(ARRAY_REVERSE(JSON_QUERY_ARRAY({})), ', ') || ']'",
+                self._series_object
+            )
+        else:
+            array_expr = self._series_object.expression
+        expression = Expression.construct(f'''JSON_QUERY({{}}, '$[{key}]')''', array_expr)
```
KathiaBarahona, line 542: I think we can simplify the expression even more by just converting the json into array once and stringify the array with `FORMAT` function and using `SAFE_OFFSET` for avoiding raising errors for index out of range.
```suggestion
        array_expr = f'JSON_QUERY_ARRAY({{}})'
        if key < 0:
            # BigQuery doesn't (yet) natively support this, so we emulate this by reversing the array
            key = -key - 1
            array_expr = f'ARRAY_REVERSE({array_expr})'

        access_array_expr = f'{array_expr}[SAFE_OFFSET({key})]'
        expression = Expression.construct(f"FORMAT('%s', {access_array_expr})", self._series_object)
```

bach/tests/functional/bach/test_series_json.py
KathiaBarahona, line 144:
```suggestion
```

KathiaBarahona (changes requested):

bach/tests/functional/bach/test_series_json.py
thijs-obj: ah yes. will remove this.

KathiaBarahona: nvm, just saw @vard-obj commented about it https://github.com/objectiv/objectiv-analytics/pull/917#discussion_r902403008

bach/bach/series/series_json.py
```diff
@@ -524,15 +524,22 @@ def _find_in_json_list(self, filtering_slice: Dict[str, str], is_start: bool):
...
-        expression = Expression.construct('JSON_QUERY_ARRAY({})[{}]', self._series_object, expr_offset)
+        """
+        Returns an item from the json array.
+        The key is treated as a 0-based index. If negative this will count from the end of the array (one
+            based). If the index does not exist this will render None/NULL.
+        This assumes the top-level item in the json is an array
+        """
+        if key < 0:
+            # BigQuery doesn't (yet) natively support this, so we emulate this by reversing the array
+            key = -key - 1
```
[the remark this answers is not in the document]
thijs-obj: ack

bach/tests/functional/bach/test_series_json.py
thijs-obj: doh, forgot about this. Will remove it

bach/bach/series/series_json.py
thijs-obj: Aah that's brilliant, I didn't know about `safe_offset`/`safe_ordinal`

bach/tests/functional/bach/test_series_json.py
thijs-obj: f3164869ec7cbfe971611443a465710ea756690c

bach/bach/series/series_json.py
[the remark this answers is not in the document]
thijs-obj: f3164869ec7cbfe971611443a465710ea756690c

bach/bach/series/series_json.py
KathiaBarahona: true, no need for format

KathiaBarahona (approved):
lgtm, tho I think we are missing a test were we compare if results from using a positive offset are yielding the same as negative offsets if both point to the same position. 
For example: 
```python
list_a = [0, 1, 2, 3]
assert list_a[0] == list_a[-4]
```

thijs-obj:
> lgtm, tho I think we are missing a test were we compare if results from using a positive offset are yielding the same as negative offsets if both point to the same position. For example:
> 
> ```python
> list_a = [0, 1, 2, 3]
> assert list_a[0] == list_a[-4]
> ```

Added a test case in https://github.com/objectiv/objectiv-analytics/pull/917/commits/9e50aef2e8f90fb96ab5809aa03e09049b624bcd

vard-obj (approved):

PR #918, opened 2022-06-21
Fix `requiresContext` bug

borft:
Fixes a bug where `requiresContext` would be persisted between validation of different events, leading to incorrectly invalidating events because of missing contexts

use deepcopy() for initialising the list of required contexts, to prevent adding any to the original schema

### DIFF
--- a/backend/objectiv_backend/schema/event_schemas.py
+++ b/backend/objectiv_backend/schema/event_schemas.py
@@ -101,7 +101,7 @@ class EventSubSchema:
             raise ValueError(f'Not a valid event_type {event_type}')
 
         event_types: Set[EventType] = {event_type}
-        context_types: Set[ContextType] = set(self.schema[event_type].get('requiresContext', []))
+        context_types: Set[ContextType] = set(deepcopy(self.schema[event_type].get('requiresContext', [])))
         parents: List[EventType] = self.schema[event_type].get('parents', [])
         for parent in parents:
             parent_event_types, parent_context_types = \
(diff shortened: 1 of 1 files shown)


backend/tests/schema/test_schema.py
```diff
@@ -202,3 +202,20 @@ def test_add_context_with_optionals_set():
     assert marketing_context['creative_format'] == context_vars['creative_format']
     assert 'marketing_tactic' in marketing_context
     assert marketing_context['marketing_tactic'] == context_vars['marketing_tactic']
+
+
+def test_required_context_broken_state():
```
thijs-obj, line 207: What does this test, test exactly? If I run this test against the code in `main` it also passes

backend/objectiv_backend/schema/event_schemas.py
```diff
@@ -125,7 +125,7 @@ def get_all_parent_event_types(self, event_type: EventType) -> Set[EventType]:
         """
         if not self.is_valid_event_type(event_type):
             raise ValueError(f'Not a valid event_type {event_type}')
-        return self._compiled_all_parents_and_required_contexts[event_type][0]
+        return {e for e in self._compiled_all_parents_and_required_contexts[event_type][0]}
```
thijs-obj, line 128: This seems like a good change, much safer this way. Perhaps we should do this for all functions that return a mutable object? 
Quick list:
* `list_event_types`
* `ContextSubSchema.list_context_types`
* `ContextSubSchema.get_all_parent_context_types`
* `ContextSubSchema.get_all_required_context_types`
* `ContextSubSchema.get_all_child_context_types`

Might have missed some

thijs-obj (commented):
Left a question and comment

backend/tests/schema/test_schema.py
borft: interesting, it doesn't for me; it fails with missing contexts. The problem is that the list of required contexts in the schema gets polluted by an individual validation call. So what happens is, that the first ApplicationLoadedEvent has a LocationContext, that requires a PathContext. The second ApplicationLoadedEvent doesn't have one, and as such shouldn't require the PathContext.

If the schema gets altered, the seconds case doesn't pass.

thijs-obj: totally my bad. I was testing with old schema files and then it didn't trigger.

This test works as you intended

thijs-obj (approved):
LGTM

backend/objectiv_backend/schema/event_schemas.py
borft: makes sense, fixed those as well

Issue #909, opened 2022-06-20
simplify sql for extracting key-values from objects


hendrik-obj:
Apply suggesion from https://github.com/objectiv/objectiv-analytics/pull/855#discussion_r896096485

thijs-obj:
Closing as https://github.com/objectiv/objectiv-analytics/pull/920 has been merged

PR #920, opened 2022-06-21
Simplify _bigquery_filter_keys_of_dicts as discussed

thijs-obj:
Addresses https://github.com/objectiv/objectiv-analytics/issues/909

before
```sql
select `_index_event_id` as `_index_event_id`, '[' || ARRAY_TO_STRING(ARRAY(select concat('{', """\"_type\"""", """: """, JSON_QUERY(item, '$."_type"'), """, """, """\"id\"""", """: """, JSON_QUERY(item, '$."id"'), '}') from unnest(json_query_array(`location_stack`, '$')) as item), ', ') || ']' as `b` 
from `loaded_data___039d63a201b8851fa7521f6ed9100a90` 
```

after
```sql
select `_index_event_id` as `_index_event_id`, TO_JSON_STRING(ARRAY(
                SELECT STRUCT(JSON_VALUE(item, '$."_type"') as `_type`, JSON_VALUE(item, '$."id"') as `id`)
                FROM UNNEST (JSON_QUERY_ARRAY(`location_stack`, '$')) AS item WITH OFFSET AS pos
                ORDER BY pos)) as `b` 
from `loaded_data___039d63a201b8851fa7521f6ed9100a90` 
```

### DIFF
--- a/modelhub/modelhub/series/series_objectiv.py
+++ b/modelhub/modelhub/series/series_objectiv.py
@@ -74,6 +74,8 @@ class ObjectivStack(JsonAccessor, Generic[TSeriesJson]):
         """
         Return a new Series, that consists of the same top level array, but with all the sub-dictionaries
         having only their original fields if that field is listed in the keys parameter.
+
+        The fields of the sub-dictionaries MUST be scalar values (i.e. string, number, or boolean).
         """
         engine = self._series_object.engine
         if is_postgres(engine):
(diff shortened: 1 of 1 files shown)


hendrik-obj (approved):
lgtm

modelhub/modelhub/series/series_objectiv.py
```diff
@@ -98,44 +100,29 @@ def _postgres_filter_keys_of_dicts(self, keys: List[str]) -> 'TSeriesJson':
...
+                '''JSON_VALUE(item, '$."{}"') as {}''',
+                Expression.raw(key),
+                Expression.identifier(key)
+            )
+            object_entries_expressions.append(entry_expr)
+        expression = Expression.construct(
+            '''TO_JSON_STRING(ARRAY(
+                SELECT STRUCT({})
+                FROM UNNEST (JSON_QUERY_ARRAY({}, '$')) AS item WITH OFFSET AS pos
+                ORDER BY pos))''',
```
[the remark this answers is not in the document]
thijs-obj: I think so:
https://cloud.google.com/bigquery/docs/reference/standard-sql/arrays#:~:text=Because%20UNNEST%20destroys%20the%20order,the%20rows%20by%20their%20offset.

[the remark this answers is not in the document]
hendrik-obj: ah thanks

modelhub/modelhub/series/series_objectiv.py
```diff
@@ -74,6 +74,8 @@ def filter_keys_of_dicts(self, keys: List[str]) -> 'TSeriesJson':
         """
         Return a new Series, that consists of the same top level array, but with all the sub-dictionaries
         having only their original fields if that field is listed in the keys parameter.
+
+        The fields of the sub-dictionaries MUST be scalar values (i.e. string, number, or boolean).
```
KathiaBarahona, line 78: Mm yes and no, seems that if its a non-scalar value it will return NULL instead.

https://cloud.google.com/bigquery/docs/reference/standard-sql/json_functions#json_value
```text
If json_path returns a JSON null or a non-scalar value (in other words, if json_path refers to an object or an array), then a SQL NULL is returned.
```

KathiaBarahona (commented):
A small comment regards docstrings.

Also, I have two concerns regards results yield by BigQuery and Postgres. Seems we get different behaviors for the next two cases:
1. `key` is not present. Postgres will not include the key in the filtered result, meanwhile BigQuery does, it will set it with a null value.
2. When the `location_stack` is null (shouldn't happen ofc), Postgres will keep the NULL value, but BigQuery returns an empty array

Postgres Results:
image

Bigquery Results:
image

IMHO all engines should yield same results.

thijs-obj:
> A small comment regards docstrings.
> 
> Also, I have two concerns regards results yield by BigQuery and Postgres. Seems we get different behaviors for the next two cases:
> 
> 1. `key` is not present. Postgres will not include the key in the filtered result, meanwhile BigQuery does, it will set it with a null value.
> 2. When the `location_stack` is null (shouldn't happen ofc), Postgres will keep the NULL value, but BigQuery returns an empty array
> 
> Postgres Results: image
> 
> Bigquery Results: image
> 
> IMHO all engines should yield same results.


Good points, nice research. This behaviour should be the same. I'll have a look at this.

thijs-obj:
Okay, a bit lame solution, but I just removed the function `filter_keys_of_dicts` and used the database specific implementations directly in `LocationStack.feature_stack()`.

In a LocationStack both the `_type` and `id` keys should always be present. So the behaviour should be the same for both engines. Also I don't think we can expect LocationStacks to be None. Since we only added `filter_keys_of_dicts()` very recently, I think it is safe to assume that no end-user is depending on its existence. So this, however lame, should solve the problems

Please have a look @KathiaBarahona

modelhub/modelhub/series/series_objectiv.py
```diff
@@ -253,8 +186,66 @@ def feature_stack(self) -> 'SeriesLocationStack':
             and a `id` key.
             """
             keys = ['_type', 'id']
```
KathiaBarahona, line 188: **question**: why not defining a constant as class attribute? also, I don't really understand the purpose of this property... feels like we can just hardcore the column expressions if the keys are always the same. IMHO  both functions should be in bach.

KathiaBarahona: or even just extract them in `ExtractedContextPipeline` :woman_shrugging:

thijs-obj:
Discussed on video call:
* Merging this PR
* Created new issue for future improvements: https://github.com/objectiv/objectiv-analytics/issues/928

Issue #912, opened 2022-06-20
simplify array_contains generated query for bq


hendrik-obj:
since https://github.com/objectiv/objectiv-analytics/pull/900 passing lists with multiple items to search is not supported anymore. Therefore, the sql in array_contains for bq can be greatly simplified.

thijs-obj:
Closing as https://github.com/objectiv/objectiv-analytics/pull/927 has been merged

PR #927, opened 2022-06-22
Simplify array_contains for BQ

thijs-obj:
This resolves issue #912

SQL before:
```sql
(CASE WHEN (`column` is not null) THEN (select array_length(array(select searching_value from UNNEST(JSON_QUERY_ARRAY("""[\"a\"]""")) as searching_value where searching_value in UNNEST(JSON_QUERY_ARRAY(`column`))))) END = 1) as `a`
```

SQL after
```sql
IF(`column` IS NULL, NULL, (SELECT "\"a\"" IN UNNEST(JSON_QUERY_ARRAY(`column`)))) as `a`
```

### DIFF
--- a/bach/bach/series/series_json.py
+++ b/bach/bach/series/series_json.py
@@ -409,8 +409,7 @@ class JsonAccessor(Generic[TSeriesJson]):
         :param item: item to be verified if it is contained by the array. Only supports scalar values.
         :returns: boolean series indicating if the array contains the element.
 
-        This assumes the top-level item in the json is an array. Will result in an exception (later on) if
-        that's not the case!
+        This assumes the top-level item in the json is an array.
         """
 
         return self._implementation.array_contains(item)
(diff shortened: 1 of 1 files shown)


KathiaBarahona (approved):
nice! lgtm

bach/bach/series/series_json.py
```diff
@@ -583,33 +582,17 @@ def get_array_length(self) -> 'SeriesInt64':
 
     def array_contains(self, item: Union[int, float, bool, str]) -> 'SeriesBoolean':
```
KathiaBarahona, line 583: I think we should also consider `None`. Since we can have smth like: `['a', None]`, when loading this as a CTE, it will be:
`"""[\"a\", null]"""`

So doing smth like: `IF(column IS NULL, NULL, (SELECT """null""" IN UNNEST(JSON_QUERY_ARRAY(column)))) ` is valid

vard-obj: agree, taking into account `None` is important.

vard-obj (approved):

bach/bach/series/series_json.py
thijs-obj: Fixed this in https://github.com/objectiv/objectiv-analytics/pull/926

thijs-obj: And merged that branch into this one

thijs-obj:
I later realized that this implementation doesn't actually work well for floats. Added an issue to the backlog to look at that at some point in the future: https://github.com/objectiv/objectiv-analytics/issues/934 Not the highest priority I guess

PR #931, opened 2022-06-23
WIP: prototype Bq materialization prototype with data fix

thijs-obj:
This reverts commit 6f1a9b17df97c581f9d07d7642f60684ac744d36.


modelhub/modelhub/stack/extracted_contexts.py
```diff
@@ -143,6 +144,43 @@ def _get_base_dtypes(self) -> Mapping[str, StructuredDtype]:
         }
 
     def _get_initial_data(self) -> bach.DataFrame:
+        if is_bigquery(self._engine):
+            # TODO: this is a total hack as a proof of concept
+            extracted_data = CustomSqlModelBuilder(
+                '''
+                  SELECT
+                      CAST(`contexts_io_objectiv_taxonomy_1_0_0`[OFFSET(0)].`event_id` AS STRING) AS event_id,
```
thijs-obj, line 152: Note that we here use the `event_id` from `contexts_io_objectiv_taxonomy_1_0_0` and not the `event_id` column that is also in the table. For past data (somewhere before April 2022) we have rows where those event_ids do not match.

modelhub/modelhub/stack/extracted_contexts.py
```diff
@@ -98,15 +99,15 @@ def __init__(self, engine: Engine, table_name: str):
...
-        # check if table has all required columns for pipeline
-        dtypes = bach.from_database.get_dtypes_from_table(
-            engine=self._engine,
-            table_name=self._table_name,
-        )
-        self._validate_data_dtypes(
-            expected_dtypes=dict(self._get_base_dtypes()),
-            current_dtypes=dtypes,
-        )
+        # # check if table has all required columns for pipeline
```
thijs-obj, line 102: This is actually completely unrelated to the fix. Just had this outcommented to save a few seconds as I ran this very often

PR #939, opened 2022-06-24
ModelHub BQ: Drop rows with duplicated ids

KathiaBarahona:
Removed rows with duplicated event_ids, now modelhub will consider only the first event based on time.

### DIFF
--- a/modelhub/modelhub/stack/extracted_contexts.py
+++ b/modelhub/modelhub/stack/extracted_contexts.py
@@ -209,7 +209,15 @@ class ExtractedContextsPipeline(BaseDataPipeline):
         # this materialization is to generate a readable query
         df_cp = df_cp.materialize(node_name='bq_extra_processing')
 
-        # remove duplicated event_ids
+        # Remove duplicated event_ids
+        # Unfortunately, some events might share an event ID due to
+        # browser pre-cachers or scraping bots sending the same event multiple time. Although,
+        # legitimate clients might try to send the same events multiple times,
+        # in an attempt to make sure that events do not get lost in case of connection problems
+        # and/or app reloads. We can easily recognize these events as they'll have non-unique event-ids.
+        # In all cases we are only interested in the first event. On postgres we achieve this by having a
+        # primary key index on event-id. On BigQuery such indexes are not possible. Instead, we here filter
+        # out duplicate event-ids, keeping the first event with each id.
         df_cp = df_cp.drop_duplicates(subset=['event_id'], sort_by=['time'], keep='first')
 
         # BQ data source has no moment and day columns, therefore we need to generate them
--- a/modelhub/modelhub/stack/extracted_contexts.py
+++ b/modelhub/modelhub/stack/extracted_contexts.py
@@ -203,8 +203,14 @@ class ExtractedContextsPipeline(BaseDataPipeline):
         if is_postgres(self._engine):
             return df_cp
 
+        # remove taxonomy column, no longer needed
+        df_cp = df_cp.drop(columns=[self._taxonomy_column.name])
+
         # this materialization is to generate a readable query
-        df_cp = df_cp.materialize(node_name='bq_moment_day_extraction')
+        df_cp = df_cp.materialize(node_name='bq_extra_processing')
+
+        # remove duplicated event_ids
+        df_cp = df_cp.drop_duplicates(subset=['event_id'], sort_by=['time'], keep='first')
 
         # BQ data source has no moment and day columns, therefore we need to generate them
         # based on the time value from the taxonomy column
--- a/bach/bach/dataframe.py
+++ b/bach/bach/dataframe.py
@@ -2839,15 +2839,19 @@ class DataFrame:
 
             if dedup_sort:
                 df = df.sort_values(by=dedup_sort, ascending=ascending)
-            window = df.groupby(by=dedup_on).window(end_boundary=end_boundary, end_value=None)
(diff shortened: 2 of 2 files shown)


modelhub/modelhub/stack/extracted_contexts.py
```diff
@@ -203,8 +203,14 @@ def _apply_extra_processing(self, df: bach.DataFrame) -> bach.DataFrame:
...
 
+        # remove taxonomy column, no longer needed
+        df_cp = df_cp.drop(columns=[self._taxonomy_column.name])
+
         # this materialization is to generate a readable query
-        df_cp = df_cp.materialize(node_name='bq_moment_day_extraction')
+        df_cp = df_cp.materialize(node_name='bq_extra_processing')
+
+        # remove duplicated event_ids
+        df_cp = df_cp.drop_duplicates(subset=['event_id'], sort_by=['time'], keep='first')
```
thijs-obj, line 213: nice that we can leverage bach here :+1:

modelhub/modelhub/stack/extracted_contexts.py
```diff
@@ -203,8 +203,14 @@ def _apply_extra_processing(self, df: bach.DataFrame) -> bach.DataFrame:
...
             return df_cp
 
+        # remove taxonomy column, no longer needed
+        df_cp = df_cp.drop(columns=[self._taxonomy_column.name])
+
         # this materialization is to generate a readable query
-        df_cp = df_cp.materialize(node_name='bq_moment_day_extraction')
+        df_cp = df_cp.materialize(node_name='bq_extra_processing')
+
+        # remove duplicated event_ids
```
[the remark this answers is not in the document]
thijs-obj:
```suggestion
        # Remove duplicated event_ids
        # Legitimate clients might send the same events multiple times, in an attempt to make sure that
        # events do not get lost in case of connection problems and/or app reloads. We can easily recognize
        # these events as they'll have non-unique event-ids. Additionally, some scraping bots will send the
        # same event multiple times because they cache javascript results.
        # In all cases we are only interested in the first event. On postgres we achieve this by having a
        # primary key index on event-id. On BigQuery such indexes are not possible. Instead, we here filter
        # out duplicate event-ids, keeping the first event with each id.
```

Just an idea. Could well imagine that this can be shortened significantly. But I think it's fair to conclude that why we need to do this for BigQuery but not Postgres is not obvious, so deserves some explanation

thijs-obj (commented):
Comment about comments. Otherwise good for me

modelhub/modelhub/stack/extracted_contexts.py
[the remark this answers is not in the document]
KathiaBarahona: Sure, tho maybe in a future we should dedicate a doc just for this issue and add the reference to it in the docstring. Because I find having this just commented in some random module of the entire project is like  a needle in a haystack and its not just an issue with modelhub as itself.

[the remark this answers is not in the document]
KathiaBarahona: Addressed here https://github.com/objectiv/objectiv-analytics/pull/939/commits/45af3ad500cebc7acedffeddabff9838b2b3c1f9

Issue for the doc page: https://github.com/objectiv/objectiv-analytics/issues/942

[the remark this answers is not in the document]
thijs-obj: Yes, I very much agree with you. Until we have proper documentation let's document it here, but let's work on having proper documentation

thijs-obj (approved):

modelhub/modelhub/stack/extracted_contexts.py
```diff
@@ -203,8 +203,22 @@ def _apply_extra_processing(self, df: bach.DataFrame) -> bach.DataFrame:
...
+        # Remove duplicated event_ids
+        # Unfortunately, some events might share an event ID due to
+        # browser pre-cachers or scraping bots sending the same event multiple time. Although,
+        # legitimate clients might try to send the same events multiple times,
+        # in an attempt to make sure that events do not get lost in case of connection problems
+        # and/or app reloads. We can easily recognize these events as they'll have non-unique event-ids.
+        # In all cases we are only interested in the first event. On postgres we achieve this by having a
+        # primary key index on event-id. On BigQuery such indexes are not possible. Instead, we here filter
+        # out duplicate event-ids, keeping the first event with each id.
+        df_cp = df_cp.drop_duplicates(subset=['event_id'], sort_by=['time'], keep='first')
```
hendrik-obj, line 221: checked this manually now. do we need a test that tests that duplicated rows are actually removed?

hendrik-obj (commented):
lgtm. one comment about possible test

modelhub/modelhub/stack/extracted_contexts.py
KathiaBarahona: mmm we could but its basically re testing what bach already tests :thinking:

thijs-obj: > mmm we could but its basically re testing what bach already tests thinking

To be fair: we could say that about a lot of code in the modelhub: in the end we mostly use bach functionality. Doesn't mean we shouldn't test that the combination of statements we have crafted here actually does what we want that combination of statements to do.

thijs-obj: I didn't comment anything about tests before, as I think merging this has a very high priority. But if we ever get a chance to look at this again in the future, I think testing this makes _a lot_ of sense

PR #940, opened 2022-06-27
Experiment: Zero deps trackers

sdirosa:
Currently our Trackers have two dependencies:
- fast-deep-equal
- UUID

In this PR I've tried to remove both of them by switching to alternate methods. 

**fast-deep-equal**
For `fast-deep-equal`, I decided to use instead JSON.stringify.
The main rationale here is that this function was used to compare React's useEffect dependencies in some helper hooks. Fast-deep-equal worked fine but, for what we need to do, using JSON to stringify both objects and compare the results works equally as well and it's actually faster. No compromise here.

**UUID**
We rely on the well known [uiid](https://github.com/uuidjs/uuid) package. 
The main issue with it is that it relies on `crypto`. This has proved unreliable in a number of cases:
- Google Bot is known to run in an environment (maybe they have their own implementation, or some caching going on) that does **not** generate random values. We have had several duplicates popping in at regular intervals. Something that should never happen.
- React Native doesn't have `crypto` at all. Which means we have to preload a polyfill to implement the same API.
- Latest Jest is not compatible with the uuid package unless we use a custom resolver or some other workaround.

I checked lots of discussion, including comments from the author of UUID, on what would be a decent compromise to generate valid UIID v4 with very low chances of collisions and without having to rely on crypto. 

I also checked the source code of libraries like [nanoid](https://github.com/ai/nanoid) (they have a non-secure version of the library that doesn't rely on crypto) and what several crypto polyfills do under the hood.

In most cases it seems to boil down to using a combination of timestamps + one or more Math.random calls and some base 16 or base 32 conversions. 

I've written a custom implementation that does just that. It's very simple:

```
function generateUUID()  {
  const rng = Date.now().toString(16) + Math.random().toString(16) + '0'.repeat(16);
  return [rng.substring(0, 8), rng.substring(8, 12), '4000-8' + rng.substring(13, 16), rng.substring(16, 28)].join('-');
};
```

Check the code for more info on the limitations of this approach.

### DIFF
--- a/tracker/core/react/package.json
+++ b/tracker/core/react/package.json
@@ -68,8 +68,7 @@
   },
   "dependencies": {
     "@objectiv/schema": "^0.0.21-4",
-    "@objectiv/tracker-core": "~0.0.21-4",
-    "fast-deep-equal": "^3.1.3"
+    "@objectiv/tracker-core": "~0.0.21-4"
   },
   "peerDependencies": {
     "react": ">=16.8",
--- a/tracker/core/react/src/hooks/useOnChange.ts
+++ b/tracker/core/react/src/hooks/useOnChange.ts
@@ -3,7 +3,6 @@
  */
 
 import { useEffect, useRef } from 'react';
-import isEqual from 'fast-deep-equal';
 import { OnChangeEffectCallback } from '../types';
 
 /**
@@ -16,7 +15,7 @@ export const useOnChange = <T>(state: T, effect: OnChangeEffectCallback<T>) => {
   latestEffectRef.current = effect;
 
   useEffect(() => {
-    if (!isEqual(previousStateRef.current, state)) {
+    if (JSON.stringify(previousStateRef.current) !== JSON.stringify(state)) {
       effect(previousStateRef.current, state);
       previousStateRef.current = state;
     }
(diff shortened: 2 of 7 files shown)


sdirosa:
@jansentom Could you take a look and share your thoughts. This is not urgent at all, but just something I've been thinking about for a while.

In general, generating UUIDs in FE is not really reliable anyway, as explained above in the description. I would say that if we really want them being 100% unique, then we should probably generate them in the Collector before storing the events. That said, I think the implementation here is probably more than enough for us.

As FE is concerned, these identifiers are useful exclusively for debugging. E.g. when the Tracker sends an invalid event, the Collector can point out which event exactly was problematic in the error message it sends back.
For all we care they could be just be a timestamp with an incremental value appended to it.

Curious about your thoughts.

sdirosa:
@miike Interesting for you as well to take a look at

miike:
>I would say that if we really want them being 100% unique

I don't think these need to be unique tbh - and the compute / memory overhead of doing so just exceeds the benefit in doing so.

>We have had several duplicates popping in at regular intervals. Something that should never happen.

Googlebot has a weird implementation of RNG (which will apply to both rng and crypto) you can pretty much expect to see duplicates here even with a PRNG component just due to their API implementation, Googlebot is the most common but their are other crawlers / bots that do similar things.

If the intent is to generate a new UUID I'd almost be tempted to something like UUIDv7 (v6-v8 are still in draft) as I think it retains some of the same properties while helping to conform to the draft standard and some new time ordering features that can be quite nice.

sdirosa:
Closing this experiment. Relevant changes have been implemented in other PRs.

PR #944, opened 2022-06-27
Bq materialization test follow up

thijs-obj:
Addresses a TODO from earlier branch https://github.com/objectiv/objectiv-analytics/pull/901/files#r902742839

Change:
* For `SqlModel._calculate_hash()` now use the Materialization's unique type-name, instead of the Materialization type's full tuple of properties. This means that if we ever add/remove a property the hashes don't change again.
* All tests that (indirectly) check the hash
* Made implementation of `Materialization.has_lasting_effect` consistent with other properties.

### DIFF
--- a/bach/sql_models/model.py
+++ b/bach/sql_models/model.py
@@ -75,7 +75,7 @@ class Materialization(Enum):
         return self.value.has_lasting_effect
 
     @classmethod
-    def get_by_name(cls, type_name) -> 'Materialization':
+    def get_by_type_name(cls, type_name) -> 'Materialization':
         for mat in cls:
             if mat.type_name == type_name:
                 return mat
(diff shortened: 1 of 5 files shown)


bach/sql_models/model.py
```diff
@@ -35,25 +35,32 @@ class MaterializationType(NamedTuple):
...
-    name: str
+    type_name: str
     is_statement: bool
     is_cte: bool
+    has_lasting_effect: bool
 
 
 class Materialization(Enum):
-    CTE = MaterializationType('cte', False, True)
+    CTE = MaterializationType('cte', False, True, False)
```
KathiaBarahona, line 47: **suggestion**: can we use keywords for the booleans? There are more than 3 parameters, so kinda gets a bit difficult to follow.

KathiaBarahona (approved):
lgtm

thijs-obj:
Review comment has been applied in https://github.com/objectiv/objectiv-analytics/pull/944/commits/c09cf142e48ebdb99ad800973bf45a15c95362ea. Merging this

PR #953, opened 2022-06-28
Bach: SeriesString.str.replace

KathiaBarahona:


bach/tests/functional/bach/test_series_string.py
```diff
@@ -110,3 +110,24 @@ def test_get_dummies(engine) -> None:
...
         ],
     )
+
+
+def test_string_replace(engine) -> None:
+    bt = get_df_with_test_data(engine)
+    municipality = bt['municipality'].sort_index()
+    assert isinstance(municipality, SeriesString)
+
+    municipalities = municipality.to_numpy()
```
thijs-obj, line 120:
```suggestion
    municipalities = ['Leeuwarden', 'Súdwest-Fryslân', 'Súdwest-Fryslân']
```
Doing `to_numpy()` involves an extra query, which takes ~1.5 seconds for BigQuery. Considering in how many places we have hardcoded the expected values already, I don't think it's bad to do here, and it does safe run-time (and personally, I find it more readable actually)

bach/tests/functional/bach/test_series_string.py
```diff
@@ -110,3 +110,24 @@ def test_get_dummies(engine) -> None:
...
+    replace_pat = ['-', 'west', 'ee', 'ú', 'â']
```
thijs-obj, line 121:
```suggestion
    replace_pat = ['-', 'west', 'ee', 'ú', 'â', 'e']
```
Having at least one case where we replace more than one occurrence in the string seems like a good idea to me

thijs-obj (commented):
Code looks good to me. Some comments on the tests

bach/tests/functional/bach/test_series_string.py
KathiaBarahona: Done here https://github.com/objectiv/objectiv-analytics/pull/953/commits/110bcaa49477d0797201c744fa268b548198d474

PR #954, opened 2022-06-28
Bach BQ: Use total_seconds in `SeriesTimedelta.mode` for BigQuery

KathiaBarahona:
`APPROX_TOP_COUNT` does not support type INTERVAL. Had to use total seconds instead and convert it back into timdelta.


thijs-obj (approved):
LGTM

PR #962, opened 2022-06-29
Also support temp_table in Series.materialize() for consistency

thijs-obj:
We added `materialization` parameter to `DataFrame.materialize()` (this PR: https://github.com/objectiv/objectiv-analytics/pull/901), this adds the same parameter to `Series.materialize()`.

Probably a lot less useful on a Series than on DataFrames. But I ran into it not being possible yesterday while trying to get a notebook to work. Which was annoying, as it's so simple to support.

Please have a look @vard-obj @KathiaBarahona


vard-obj (approved):
lgtm

PR #966, opened 2022-06-29
Model hub BQ: Make sampling work for open-taxonomy notebook

thijs-obj:


notebooks/open-taxonomy-how-to.ipynb
```diff
@@ -372,7 +372,10 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "df_sample = df.get_sample(table_name='sample_data', sample_percentage=10, overwrite=True)"
+    "# table_name = 'sample_data'\n",
+    "table_name = 'objectiv-production.writable_dataset.sample_data'\n",
```
thijs-obj, line 376: I created a dataset to which an account has write access. This account also has read access to the `objectiv-production.snowplow` dataset that contains the base data[*]. I figured we don't want to give this account write access to the dataset that contains the actual base data.

The 'problem' is that we then need to specify the full table name here, otherwise it won't work. But that means the code diverges between Postgres and BigQuery. Not really a big deal, I think. But not sure how we should handle that here? Do we leave a comment that tells the user how to do this with BigQuery and by default make it work with Postgres?

Of course on Postgres we also assume the user has write access

[*] I have communicated all my dataset and authorization changes to Mike

thijs-obj:
Going to close this PR. This just works on BigQuery, but the user does need to specify a full table name that he has write access to. We don't want to change the notebook tho, as that should work out of the box with Postgres

PR #974, opened 2022-07-01
replace hist function with internal bach

hendrik-obj:
- [ ] replace use of pandas histogram function with bach's for all notebooks / examples


vard-obj (approved):
lgtm

PR #976, opened 2022-07-04
bach: Add `DataFrame.database_create_table()`

thijs-obj:
Added `DataFrame.database_create_table()` to easily create a database table with the current dataframe.

`database_create_table()` vs `savepoints`:
The functionality in this PR duplicates some functionality in `df.savepoints`. But upon reflection I think Savepoints is just too complex. Savepoints was meant to built whole graphs of tables/views/queries and being able to write them all to the database, or in the future export them all to BI tools, or even export them as a graph to tools like dbt. But that makes it rather complex to use, imho. The interface should be very simple for end-users for simple things like creating (potentially very big) tables.

### DIFF
--- a/bach/bach/dataframe.py
+++ b/bach/bach/dataframe.py
@@ -959,7 +959,7 @@ class DataFrame:
         Changes to data, index, sorting, grouping etc. on the copy will not affect the original.
         The savepoints on the other hand will be shared by the original and the copy.
 
-        If you want to create a snapshot of the data, have a look at :py:meth:`get_sample()`
+        If you want to create a snapshot of the data, have a look at :py:meth:`database_create_table()`
 
         Calling `copy(df)` will invoke this copy function, i.e. `copy(df)` is implemented as df.copy()
 
@@ -1113,6 +1113,48 @@ class DataFrame:
         """
         from bach.sample import get_unsampled
         return get_unsampled(df=self)
+
+    def database_create_table(self, table_name: str, overwrite: bool = False, ) -> 'DataFrame':
+        """
+        Write the current state of the DataFrame to a database table.
+
+        The DataFrame that's returned will query from the written table for any further operations.
+
+        :param table_name: Name of the table to write to. Can include project_id and dataset
+                on BigQuery, e.g. 'project_id.dataset.table_name'
+        :param overwrite: if True, the sample data is written to table_name, even if that table already
+            exists.
+
+        :return: New DataFrame; the base_node consists of a query on the newly created table.
+
+        .. note::
+            This function queries the database.
+        .. note::
+            This function writes to the database.
+        """
+        dialect = self.engine.dialect
+        model = self.get_current_node(name='database_create_table')
+        model = model.copy_set_materialization(Materialization.TABLE)
+        model = model.copy_set_materialization_name(materialization_name=table_name)
+
+        placeholder_values = get_variable_values_sql(dialect=dialect, variable_values=self.variables)
+        model = update_placeholders_in_graph(start_node=model, placeholder_values=placeholder_values)
+
+        sql = to_sql(dialect=dialect, model=model)
+        with self.engine.connect() as conn:
+            if overwrite:
+                sql = f'DROP TABLE IF EXISTS {quote_identifier(dialect, table_name)}; {sql}'
+
+            sql = escape_parameter_characters(conn, sql)
+            conn.execute(sql)
+
+        all_dtypes = {**self.index_dtypes, **self.dtypes}
+        return self.from_table(
+            engine=self.engine,
+            table_name=table_name,
+            index=list(self.index.keys()),
+            all_dtypes=all_dtypes
+        )
 
     @overload
     def __getitem__(self, key: str) -> 'Series':
--- a/bach/tests/functional/bach/test_df_database_create.py
+++ b/bach/tests/functional/bach/test_df_database_create.py
@@ -1,0 +1,56 @@
+"""
+Copyright 2022 Objectiv B.V.
+"""
+from random import randrange
+
+import pytest
+
+from tests.functional.bach.test_data_and_utils import get_df_with_test_data, assert_equals_data
+from sql_models.sql_generator import to_sql
+from sql_models.util import quote_identifier
+
+
+def test_database_create_table(engine):
+    df = get_df_with_test_data(engine)
+
+    # We add a random value to the dataframe, that way we can check that it was actually this test run
+    # that wrote the table, and not some earlier run that happened to write the same data to the same table.
+    random_value = randrange(1_000_000)
+    df['x'] = random_value
+    expected_columns = [
+        '_index_skating_order', 'skating_order', 'city', 'municipality', 'inhabitants', 'founding', 'x'
+    ]
+    expected_data = [
+        [1, 1, 'Ljouwert', 'Leeuwarden', 93485, 1285, random_value],
+        [2, 2, 'Snits', 'Súdwest-Fryslân', 33520, 1456, random_value],
+        [3, 3, 'Drylts', 'Súdwest-Fryslân', 3055, 1268, random_value]
+    ]
+
+    df_from_table = df.database_create_table(
+        table_name='test_df_database_create__test_database_create_table',
+        overwrite=True
+    )
+    assert_equals_data(df_from_table, expected_columns=expected_columns, expected_data=expected_data)
+
+    # check that df_from_table actually queries the table
+    dialect = engine.dialect
+    assert df_from_table.base_node != df.base_node
+    new_sql = to_sql(dialect=dialect, model=df_from_table.base_node)
+    quoted_table_name = quote_identifier(dialect, 'test_df_database_create__test_database_create_table')
+    expected_sql_fragment = f'FROM {quoted_table_name}'
+    assert expected_sql_fragment in new_sql
+
+    # Test again, but with overwrite=False. We expect an error
+    random_value = randrange(1_000_000)
+    df['y'] = random_value
+    # we get a database error, so the exact error is database dependent. Both Postgres and BigQuery should
+    # raise an error containing the phrase 'already exists', but with different capitalization
+    with pytest.raises(Exception, match='[aA]lready [eE]xists'):
+        df_from_table = df.database_create_table(
+            table_name='test_df_database_create__test_database_create_table',
+            overwrite=False
+        )
+    # table should be unchanged after the last failed call
+    assert_equals_data(df_from_table, expected_columns=expected_columns, expected_data=expected_data)
+
+


bach/bach/dataframe.py
```diff
@@ -1114,6 +1114,48 @@ def get_unsampled(self) -> 'DataFrame':
...
 
+    def database_create_table(self, table_name: str, overwrite: bool = False, ) -> 'DataFrame':
+        """
+        Write the current state of the DataFrame to a database table.
+
+        The DataFrame that's returned will query from the written table for any further operations.
+
+        :param table_name: Name of the table to write to. Can include project_id and dataset
+                on BigQuery, e.g. 'project_id.dataset.table_name'
+        :param overwrite: if True, the sample data is written to table_name, even if that table already
```
hendrik-obj, line 1125:
```suggestion
        :param overwrite: if True, the data is written to table_name, even if that table already
```

bach/bach/dataframe.py
```diff
@@ -1114,6 +1114,48 @@ def get_unsampled(self) -> 'DataFrame':
         from bach.sample import get_unsampled
         return get_unsampled(df=self)
 
+    def database_create_table(self, table_name: str, overwrite: bool = False, ) -> 'DataFrame':
```
hendrik-obj, line 1117: why not: `materialize(materialization='table')`

hendrik-obj (commented):
functionally good, couple comments still

bach/bach/dataframe.py
```diff
@@ -1114,6 +1114,48 @@ def get_unsampled(self) -> 'DataFrame':
...
+        model = model.copy_set_materialization(Materialization.TABLE)
+        model = model.copy_set_materialization_name(materialization_name=table_name)
+
+        placeholder_values = get_variable_values_sql(dialect=dialect, variable_values=self.variables)
+        model = update_placeholders_in_graph(start_node=model, placeholder_values=placeholder_values)
+
+        sql = to_sql(dialect=dialect, model=model)
+        with self.engine.connect() as conn:
+            if overwrite:
+                sql = f'DROP TABLE IF EXISTS {quote_identifier(dialect, table_name)}; {sql}'
```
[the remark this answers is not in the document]
hendrik-obj: you can actually drop any table that you have the right to drop for. was already the case in the get_sample, but still something to consider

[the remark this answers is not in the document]
thijs-obj: yes this is a danger. I'll add a clear warning in the docstring. Additionally there is an issue on the backlog (https://github.com/objectiv/objectiv-analytics/issues/977) to make sure that in the demo setup the notebook user can't accidentally overwrite the events table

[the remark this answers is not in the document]
thijs-obj: > can you drop the table you are reading from? should we prevent that?

Interesting question! This would mess things up majorly, I think; I'll add a warning.
I don't think we can easily detect when this happens currently, we could detect that happens it if we implement https://github.com/objectiv/objectiv-analytics/issues/514 and add some code.

bach/bach/dataframe.py
thijs-obj: Of course I forgot an important scenario: we could also create the table the moment `materialize()` is called. But then that behaviour diverges from what we do with temporary tables. I don't think I'd like it if materialize has side-effects based on what parameters we give it :thinking:

bach/bach/dataframe.py
thijs-obj: Improved the docstring in 31ebf88e88735361b57629616ebe44fb9b797dce

bach/tests/functional/bach/test_df_database_create.py
```diff
@@ -0,0 +1,64 @@
...
+from random import randrange
+
+import pytest
+
+from tests.functional.bach.test_data_and_utils import get_df_with_test_data, assert_equals_data
+from sql_models.sql_generator import to_sql
+from sql_models.util import quote_identifier
+
+
+def test_database_create_table(engine):
```
jansentom, line 13: very nice test

bach/bach/dataframe.py
```diff
@@ -1114,6 +1114,52 @@ def get_unsampled(self) -> 'DataFrame':
...
         return get_unsampled(df=self)
 
+    def database_create_table(self, table_name: str, overwrite: bool = False, ) -> 'DataFrame':
+        """
+        Write the current state of the DataFrame to a database table.
+
+        The DataFrame that's returned will query from the written table for any further operations.
+
+        :param table_name: Name of the table to write to. Can include project_id and dataset
+                on BigQuery, e.g. 'project_id.dataset.table_name'
```
KathiaBarahona, line 1124: **question**:  Isn't it better to just support `bq_dataset` and  `bq_project_id` (same as we do in `DataFrame.from_table`) I mean I would expect a similar definition for `read` and `write` operations (in table-related param stuff)

bach/bach/dataframe.py
```diff
@@ -1114,6 +1114,52 @@ def get_unsampled(self) -> 'DataFrame':
...
+        sql = to_sql(dialect=dialect, model=model)
+        with self.engine.connect() as conn:
+            if overwrite:
+                sql = f'DROP TABLE IF EXISTS {quote_identifier(dialect, table_name)}; {sql}'
+
+            sql = escape_parameter_characters(conn, sql)
+            conn.execute(sql)
+
+        all_dtypes = {**self.index_dtypes, **self.dtypes}
+        return self.from_table(
```
KathiaBarahona, line 1156: IMHO, looks a bit odd calling a `classmethod` from `self`
```suggestion
        return DataFrame.from_table(
```

bach/bach/dataframe.py
```diff
@@ -1114,6 +1114,52 @@ def get_unsampled(self) -> 'DataFrame':
...
+            engine=self.engine,
+            table_name=table_name,
+            index=list(self.index.keys()),
```
KathiaBarahona, line 1159:
```suggestion
            index=self.index_columns,
```

bach/tests/functional/bach/test_df_database_create.py
KathiaBarahona, line 13: For avoiding conflicts with parallelized tests. (who knows if there's another test reading `test_df_database_create__test_database_create_table` :woman_shrugging: )
```suggestion
@pytest.mark.xdist_group(name="db_writers")
def test_database_create_table(engine):
```

bach/tests/functional/bach/test_df_database_create.py
```diff
@@ -0,0 +1,64 @@
...
+    df = get_df_with_test_data(engine)
+
+    # We add a random value to the dataframe, that way we can check that it was actually this test run
+    # that wrote the table, and not some earlier run that happened to write the same data to the same table.
```
KathiaBarahona, line 17: mmm wouldn't it be better to just drop the table at the end of pytest session? I mean, we should make sure we don't leave random test tables in the db.

bach/bach/dataframe.py
```diff
@@ -1114,6 +1114,52 @@ def get_unsampled(self) -> 'DataFrame':
...
+        Write the current state of the DataFrame to a database table.
+
+        The DataFrame that's returned will query from the written table for any further operations.
+
+        :param table_name: Name of the table to write to. Can include project_id and dataset
+                on BigQuery, e.g. 'project_id.dataset.table_name'
+        :param overwrite: if True, the data is written to `table_name`, even if that table already
+            exists. Beware, this will drop an existing table if `table_name` already exists. Make sure that
+            `table_name` does not contain any valuable information. Additionally, make sure that it is not a
+            source table of this DataFrame.
```
KathiaBarahona, line 1128: Let's state that if the table exists and `overwrite is False` then the user will receive an exception from the db

bach/bach/dataframe.py
```diff
@@ -1114,6 +1114,52 @@ def get_unsampled(self) -> 'DataFrame':
...
+        :return: New DataFrame; the base_node consists of a query on the newly created table.
+
+        .. warning::
+            With `overwrite=True`, if a table already exist with the given name, then that table will
+            be dropped and all data lost!
+        .. note::
+            This function queries the database.
```
KathiaBarahona, line 1135: Let's add it under https://github.com/objectiv/objectiv-analytics/blob/main/bach/docs/source/bach/core-concepts.rst#delayed-database-operations

KathiaBarahona: `get_sample` is also missing

jansentom (approved):
I'm not ready to leave savepoints behind, but do agree that we need a simpler interface for just this. Nice PR. I've added a bit to existing discussions.

KathiaBarahona (changes requested):
Cool thing to have, tho makes me uncomfortable as it feels **VERY** dangerous :skull: . For me we should ditch the `overwrite` param, IMHO bach shouldn't be dropping tables, we are heavily relying that users actually read docs. I prefer raising an exception from the db about an existing table, than having a functionality that has a high chance of putting db integrity on risk

bach/bach/dataframe.py
thijs-obj: I think it is actually more correct to use `self.from_table(` in cases like this. If for some reason somebody decides to subclass DataFrame and override `from_table`, then with `self.` we'll call the function in the subclass. Which seems like more logical behaviour to me

bach/tests/functional/bach/test_df_database_create.py
thijs-obj: We could, but it would slow the test a bit, and on BigQuery there is a quota on the number of table operations we can do. As far as I could find online [1] that quota is against the table name, so deleting it doesn't reset it. So just leaving it seemed simpler :man_shrugging: 


[1] https://stackoverflow.com/questions/54282175/does-deleting-and-creating-a-table-in-bigquery-renews-the-quota-limit-per-day#comment95387301_54282503

bach/bach/dataframe.py
thijs-obj: I agree that this is inconsistent. But I think this is actually the better interface than what we have in `DataFrame.from_table`. Do you agree? Then I'll create an issue to fix that

bach/bach/dataframe.py
thijs-obj: Fixed in d426795ff3bac2369d97d778913963dd0e26fcec

bach/bach/dataframe.py
thijs-obj: Updated a lot of docstring in d426795ff3bac2369d97d778913963dd0e26fcec

bach/tests/functional/bach/test_df_database_create.py
thijs-obj: Did change the table name slightly in d426795ff3bac2369d97d778913963dd0e26fcec, to be consistent with how we do this in test_df_sample.py. Removed a redundant `test`. So now it is `test_df_database_create__database_create_table`, before it was `test_df_database_create__test_database_create_table`

thijs-obj:
Applied all comments from the out-of-band discusison in d426795ff3bac2369d97d778913963dd0e26fcec

bach/bach/dataframe.py
KathiaBarahona: yup, then let's remove those params from `DataFrame.from_table`

bach/tests/functional/bach/test_df_database_create.py
KathiaBarahona: then I think we should have a constant for for the x_dist group, like `XDIST_GROUP_FOR_DB_WRITE_TESTS = 'db_writer_group'` and use that constant for all  current xdist marked tests. I consider that a fast approach than making sure that all tests have unique table names

bach/tests/functional/bach/test_df_database_create.py
KathiaBarahona: mmmhhh alright

KathiaBarahona (approved):
just one comment, but lgtm

bach/bach/dataframe.py
thijs-obj: https://github.com/objectiv/objectiv-analytics/issues/992

bach/tests/functional/bach/test_df_database_create.py
thijs-obj: We discussed this over video-chat. And I think the conclusion was that as it is this works. Not sure if we agreed that this was actually a better system than using `xdist_group`, but I do believe we both agreed that it works as is.

Sort of related discussion was this https://github.com/objectiv/objectiv-analytics/pull/976/files#r914069460

bach/bach/dataframe.py
thijs-obj: I think in some places we do follow that logic, but not here. For example in Series we almost always call classmethods on the specific class or on self, never `Series.<classmethod>()`. But with DataFrame we most definitely don't do that consistently, and it might actually quite a bit of refactoring to do that consistently.

bach/bach/dataframe.py
thijs-obj: PR: https://github.com/objectiv/objectiv-analytics/pull/998

PR #983, opened 2022-07-05
Sqlmodel `to_sql()` performance improvement

thijs-obj:
Change: Fixed performance issue in `find_nodes()`:
* Keep track of nodes that have already been discovered
* Don't visit nodes twice, unless the new path is longer than the previously discovered path.

This is a follow-up to https://github.com/objectiv/objectiv-analytics/pull/972/files#r913747845

### DIFF
--- a/bach/sql_models/graph_operations.py
+++ b/bach/sql_models/graph_operations.py
@@ -137,7 +137,14 @@ def find_nodes(
     """
     # result_nodes maps the id of the found objects to a FoundNode object
     result_nodes: Dict[int, FoundNode] = {}
+    # discovered maps the id of all models that are processed or are queued to be processed to the length of
+    # the reference path to get to them. Depending on the value of first_instance, this is the length of the
+    # longest or shortest reference path.
+    discovered: Dict[int, int] = {}
+    # queue contains the queue of items that have been discovered but not yet processed.
     queue: Deque[Tuple[SqlModel, RefPath]] = deque()
+
+    discovered[id(start_node)] = 0
     queue.append((start_node, tuple()))
     while queue:
         node, path = queue.popleft()
@@ -146,12 +153,24 @@ def find_nodes(
             if current_id not in result_nodes:
                 result_nodes[current_id] = FoundNode(node, path)
             elif current_id in result_nodes and not first_instance:
+                # We found a longer path to a node we already found earlier.
                 # we rely on the fact that python 3.7+ will keep the insertion order. So we'll have to
                 # remove and reinsert the item to get it in the right position in the returned result.
                 del result_nodes[current_id]
                 result_nodes[current_id] = FoundNode(node, path)
-        for next_path, next_node in node.references.items():
-            next_tuple = (next_node, path + (next_path,))
+        for next_path_step, next_node in node.references.items():
+            next_id = id(next_node)
+            next_path = path + (next_path_step,)
+            len_next_path = len(next_path)
+            if next_id in discovered:
+                # If there is already a path to next_node that's equally long or short (depending on
+                # first_instance), then we don't add next_node to the queue.
+                if first_instance and discovered[next_id] <= len_next_path:
+                    continue
+                if (not first_instance) and discovered[next_id] >= len_next_path:
+                    continue
+            discovered[next_id] = len_next_path
+            next_tuple = (next_node, next_path)
             queue.append(next_tuple)
     return list(result_nodes.values())
 
--- a/bach/tests/unit/sql_models/test_graph_operations.py
+++ b/bach/tests/unit/sql_models/test_graph_operations.py
@@ -234,6 +234,9 @@ def test_find_nodes_path_length():
     assert result == [FoundNode(model=vm2, reference_path=('ref_left', 'ref_left', 'ref_right', 'ref_left'))]
 
 
+# be explicit as this is also a performance test: max 1 sec runtime.
+# But conftests.py should already define limit for all unittests.
+@pytest.mark.timeout(1)
 def test_find_nodes_long_and_short_path_performance():
     # We'll build a graph with a simple structure but with a lot of paths. The number of joins is controlled
     # through the `depth` constant. Schematically the graph looks like this (all joins between 2 and N-1 are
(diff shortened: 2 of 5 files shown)


bach/tests/unit/sql_models/test_graph_operations.py
```diff
@@ -219,19 +219,55 @@ def test_find_nodes_path_length():
...
+    #   vm <------+                                                                                /
+    #              \              /----\             /- ... -\                /----\              /
+    #               +-- join1 <--+     +-- join2 <--+         +-- joinN-1 <--+      +-- joinN <--/
+    #                             \----/             \- ... -/                \----/
+    #
+    # There are a lot of possible reference paths from the `graph` node to `vm` node: 2^depth + 1
+    # For depth values above ~ 18 this used to get noticeably slow (> 1 second to run), for values larger
+    # than 23 it was unbearable slow (> 1 minute to run) .
+    #
+    # If this test takes longer than a second to run, then that's a regression!
```
jansentom, line 253: what about `@pytest.mark.timeout(1)` (or 3, since GHA also has to run it ;)

thijs-obj: That's a nice idea. Added pytest-timeout in 2af151a8745ad8920172dc1f61a756a83fca2ea5

bach/sql_models/graph_operations.py
```diff
@@ -146,12 +153,24 @@ def find_nodes(
...
-        for next_path, next_node in node.references.items():
-            next_tuple = (next_node, path + (next_path,))
+        for next_path_step, next_node in node.references.items():
+            next_id = id(next_node)
+            next_path = path + (next_path_step,)
+            len_next_path = len(next_path)
+            if next_id in discovered:
+                # If there is already a path to next_node that's equally long or short (depending on
+                # first_instance), then we don't add next_node to the queue.
+                if first_instance and discovered[next_id] <= len_next_path:
```
KathiaBarahona, line 168: **question** What is the purpose of finding the longest path?

KathiaBarahona: I think my problem with this is that it might be non-deterministic

bach/tests/unit/sql_models/test_graph_operations.py
```diff
@@ -219,19 +219,55 @@ def test_find_nodes_path_length():
     result = find_nodes(graph, function=lambda n: n is vm1, first_instance=True)
     assert result == [FoundNode(model=vm1, reference_path=('ref_right', 'ref'))]
     result = find_nodes(graph, function=lambda n: n is vm1, first_instance=False)
-    assert result == [FoundNode(model=vm1, reference_path=('ref_left', 'ref_right', 'ref_right', 'ref'))]
+    assert result == [FoundNode(model=vm1, reference_path=('ref_left', 'ref_left', 'ref_right', 'ref_right'))]
```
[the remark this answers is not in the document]
KathiaBarahona: **question** I think we are missing testing if 2 nodes are unreachable. For example `vm1` to `vm2`.

KathiaBarahona (approved):
lgtm. Tested versus networkx graph algorthims. 
https://colab.research.google.com/drive/1RwYcIV901jCe4t6X5SWQ0kaf_qai3sMm?usp=sharing

[all_simple_paths](https://networkx.org/documentation/latest/reference/algorithms/generated/networkx.algorithms.simple_paths.all_simple_paths.html#all-simple-paths) (uses depth-first search)
[shortest_path](https://networkx.org/documentation/latest/reference/algorithms/generated/networkx.algorithms.shortest_paths.generic.shortest_path.html) (uses dijskstra)

just a small question

bach/sql_models/graph_operations.py
thijs-obj: We use this in `to_sql()`, See the comment in that function: https://github.com/objectiv/objectiv-analytics/blob/b2829520ccf552f97358351150f3274770dc56cb/bach/sql_models/sql_generator.py#L42-L46

In `to_sql()` we use this function to find all nodes that we need to generate SQL for, e.g. to create temporary tables. By giving the longest path, we can easily sort the found nodes by path-length. The nodes with longer paths could be dependencies of the nodes with shorter paths.

For example, a query might depend on a temporary table, by sorting the nodes by path length before we start generating SQL, we know for sure that we'll generate the SQL for that temporary table before we generate the SQL for the query that needs the temporary table.

bach/tests/unit/sql_models/test_graph_operations.py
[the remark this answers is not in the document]
thijs-obj: We have a test for `find_nodes()` that checks that it correctly gives back nothing if we try to find a node that doesn't exist:
https://github.com/objectiv/objectiv-analytics/blob/b2829520ccf552f97358351150f3274770dc56cb/bach/tests/unit/sql_models/test_graph_operations.py#L147-L149

If `vm1` to `vm2` is not reachable, then that means that `vm2` is not part of the graph that starts at `vm1`, so that's sort of the same thing as finding a node that doesn't exist.

The graph is a directed a-cyclic graph. Each node together with all the nodes that can be reached from there, forms its own graph. If `vm1` is not reachable from `vm2`, then it simply is not in the graph that's `vm2`. 

It could be that they are both part of a larger graph (like here https://github.com/objectiv/objectiv-analytics/blob/b2829520ccf552f97358351150f3274770dc56cb/bach/tests/unit/sql_models/test_graph_operations.py#L196-L208). But `find_nodes()` can only be used to find nodes in a graph that starts at `start_node`. With that graph one could do `find_nodes(start_node=vm1, function=lambda node: node is vm2)` but that would simply return an empty list as `vm2` is not in the graph of `vm1`. There are no parameters to say consider the whole graph starting at `graph`, but start searching at `vm1`. So the test that checks that we correctly give back nothing if a node doesn't exist covers this I believe

thijs-obj:
> lgtm. Tested versus networkx graph algorthims. https://colab.research.google.com/drive/1RwYcIV901jCe4t6X5SWQ0kaf_qai3sMm?usp=sharing

Cool that you did such extensive testing! thanks

bach/tests/unit/sql_models/test_graph_operations.py
```diff
@@ -252,7 +255,8 @@ def test_find_nodes_long_and_short_path_performance():
     #
     # If this test takes longer than a second to run, then that's a regression!
     #
-    depth = 25
+    import time; time.sleep(0.1)
```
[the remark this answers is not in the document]
thijs-obj: removed in b2829520ccf552f97358351150f3274770dc56cb

PR #988, opened 2022-07-06
handle exceptions while connecting to pg

borft:


backend/objectiv_backend/end_points/collector.py
```diff
@@ -295,13 +296,16 @@ def write_sync_events(ok_events: EventDataList, nok_events: EventDataList, event
...
-        finally:
-            connection.close()
+            connection = get_db_connection(output_config.postgres)
+            try:
+                with connection:
+                    insert_events_into_data(connection, events=ok_events)
+                    insert_events_into_nok_data(connection, events=nok_events)
+            finally:
+                connection.close()
+        except psycopg2.OperationalError as oe:
```
jansentom, line 307: Doesn't this ignore "other" transient errors from PG, thus killing the sequence when something other than a (re)connection error occurs?

borft: It catches this:
```Exception raised for errors that are related to the database’s operation and not necessarily under the control of the programmer, e.g. an unexpected disconnect occurs, the data source name is not found, a transaction could not be processed, a memory allocation error occurred during processing, etc. It is a subclass of [DatabaseError](https://www.psycopg.org/docs/module.html#psycopg2.DatabaseError).```

We could catch the more generic DatabaseError, to handle other cases such as errors in the queries (https://www.psycopg.org/docs/module.html#dbapi-exceptions)

jansentom: Since we want max resiliency, that sounds logical. Please pick what you think works best for that.

PR #989, opened 2022-07-07
Fix features and target values order for Logistic Regression in BQ

vard-obj:


modelhub/modelhub/models/logistic_regression.py
```diff
@@ -66,8 +67,8 @@ def fit(self, X: 'DataFrame', y: 'SeriesBoolean'):
         if not isinstance(y, SeriesBoolean):
             raise TypeError(f"y is of type {type(y)}, should be SeriesBoolean")
 
-        X_p = X.to_pandas()
-        y_p = y.to_pandas()
+        X_p = X.to_pandas().sort_index()
+        y_p = y.to_pandas().sort_index()
```
hendrik-obj, line 71:
```suggestion
        data = X.copy()
        data[y.name] = y
        pdf = data.to_pandas()
        X_p = pdf[X.data_columns]
        y_p = pdf[y.name]
```
reasoning: This way you automatically use bach to check if the indexes are the same. Sorting then is not needed anymore, and only one query to export the data.

hendrik-obj: also the docstring should be changed then

modelhub/tests_modelhub/functional/modelhub/logistic_regression_test_utils.py
```diff
@@ -30,8 +30,8 @@ def __init__(self, X: bach.DataFrame, y: bach.Series, **kwargs):
         self.X = X
         self.y = y
 
-        self.X_p = self.X.to_pandas()
-        self.y_p = self.y.to_pandas()
+        self.X_p = self.X.to_pandas().sort_index()
+        self.y_p = self.y.to_pandas().sort_index()
```
hendrik-obj, line 34: if you agree with suggestion to `fit`. then this could also be adjusted:
```suggestion
        data = X.copy()
        data[y.name] = y
        pdf = data.to_pandas()

        self.X_p = pdf[X.data_columns]
        self.y_p = pdf[y.name]
```

hendrik-obj (commented):
Good that this is fixed. Do have a suggestion that might be a bit cheaper and more robust? See comments on test.

hendrik-obj (approved):
lgtm

PR #994, opened 2022-07-07
Bach: Reference correct columns in `ConcatSqlModel`

KathiaBarahona:
Seems SeriesConcatOperation was making reference to a column that doesn't exist in the base node. Also there was some duplicated code, so now we use `_get_model` from the parent class for both `DataFrame` and `Series`


thijs-obj (approved):
lgtm

PR #995, opened 2022-07-07
Bach: sort Series based on other existent columns from same base node

KathiaBarahona:
Enables `keys` param in `Series.sort_values`. This way we can sort by expressions from other series that make reference to columns that actually exist in the base node.


bach/bach/series/series.py
```diff
@@ -315,6 +320,14 @@ def index_sorting(self) -> List[bool]:
         """
         return copy(self._index_sorting)
 
+    @property
+    def sorting_keys(self) -> List[SortColumn]:
```
KathiaBarahona, line 324: @thijs-obj let me know what do you think about the naming, might not be quite clear..

bach/bach/series/series.py
```diff
@@ -838,13 +854,25 @@ def reset_index(
 
         return result
 
-    def sort_values(self, *, ascending=True):
+    def sort_values(self, *, ascending=True, keys: Optional[List['Series']] = None):
```
KathiaBarahona, line 857: @thijs-obj also here, maybe `by_series_expressions`?

bach/bach/series/series.py
thijs-obj: I think I'd prefer `order_by`, as that is what we have on DataFrame, and this is the same kind of data. But don't feel strongly about that, `sorting_keys` is okay too I think

bach/bach/series/series.py
```diff
@@ -512,6 +525,7 @@ def copy_override(
         sorted_ascending: Optional[Union[bool, NotSet]] = not_set,
         index_sorting: Optional[List[bool]] = None,
         instance_dtype: Optional[StructuredDtype] = None,
+        sorting_keys: Optional[Union[List[SortColumn], NotSet]] = not_set,
```
thijs-obj, line 528: I am not a super big fan of using `not_set` here. As it's not really needed.

For some values we don't have a choice because we cannot distinguish "set this to no" from "don't change this value"(if we don't want to change the datastructure), e.g. the only way to denote that we don't have a groupby, is by setting the value to `None`, so we cannot use `None` here to denote "don't change anything", so we must use `not_set` for that. But for `sorting_keys`, we could just say that the way to set "no sorting" is to give an empty list, and then we can use `None` here as the value for "don't change anything"

bach/bach/series/series.py
```diff
@@ -205,6 +209,7 @@ def __init__(
         self._sorted_ascending = sorted_ascending
```
thijs-obj, line 209: I think we should get rid of both `sorted_ascending` and `index_sorting`, as we can store that information in _sorting_keys

thijs-obj (commented):
Didn't have time for a very detailed review. So probably best if some other people look at this too.

I really like the idea of storing sorting information in the same way in Series as we do in DataFrame.

bach/bach/series/series.py
KathiaBarahona: yup, but I do think we should do a separate PR for that. Else this PR will get quite big and this is more a hotfix

thijs-obj: Reason I suggested it, is because I think it actually makes things simpler. But if you think it's better to do it in a follow-up, then that's fine for me :+1: Is there an issue?

We do need to make sure we handle the edge cases of these 3 different sorting data fields interacting correctly.
Some examples that are currently weird (imho):
1. 
```python 
series = series.sort_index()
series = series.sort_values(keys=[df.some_series])
```
I'd expect the `sort_values()` call to change the sorting to only use `some_series`, but if I am reading the code correctly in this case we sort on the index, and after that we sort on `some_series`

Doing the calls other way around is not quite as weird, but also gives a different result to what I'd expect

2.
```python 
series = series.sort_values(ascending=True)
series = series.sort_values(keys=[df.some_series])
```
Again, I'd expect the series to be sorted by `some_series`, but instead it's sorted by the values of the series, ascending.

bach/bach/series/series.py
```diff
@@ -838,13 +854,25 @@ def reset_index(
...
         Returns a new instance and does not actually modify the instance it is called on.
 
         :param ascending: Whether to sort ascending (True) or descending (False)
+        :param keys: Apply sorting based on expressions from other series that share same
+        base node as caller
         """
+        if keys and any(key.base_node != self.base_node for key in keys):
+            raise ValueError('All series provided as keys must have the same base_node as caller.')
+
+        if keys:
```
thijs-obj, line 869:
```suggestion
        if keys is not None:
```
This would allow one to set the sorting to 'no sorting' by passing `[]`. Otherwise there is no way to remove sorting

bach/bach/utils.py
```diff
@@ -80,3 +86,21 @@ def is_valid_column_name(dialect: Dialect, name: str) -> bool:
         prefix_ok = not any(name.startswith(prefix) for prefix in reserved_prefixes)
         return len_ok and pattern_ok and prefix_ok
     raise DatabaseNotSupportedException(dialect)
+
+
+def validate_sorting_expressions(node: BachSqlModel, order_by: List[SortColumn]) -> None:
```
thijs-obj, line 91: some docstring explaining what we validate would be niec

bach/bach/series/series.py
```diff
@@ -838,13 +854,25 @@ def reset_index(
...
 
-    def sort_values(self, *, ascending=True):
+    def sort_values(self, *, ascending=True, keys: Optional[List['Series']] = None):
         """
         Sort this Series by its values.
         Returns a new instance and does not actually modify the instance it is called on.
 
         :param ascending: Whether to sort ascending (True) or descending (False)
+        :param keys: Apply sorting based on expressions from other series that share same
+        base node as caller
```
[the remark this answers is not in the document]
thijs-obj: Might actually make sense to have a completely different function for sorting by other series? Would make it a lot easier to write non-confusing docstring

bach/bach/series/series.py
```diff
@@ -125,6 +125,7 @@ def __init__(
         sorted_ascending: Optional[bool],
         index_sorting: List[bool],
         instance_dtype: StructuredDtype,
+        sorting_keys: Optional[List[SortColumn]] = None,
```
thijs-obj, line 128: not in docstring

thijs-obj (commented):
Did a thorough review this time. left some comments

bach/bach/series/series.py
KathiaBarahona: mmm true, will change this

bach/bach/series/series.py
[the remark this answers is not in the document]
thijs-obj: `sort_by_series` sounds good to me

bach/bach/series/series.py
KathiaBarahona: Created this issue: https://github.com/objectiv/objectiv-analytics/issues/1021
Now there's a separate sorting function for doing it with Series: https://github.com/objectiv/objectiv-analytics/pull/995/commits/7b41884be66ed08c546c7cde09e2379398918f37

bach/bach/series/series.py
KathiaBarahona: Removed this param and created a different function for this functionality https://github.com/objectiv/objectiv-analytics/pull/995/commits/7b41884be66ed08c546c7cde09e2379398918f37

bach/bach/series/series.py
KathiaBarahona: Changed in https://github.com/objectiv/objectiv-analytics/pull/995/commits/7b41884be66ed08c546c7cde09e2379398918f37

bach/bach/series/series.py
KathiaBarahona: Changed https://github.com/objectiv/objectiv-analytics/pull/995/commits/7b41884be66ed08c546c7cde09e2379398918f37

bach/bach/series/series.py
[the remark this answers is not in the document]
KathiaBarahona: Done here https://github.com/objectiv/objectiv-analytics/pull/995/commits/7b41884be66ed08c546c7cde09e2379398918f37

bach/bach/series/series.py
KathiaBarahona: added as `order_by` https://github.com/objectiv/objectiv-analytics/pull/995/commits/7b41884be66ed08c546c7cde09e2379398918f37

bach/bach/series/series.py
```diff
@@ -858,6 +877,28 @@ def sort_values(self, *, ascending=True):
             return self
         return self.copy_override(sorted_ascending=ascending)
 
+    def sort_by_series(self, by: List['Series'], ascending: Union[bool, List[bool]] = True):
```
thijs-obj, line 880:
```suggestion
    def sort_by_series(self: T, by: List['Series'], ascending: Union[bool, List[bool]] = True) -> T:
```
Just a bit nicer for calling code

bach/bach/series/series.py
```diff
@@ -858,6 +877,28 @@ def sort_values(self, *, ascending=True):
...
+        """
+        Sort this Series by other Series that have the same base node as the caller.
```
thijs-obj, line 882:
```suggestion
        Sort this Series by other Series, that have the same base node as this Series.
```

"the caller" I would understand as the function calling this function.

bach/bach/series/series.py
thijs-obj: These edge cases are fixed with https://github.com/objectiv/objectiv-analytics/commit/7b41884be66ed08c546c7cde09e2379398918f37, or are they not important and will they be fixed by #1021?

bach/bach/series/series.py
thijs-obj, line 880: Nice to have a separate function here

bach/bach/series/series.py
KathiaBarahona: Will be fixed in #1021 as this PR only appends to the current sorting

bach/bach/series/series.py
KathiaBarahona: I think we should do this stuff in a separate PR, there are a lot of functions in Series class that should be correctly typed. TBH I just followed the current pattern

KathiaBarahona: https://github.com/objectiv/objectiv-analytics/issues/1032

bach/bach/series/series.py
```diff
@@ -316,6 +324,14 @@ def index_sorting(self) -> List[bool]:
         """
         return copy(self._index_sorting)
 
+    @property
+    def order_by(self) -> List[SortColumn]:
+        """
+        Get the series expressions for sorting this Series. Columns referenced on sorting expressions
```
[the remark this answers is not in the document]
KathiaBarahona: Changed here https://github.com/objectiv/objectiv-analytics/pull/995/commits/4fc35dd3db64241f29dc2e7a72eb5183baa5f863

bach/bach/series/series.py
KathiaBarahona: Done here https://github.com/objectiv/objectiv-analytics/pull/995/commits/4fc35dd3db64241f29dc2e7a72eb5183baa5f863

bach/bach/series/series.py
thijs-obj: Ok fine with me if we actually make sure we do #1021 then. Still think this is a weird order to do things.

bach/bach/series/series.py
thijs-obj: another thing: maybe for consistency we should put an `*, ` before `ascending`. We have that for sort_index and sort_values

bach/bach/series/series.py
KathiaBarahona: The main reason is because of the amount of changes #1021 introduces :woman_shrugging: . The purpose of this PR was more of a hotfix.

bach/bach/series/series.py
KathiaBarahona: > another thing: maybe for consistency we should put an `*, ` before `ascending`. We have that for sort_index and sort_values

Then we should follow this for `DataFrame` as well then. IMHO we should create a guide with this type of practices when contributing to Bach, as this type of nitpicking can be avoided in a future

KathiaBarahona: typing for the method here https://github.com/objectiv/objectiv-analytics/pull/995/commits/b76152d17cad2ab0cebe2cfcd3ef2663eec66001

KathiaBarahona: PR for rest of typings: https://github.com/objectiv/objectiv-analytics/pull/1038

thijs-obj: > Then we should follow this for DataFrame as well then. IMHO we should create a guide with this type of practices when contributing to Bach, as this type of nitpicking can be avoided for new contributors.

Totally okay with me not to do this change. It's fully arbitrary. I was looking at the other function definitions, and I saw we did it like that there, and figured I point it out. Didn't mean to imply this is a change that would be a blocker for me or any of the sort

Issue #992, opened 2022-07-07
BQ: make DataFrame.from_table work with full BQ table name


thijs-obj:
DataFrame.from_table() currently has `bq_dataset` and `bq_project_id` parameters to get data from tables that are in another BQ project. More consistent would be to support table_names of the form `bq_project_id.bq_dataset.table_name`. Like we do in `DataFrame.database.create_table()`. See discussion https://github.com/objectiv/objectiv-analytics/pull/976/files#r914061937

PR #998, opened 2022-07-07
Bach BQ: support full table names in `from_table()`

thijs-obj:
This solves issue https://github.com/objectiv/objectiv-analytics/issues/992

Change:
* Support full table names for BigQuery in `from_table()`, that is `project_id.dataset.table_name` is now supproted
* Remove `bq_dataset` and `bq_project_id` parameters. I don't think they were every used anywhere. And I highly doubt any users depend on it. So seems safe to remove

Reason for this change is more consistency, between what we allow in `database_create_table()`, `get_sample()`, and now `from_table()`. Resolves this discussion in a recent PR: https://github.com/objectiv/objectiv-analytics/pull/976/files#r914061937


bach/bach/from_database.py
```diff
@@ -71,6 +65,25 @@ def get_dtypes_from_table(
...
+    """ Split a BigQuery table name into project, dataset and actual table_name. """
+    parts = table_name.split('.')
+    if len(parts) == 3:
+        return FullTableName(project_id=parts[0], dataset=parts[1], table_name=parts[2])
+    if len(parts) == 2:
+        return FullTableName(project_id=None, dataset=parts[0], table_name=parts[1])
+    if len(parts) == 1:
+        return FullTableName(project_id=None, dataset=None, table_name=parts[0])
+    raise ValueError(f'Expected table_name in format "{{project_id}}.{{dataset}}.{{table_name}}"'
+                     f', value: {table_name}')
```
KathiaBarahona, line 84: mmm  why not just having a regex? smth like `r'^([^.]+\.){0,2}(?P<table_name>[^.]+)$'`. Seems to me we just need to check if the string is in the correct format and just replace the string from the `table_name` group with `INFORMATION_SCHEMA.COLUMNS`.

bach/bach/dataframe.py
```diff
@@ -597,7 +575,7 @@ def from_table(
         }
         sql_params.update(column_placeholders)
 
-        sql = f'SELECT {column_stmt} FROM {sql_table_name_template}'
+        sql = f'SELECT {column_stmt} FROM {{table_name}}'
```
KathiaBarahona, line 578: **question**: shouldn't we check the `table_name` is in the right format for BQ? Else user will only know they provided the wrong table_name till they try to execute a query (if they actually provided the dtypes)

bach/bach/from_database.py
```diff
@@ -71,6 +65,25 @@ def get_dtypes_from_table(
     return _get_dtypes_from_information_schema_query(engine=engine, query=sql)
 
 
+class FullTableName(NamedTuple):
+    project_id: Optional[str]
+    dataset: Optional[str]
+    table_name: str
+
+
+def _split_bq_table_name(table_name: str) -> FullTableName:
```
KathiaBarahona, line 74: Can we add a test for this?

KathiaBarahona (commented):
Some comments

bach/bach/dataframe.py
thijs-obj: In general, if a user gives a non-existing table name, then we won't know that that table doesn't exist until we try to query it.

We could check the specific sub-case where the table is not only non-existent but also the name would never be allowed as a table name. But I am doubting a bit how much extra worth that would have to the user.

bach/bach/from_database.py
thijs-obj: Well, I thought having `split('.')` was simpler than having a regex, and the result from split is somewhat similar as what it was before: three separate values, two of which could be None. This made the code change in `get_dtypes_from_table()` sort of smaller.

We don't just need to replace `table_name` in the regex with `INFORMATION_SCHEMA.COLUMNS`, we also need the `table_name` value itself

bach/bach/from_database.py
```diff
@@ -35,31 +35,25 @@ def get_dtypes_from_model(engine: Engine, node: SqlModel) -> Dict[str, Structure
...
-    bq_dataset: Optional[str] = None,
-    bq_project_id: Optional[str] = None
 ) -> Dict[str, StructuredDtype]:
     """
     Query database to get dtypes of the given table.
     :param engine: sqlalchemy engine for the database.
-    :param table_name: the table name for which to get the dtypes
-    :param bq_dataset: BigQuery-only. Dataset in which the table resides, if different from engine.url
-    :param bq_project_id: BigQuery-only. Project of dataset, if different from engine.url
+    :param table_name: the table name for which to get the dtypes.
```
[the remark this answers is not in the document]
thijs-obj: doh. Fixed in https://github.com/objectiv/objectiv-analytics/pull/998/commits/93fd8777e906fa96d474f261bc685a381f6235ba

bach/bach/from_database.py
thijs-obj: I further simplified and let go of splitting this in three parts. Split in two now. And refactored it all a bit in https://github.com/objectiv/objectiv-analytics/pull/998/commits/93fd8777e906fa96d474f261bc685a381f6235ba

let me know what you think

PR #1001, opened 2022-07-08
Deterministic sorting for example notebooks

hendrik-obj:
add deterministic sorting
todo
- [x] adjust example pages same way as notebooks


jansentom (approved):
LGTM

PR #1003, opened 2022-07-11
Add BigQuery usage guide

vard-obj:


bach/docs/source/bach/core-concepts.rst
```diff
@@ -50,4 +50,30 @@ situations:
...
   and mainly uses NaN. When outputting data from a Bach DataFrame to a pandas DataFrame, most of this
   distinction is lost again.
-* In a Bach DataFrame column names must be unique, in pandas this is not the case
\ No newline at end of file
+* In a Bach DataFrame column names must be unique, in pandas this is not the case
+
+
+BigQuery Tips
+-------------
+The columns in Bach DataFrame are the database columns, hence the column name must contain only letters (`a-z`, `A-Z`), numbers (`0-9`), or underscores (`_`), and it must start with a letter or underscore. Especially remember this during [`DataFrame.unstack()`](/modeling/bach/api-reference/DataFrame/bach.DataFrame.unstack.mdx) usage.
```
thijs-obj, line 58: I think we need to connect these restrictions on column name to the restrictions that BigQuery places on them. Because a user might not be aware that BQ restricts column names.

E.g. something like
```suggestion
The series in a Bach DataFrame map directly to database columns, hence a series names must be a valid BigQuery column name,
and contain only letters (`a-z`, `A-Z`), numbers (`0-9`), or underscores (`_`), and it must start with a letter or underscore. Especially remember this during :py:meth:`DataFrame.unstack()`usage.
```

Also changed the link to unstack, I think this should be automatically expanded correctly

bach/docs/source/bach/core-concepts.rst
```diff
@@ -50,4 +50,30 @@ situations:
...
+
+For getting a sample of the data one can use:
```
thijs-obj, line 60: I think we need to explain why this is a BigQuery specific tip.

For example something like "Querying big datasets can be expensive with BigQuery. To get a smaller sample of the current DataFrame use:"

bach/docs/source/bach/core-concepts.rst
```diff
@@ -50,4 +50,30 @@ situations:
...
+
+.. code-block:: console
+
+    table_name = 'objectiv-production.writable_dataset.table_name'
+    df.get_sample(table_name, sample_percentage=10, overwrite=True)
+
+It creates a permanent table, and if the table exists it overwrites it (make sure you have a write access). For BigQuery `seed` parameter is not implemented.
+
+If you are planning to do a lot of operations on a given Bach DataFrame and already the underlying SQL query is complex it would be more optimal for the database first to materialize the current DataFrame as a temporary table:
```
thijs-obj, line 69: I think we need to explain that BigQuery cannot always plan very complex queries. And mention here what error the user will get in that case. Then the user can recognize that error if he runs into that, and apply this tip.

bach/docs/source/bach/core-concepts.rst
```diff
@@ -50,4 +50,30 @@ situations:
 * Bach DataFrames can distinguish between `NULL`/`None` and Not-a-Number (`NaN`). Pandas generally doesn't
   and mainly uses NaN. When outputting data from a Bach DataFrame to a pandas DataFrame, most of this
   distinction is lost again.
-* In a Bach DataFrame column names must be unique, in pandas this is not the case
\ No newline at end of file
+* In a Bach DataFrame column names must be unique, in pandas this is not the case
+
+
+BigQuery Tips
```
thijs-obj, line 56: We have three tips here now:
1) Use valid column names
2) Getting a data sample from a DataFrame
3) Using temporary tables to limit query complexity.

Can we make every tip it's own sub-section? Or change the formatting a bit so it's clear where one tips ends and the next starts?

bach/bach/dataframe.py
```diff
@@ -1145,8 +1145,8 @@ def database_create_table(self, table_name: str, *, if_exists: str = 'fail') ->
         :param if_exists: {'fail', 'replace'}. How to behave if the table already exists:
             * fail: Raise an Exception.
             * replace: Drop the table before inserting new values. All data in that table will be lost! Make
-                sure that `table_name` does not contain any valuable information. Additionally, make sure
```
thijs-obj, line 1148: This formatting mistake was my bad :speak_no_evil:. Is also fixed in #1004 by Ivar

thijs-obj (commented):
Left a few comments. Let me know if you have any questions

bach/docs/source/bach/core-concepts.rst
```diff
@@ -50,4 +50,38 @@ situations:
...
+1. Use valid column names
+    The series in a Bach DataFrame map directly to database columns, hence a series names must be a valid BigQuery column name, and contain only letters (`a-z`, `A-Z`), numbers (`0-9`), or underscores (`_`), and it must start with a letter or underscore. Especially remember this during :py:meth:`DataFrame.unstack()` usage.
+
+2. Getting a data sample from a DataFrame
+    Querying big datasets can be expensive with BigQuery. To get a smaller sample of the current DataFrame use :py:meth:`DataFrame.get_sample()`:
+
+    .. code-block:: console
+
+        table_name = 'objectiv-production.writable_dataset.table_name'
+        df.get_sample(table_name, sample_percentage=10, overwrite=True)
```
thijs-obj, line 68: I think it is better to NOT use `overwrite=True` in examples. `overwrite=True` can be dangerous if the table_name already exists. Users could assume that example code is 'safe', so we need to have safe code or clear warnings. 
```suggestion
        df.get_sample(table_name, sample_percentage=10)
```

bach/docs/source/bach/core-concepts.rst
```diff
@@ -50,4 +50,38 @@ situations:
...
+
+    It creates a permanent table, and if the table exists it overwrites it (make sure you have a write access). For BigQuery `seed` parameter is not implemented.
```
thijs-obj, line 70: I would remove this part: "For BigQuery `seed` parameter is not implemented."

The goal of this tip is to point out to the reader that querying big tables can be expensive and that get_sample() is an easy way to get a smaller dataset.

Details about what parameters are supported just distracts from that. Users can find it in the api reference, if they are interested. Most users won't even know that the parameter exists, and to them it will be double confusing.\

bach/docs/source/bach/core-concepts.rst
```diff
@@ -50,4 +50,38 @@ situations:
...
+
+3. Using temporary tables to limit query complexity
+    Sometimes complex operations on Bach cannot be executed on BigQuery and you might get an error:
+
+        Resources exceeded during query execution: Not enough resources for query planning - too many subqueries or query is too complex.
+
+    So, if the underlying SQL query is complex for your data and you still need to apply other operations, in order to avoid the query to become too complex one has to first materialize the current DataFrame as a temporary table:
```
thijs-obj, line 77: The right formulation doesn't quite come to me at this moment. But we should be very clear that if users get the error above, then they need to take some steps back and materialize. If the error occurs, then it's too late to materialize

ivarpruijn:
@vard-obj As mentioned last week, I opened a new PR that covers your content and more (https://github.com/objectiv/objectiv-analytics/pull/1029). Shall we close this one?

Issue #806, opened 2022-06-01
Bach BQ: boolean min/max


thijs-obj:
* Fix min() and max() on boolean series
* Make sure all `SeriesBoolean` and `describe()` tests work

PR #1009, opened 2022-07-12
Make boolean min() and max() work on BQ

thijs-obj:
Fixes a small issue: `SeriesBoolean.min()` and max() on BigQuery

This solves https://github.com/objectiv/objectiv-analytics/issues/806


KathiaBarahona (approved):
lgtm

PR #1011, opened 2022-07-12
Support input tracking for HTMLSelectElements in Browser/Angular SDKs

sdirosa:
Our previous implementation only supported tracking HTMLInputElements. These changes add support to HTMLSelectElement as well. 

`tagInput` can be used for both inputs and selects now, including value tracking. 

Value tracking for selects will populate the InputValueContext with the selected option value, or if not present, the option text.

We also fixed a bug in the validation of the options which could cause unexpected options to pass through. Now the tagInput options are stricter.

**NOTE**: this version is on NPM as v0.0.22-3@next

**TIP**: filter out JSON files for a much easier to review diff

### DIFF
--- a/tracker/trackers/browser/src/common/guards/isTagLocationOptions.ts
+++ b/tracker/trackers/browser/src/common/guards/isTagLocationOptions.ts
@@ -2,8 +2,8 @@
  * Copyright 2021-2022 Objectiv B.V.
  */
 
-import { isTrackBlursAttribute } from '@objectiv/tracker-browser';
 import { TagLocationOptions } from '../../definitions/TagLocationOptions';
+import { isTrackBlursAttribute } from './isTrackBlursAttribute';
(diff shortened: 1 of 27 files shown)


tracker/trackers/browser/src/common/guards/isTagLocationOptions.ts
```diff
@@ -19,7 +20,7 @@ export const isTagLocationOptions = (object: Partial<TagLocationOptions>): objec
     return false;
   }
 
-  if (object.trackBlurs !== undefined && object.trackBlurs !== true && object.trackBlurs !== false) {
+  if (object.trackBlurs !== undefined && !isTrackBlursAttribute(object.trackBlurs)) {
     return false;
   }
```
sdirosa, line 25: This makes the options validation much stricter.

PR #1015, opened 2022-07-13
ModelHub: table format using list-table directive

KathiaBarahona:


ivarpruijn (approved):
Looks good in Docusaurus now

Issue #854, opened 2022-06-08
modelhub: make mypy effective


thijs-obj:
Currently running mypy on modelhub does not do much, as it doesn't complain if it cannot find types for imports.

1. We should add the following parameters to mypy for modelhub (both Makefile and github action): `--check-untyped-defs --follow-imports=error`. This will cause things to fail initially, see step two also.
2. The `objectiv_bach` package should properly tell mypy that it supports types. See https://peps.python.org/pep-0561/#packaging-type-information
3. Fix any remaining issues with mypy

Can of course also be split in two issues if this turns out to be a lot of work.

thijs-obj:
Done in #1020. Actually not using --follow-imports=error, like in this issue description, but `normal` which is what we need, in my current thinking

PR #1020, opened 2022-07-14
Bach modelhub mypy

thijs-obj:
Changes:
* Set mypy parameter `check_untyped_defs=True` for modelhub. This makes mypy check all code in modelhub. Without this it only checks functions that have at least one type-annotation.
  * Add an `type: ignore` in a place where we mistreat types on purpose that was previously not checked by mypy.
* Move check_untyped_defs parameter in bach from commandline to `mypy.ini`

This, together with #1019, resolves issue https://github.com/objectiv/objectiv-analytics/issues/854

### DIFF
--- a/modelhub/modelhub/series/series_objectiv.py
+++ b/modelhub/modelhub/series/series_objectiv.py
@@ -175,7 +175,8 @@ class SeriesLocationStack(SeriesJson):
 
             Returns the navigation stack from the location stack.
             """
-            return self[{'_type': 'NavigationContext'}: None]
+            # type ignore, as mypy doesn't like a dict in a Slice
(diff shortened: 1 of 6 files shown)


modelhub/modelhub/series/series_objectiv.py
```diff
@@ -175,7 +175,8 @@ def navigation_features(self):
 
             Returns the navigation stack from the location stack.
             """
-            return self[{'_type': 'NavigationContext'}: None]
+            # type ignore, as mypy doesn't like a dict in a Slice
+            return self[{'_type': 'NavigationContext'}: None]  # type: ignore
```
thijs-obj, line 179: Was actually surprised that this was the only thing that broke. So we were doing pretty good :)

PR #1031, opened 2022-07-15
Configurable TrackerEvent id generator

sdirosa:
These changes have been prereleased as v0.0.23-1@next

---

Currently, UUIDs are generated via a helper method, provided by Core Tracker, which relies on the [uuid](https://github.com/uuidjs/uuid) package.

In this PR, while leaving the default behavior unchanged, we introduce the option to override the UUID generation implementation with a custom one, when configuring a Tracker instance.

Example:
```ts
const tracker = new Tracker({ 
  applicationId: 'test-app-id', 
  generateUUID: () => 'not-so-unique-after-all' 
});
```

This will allow us to experiment with alternative identifier generation methods without having to rebuild our SDKs. It also provides a way for platform-specific SDKs to use their own implementations. Such as React Native, which lacks crypto.

For example we can easily test our custom implementation we theorized in https://github.com/objectiv/objectiv-analytics/pull/940 and implemented here as the new `uuidv4` module:

```ts
import { uuidv4 } from '@objectiv/tracker-core';

const tracker = new Tracker({ 
  applicationId: 'test-app-id', 
  generateUUID: uuidv4
});
```

It's also possible to invoke the specific implementations:
```ts
import { uuidv4 } from '@objectiv/tracker-core';

// Pick the best available implementation for this environment
uuidv4();

// Or use one of the implementations directly, this is mostly useful for testing / debugging
uuidv4.crypto_RandomUUID();
uuidv4.crypto_GetRandomValues();
uuidv4.dateNow_MathRandom();
```


tracker/core/tracker/src/TrackerEvent.ts
```diff
@@ -11,53 +11,54 @@ import {
...
+     */
+    id: string;
 
-/**
- * An Event before it has been handed over to the Tracker.
- * Properties that will be set by the Tracker or Transport are omitted: `id`, `time`
- */
-export type UntrackedEvent = Omit<AbstractEvent, 'id' | 'time'>;
+    /**
+     * The tracking time.
```
jansentom, line 33: in s or ms? Might be nice to add.

jansentom (commented):
:+1:  Tests are really easy to follow for the logic.

Your explaination on Chat that basically every browser that can run a modern framework probably also belongs in uuidv4.ts. I'm not sure how react native behaves though?

tracker/core/tracker/src/TrackerEvent.ts
sdirosa: Actually I can inherit this from AbstractEvent of the schema, so no need to redefine it here. Thanks for pointing it out

sdirosa: Fixed!

sdirosa:
> I'm not sure how react native behaves though?

Nice one. React Native doesn't have crypto, instead relies on a [polyfill implementing getRandomValues](https://www.npmjs.com/package/react-native-get-random-values).

The polyfill is a dependency of the React Native SDK. This means that this module will detect getRandomValues as being present and use it, effectively using the polyfill.

tracker/core/tracker/src/Tracker.ts
```diff
@@ -32,6 +32,11 @@ export type TrackerConfig = ContextsConfig & {
    */
   applicationId: string;
 
+  /**
+   * A function used by the Tracker to set the `id` of a TrackerEvent. Defaults to the `generateUUID` helper.
```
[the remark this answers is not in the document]
sdirosa: I've added a better description of why UUIDs are important

Issue #1036, opened 2022-07-18
Bach: Series.__init__ does not validate if series in `index` have the same base_node as the passed `base_node`


KathiaBarahona:
In DataFrame class when instancing a new DataFrame, we actually validate if all passed series have the same base_node as the parent DataFrame. We should do this for Series class as well

PR #1037, opened 2022-07-18
Bach: Refactor Series sorting behavior

KathiaBarahona:
Removed `Series.index_sorting` and `Series.sorted_ascending` properties. Now it relies only in `order_by` property as we do in `DataFrame`


bach/bach/operations/concat.py
```diff
@@ -321,15 +321,19 @@ def _get_concatenated_object(self) -> Series:
...
+            base_node=node,
             name=final_result_series.name,
             expression=Expression.column_reference(final_result_series.name),
-            index={} if self.ignore_index else main_series.index,
+            index={} if self.ignore_index else index,
             group_by=None,
-            sorted_ascending=None,
-            index_sorting=[] if self.ignore_index else main_series.index_sorting,
+            order_by=[],
             instance_dtype=series_cls.dtype  # TODO: make this work for structural types too
```
KathiaBarahona, line 337: `Series.__init__` is not validating if indexes have same node as parent Series. Opened issue: #1036

bach/bach/series/series.py
```diff
@@ -538,18 +507,13 @@ def copy_override(
         Special case:
         * If index is not None, then index_sorting is automatically set to `[]` unless overridden
```
KathiaBarahona, line 508: I'll remove this

bach/bach/series/series.py
```diff
@@ -125,8 +125,6 @@ def __init__(
         name: str,
         expression: Expression,
         group_by: Optional['GroupBy'],
-        sorted_ascending: Optional[bool],
-        index_sorting: List[bool],
         instance_dtype: StructuredDtype,
         order_by: Optional[List[SortColumn]] = None,
```
thijs-obj, line 129: request: could we make this non-optional? I have the feeling that having this optional just hides places where we forget to call this correctly.

```suggestion
        order_by: List[SortColumn],
```

Additionally, this will save like a line of code in this function

bach/bach/series/series.py
```diff
@@ -748,10 +714,16 @@ def __set_item_with_merge(self, other: 'Series') -> Tuple['Series', 'Series']:
             # incorrect second expression = a + b - b
             # correct second expression = a + b__other - b
             caller_expr = self.expression.replace_column_references(other.name, f'{other.name}__other')
-            caller_series = self.copy_override(base_node=df.base_node, expression=caller_expr)
+            caller_series = self.copy_override(
```
thijs-obj, line 717: Question: Is this an unrelated fix?

bach/bach/series/series.py
```diff
@@ -932,25 +901,14 @@ def to_frame(self) -> DataFrame:
 
         The DataFrame returned has the grouping and sorting also set like this Series had.
         """
-        if self._sorted_ascending is not None:
```
thijs-obj, line 935: nice to get this cleaned up :ok

thijs-obj (changes requested):
When searching for `index_sorting` I find 8 occurrences. `sorted_ascending` gives 3 results. From the looks of it all of those need to be changed.

Nice to get this cleaned up!

bach/bach/series/series.py
KathiaBarahona: Related to issue https://github.com/objectiv/objectiv-analytics/issues/1036. Since now we base_nodes when setting sort, it was raising an error

bach/bach/series/series.py
KathiaBarahona: mmm has the same typing as `DataFrame.__init__`. Was aiming more for consistency.

KathiaBarahona:
> When searching for `index_sorting` I find 8 occurrences. `sorted_ascending` gives 3 results. From the looks of it all of those need to be changed.
> 
> Nice to get this cleaned up!

Thanks for checking! Missed a couple of them, fixed here https://github.com/objectiv/objectiv-analytics/pull/1037/commits/433dd32f24d3483112e5007076cc25f3931bf84c

bach/bach/series/series.py
KathiaBarahona: Matching what the docstring says  https://github.com/objectiv/objectiv-analytics/blob/17c87cea3ea87cca0dcb6842a7d6b77d30025ee3/bach/bach/dataframe.py#L192

thijs-obj: okay I think we disagree. I just think it is slightly better, because it saves a bit of code and is less ambiguous between functions. You feel different, that's fine. Let's go with the changes you have made.

bach/bach/series/series.py
thijs-obj: ack :+1:

thijs-obj (approved):

PR #1041, opened 2022-07-18
useOnChange equality check via isContextEqual

sdirosa:
~~Simply use JSON stringification to verify whether the state given to `useOnChange` has changed, instead of using fast-deep-equal.~~

~~For all intent and purposes we don't really need deep referential checks for these objects. JSON.stringify works just fine for all of our use cases.~~

Swapped fast-deep-equal with a custom predicate, `isContextEqual`, to validate and compare two AbstractContext based objects.

Performance is more or less the same as fast-deep-equal but the validation of the input is much more strict.

I've also removed the redundant `useTrackOnChange`. Instead I've improved `useOnToggle` (and consequently `useTrackOnToggle` as well) and made it so it can monitor both boolean variables and predicates; useOnToggle and useTrackOnToggle can now be used in combination with any comparison library.


tracker/core/react/src/index.ts
```diff
@@ -52,13 +52,12 @@ export * from './hooks/eventTrackers/useSuccessEventTracker';
...
 export * from './hooks/eventTrackers/useVisibilityTracker';
 
-export * from './hooks/useOnChange';
+export * from './hooks/useOnContextChange';
 export * from './hooks/useOnMount';
 export * from './hooks/useOnMountOnce';
 export * from './hooks/useOnToggle';
 export * from './hooks/useOnUnmount';
 export * from './hooks/useOnUnmountOnce';
-export * from './hooks/useTrackOnChange';
```
sdirosa, line 61: This hook was useless. We can use `useOnToggle` and pass any comparison predicate, which is a lot more flexible than this ever was.

tracker/core/tracker/src/isContextEqual.ts
```diff
@@ -0,0 +1,48 @@
...
+  for (let i = 0; i < aLength; i++) {
+    const aKey = aKeys[i];
+
+    if (a[aKey] !== b[aKey]) {
+      return false;
+    }
+  }
+
+  return true;
+};
```
sdirosa, line 48: This has the same performance, slightly faster actually, than fast-deep-equal with the added benefit that it does a lot more validation to work only with valid Contexts from our Taxonomy, instead of any object.

jansentom: that should be more than fast enough. thanks for the explanation

sdirosa: Forgot to post the link to the performance test: https://www.measurethat.net/Benchmarks/Show/19905/0/iscontextequal-vs-fast-deep-equal#latest_results_block

jansentom: that's quite close!

PR #1046, opened 2022-07-19
Bach: Speed up slow tests

KathiaBarahona:
Github checks are veeery slow, so changed a bit some of the slowest tests. Still missing some, but this will help for now. Also, I think we should return a pandas DataFrame when getting any of the default testing dataframes. Some times we do:
``` python
bt = get_df_with_test_data(engine, full_data_set=True)
pdf = bt.to_pandas()
```
will be much easier (and efficient) to have:
``` python
pdf, bt = get_df_with_test_data(engine, full_data_set=True)
```


KathiaBarahona:
```
13.07s call     tests/functional/bach/test_df_window.py::test_rolling_defaults_vs_pandas
9.20s call     tests/functional/bach/test_df_database_create.py::test_database_create_table[engine1]
8.24s call     tests/functional/bach/test_df_sample.py::test_sample_w_temp_tables[engine1]
6.59s call     tests/functional/bach/test_df_sort_values.py::test_sort_values_parameters[engine1]
6.52s call     tests/functional/bach/test_df_groupby.py::test_groupby_frame_split_series_aggregation[engine1]
```
After changes with this PR, those are the slowest test (when running in my pc)

PR #1051, opened 2022-07-20
WIP: Do not assume `None` is a string  - Bach BQ test fix

thijs-obj:
Changes:
* Do not assume that `None` is a string. This was not correct on BigQuery
* Fix test `test_merge_right_join_shared_on` and some clean-up

Some background. In this query: `select x, None as y from some_table`, the value `y` will be of type `text` on Postgres, but of type `INT64` on BigQuery. In the past we just automatically made all `None` values `SeriesString`, that could lead to errors. E.g. in the merge test, the generated SQL compared a `STRING` with a `INT64`, which lead to an error.


bach/tests/functional/bach/test_df_merge.py
```diff
@@ -302,15 +301,18 @@ def test_merge_right_join(engine):
...
+    bt = bt.materialize()
+
     bt = bt.reset_index(drop=True)
 
-    mt = get_bt_with_railway_data()[['station', 'platforms']]
+    mt = get_df_with_railway_data(engine)[['station', 'platforms']]
     mt = mt.reset_index(drop=True)
 
     result = bt.merge(mt, how='right', on='station')
+    result = result.sort_values('station')
```
thijs-obj, line 315: Without this the results were non-deterministic

thijs-obj:
Closing in favor of https://github.com/objectiv/objectiv-analytics/pull/1066. accidentally rebased after pushing, so had to create new branch

PR #1054, opened 2022-07-20
Bach: dont run pg tests twice in CI

jansentom:
Make sure we don't do double work for PG in our workflows.


KathiaBarahona (approved):

bach/tests/conftest.py
```diff
@@ -76,7 +76,11 @@ class DB(Enum):
 
 
 @pytest.fixture()
-def pg_engine() -> Engine:
+def pg_engine(request: SubRequest) -> Engine:
+    if request.config.getoption("big_query"):
```
thijs-obj, line 80: we could do
```suggestion
    if DB.POSTGRES not in _ENGINE_CACHE:
```
then it would also keep working if we add, for example, an `--athena` option

thijs-obj (approved):
Small suggestion, but without also LGTM

bach/tests/conftest.py
jansentom: like it, applied

PR #1061, opened 2022-07-20
Bach: SeriesJson.json.flatten_array implementation

KathiaBarahona:


thijs-obj (approved):
Not 100% sure how the queries will perform that we generate here, but can't be too bad I think.

LGTM

PR #1062, opened 2022-07-21
Add aggregate function to SeriesString: to_json_array

thijs-obj:
**Change** New aggregation function: `SeriesString.to_json_array`

Please have a look @KathiaBarahona and @vard-obj 

@vard-obj: does this full-fill your requirements?

### DIFF
--- a/bach/bach/series/series_string.py
+++ b/bach/bach/series/series_string.py
@@ -229,7 +229,7 @@ class SeriesString(Series):
     def _comparator_operation(self, other, comparator, other_dtypes=tuple(['string'])) -> 'SeriesBoolean':
         return super()._comparator_operation(other, comparator, other_dtypes)
 
-    def to_json_array(self, window: Optional[WrappedWindow] = None) -> 'SeriesJson':
+    def to_json_array(self, partition: Optional[WrappedPartition] = None) -> 'SeriesJson':
         """
         Aggregate function: Group the values of this Series into a json array
 
--- a/bach/bach/series/series_string.py
+++ b/bach/bach/series/series_string.py
@@ -234,12 +234,13 @@ class SeriesString(Series):
         Aggregate function: Concatenate the values of this Series into a json array
 
         The order of the values in the array will be based of the order of the values in this Series. If
-        this Series does not have a deterministic sorting, then the values are sorted by the values
-        themselves. A difference in sorting can occur between the resulting array and the values in a Series
-        when the Series contains None/NULL values. Null values will be sorted first when sorting ascending,
-        contrary to sorting inside a DataFrame or a Series where they come last.
+        this Series does not have a deterministic sorting, then the values are additionally sorted by the
+        values themselves. A difference in sorting can occur between the resulting array and the values in a
+        Series when the Series contains None/NULL values. Null values will be sorted first when sorting
+        ascending, contrary to sorting inside a DataFrame or a Series where they come last.
 
-        :param partition: The partition to apply
+        :param partition: The partition to apply, optional.
+        :return: SeriesJson containing an array of strings on each row.
         """
         order_by = self.order_by
         # Add this series as the final column to sort on. If the order_by is deterministic then this won't
--- a/bach/bach/series/series_string.py
+++ b/bach/bach/series/series_string.py
@@ -290,5 +290,4 @@ class SeriesString(Series):
                 expr = Expression.construct('{} {}', level_expr, asc_expr)
                 expressions.append(expr)
 
-        join_expressions(expressions)
         return Expression.construct('order by {}', join_expressions(expressions))
--- a/bach/bach/series/series_string.py
+++ b/bach/bach/series/series_string.py
@@ -272,13 +272,12 @@ class SeriesString(Series):
 
         # This logic is partially duplicating DataFrame._get_order_by_clause(), but we can't quite re-use
         # that as some of the checks there are not applicable (e.g. it's fine to use a column that we are not
-        # grouping on).
+        # grouping on), and the sorting of null values is slightly different.
         if not order_by:
             raise ValueError('order_by is empty')
 
         expressions: List[Expression] = []
         for sc in order_by:
-            # pandas sorts by default all nulls last
             if sc.expression.has_multi_level_expressions:
                 multi_lvl_exprs = [
                     level_expr for level_expr in sc.expression.data if isinstance(level_expr, Expression)
--- a/bach/bach/series/series_string.py
+++ b/bach/bach/series/series_string.py
@@ -2,18 +2,20 @@
 Copyright 2021 Objectiv B.V.
 """
 import re
-from typing import Union, TYPE_CHECKING, Optional, Pattern
+from typing import Union, TYPE_CHECKING, Optional, Pattern, List
 
 from sqlalchemy.engine import Dialect
 
 from bach.series import Series
-from bach.expression import Expression
+from bach.expression import Expression, AggregateFunctionExpression, join_expressions
+from bach.series.series import WrappedPartition
 from bach.types import StructuredDtype
 from sql_models.constants import DBDialect
+from sql_models.util import DatabaseNotSupportedException, is_bigquery, is_postgres
 
 if TYPE_CHECKING:
     from bach.series import SeriesBoolean
-    from bach import DataFrame
+    from bach import DataFrame, SortColumn, SeriesJson
 
 
 class StringOperation:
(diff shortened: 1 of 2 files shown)


bach/bach/series/series_string.py
```diff
@@ -226,3 +228,67 @@ def __add__(self, other) -> 'Series':
...
+        this Series does not have a deterministic sorting, then the values are sorted by the values
+        themselves. A difference in sorting can occur between the resulting array and the values in a Series
+        when the Series contains None/NULL values. Null values will be sorted first when sorting ascending,
+        contrary to sorting inside a DataFrame or a Series where they come last.
+
+        :param partition: The partition to apply
+        """
+        order_by = self.order_by
+        # Add this series as the final column to sort on. If the order_by is deterministic then this won't
+        # change the sorting. If order_by was not deterministic, then this will make it deterministic.
```
thijs-obj, line 246: We could probably use this trick in other places too, to get rid of non-determinism

bach/bach/series/series_string.py
```diff
@@ -226,3 +228,66 @@ def __add__(self, other) -> 'Series':
 
     def _comparator_operation(self, other, comparator, other_dtypes=tuple(['string'])) -> 'SeriesBoolean':
         return super()._comparator_operation(other, comparator, other_dtypes)
+
+    def to_json_array(self, partition: WrappedPartition = None) -> 'SeriesJson':
```
thijs-obj, line 232: Not 100% sure about the name, was also thinking about things like `aggregate_to_json_array`

thijs-obj, line 232: Support the partition parameter here, as we also seem to do that for other aggregation functions

bach/bach/series/series_string.py
```diff
@@ -226,3 +228,66 @@ def __add__(self, other) -> 'Series':
...
+        return result.copy_override_type(SeriesJson)
+
+    @staticmethod
+    def _get_order_by_expression(order_by: List['SortColumn']) -> Expression:
+        """
+        Internal helper function: Convert order_by into an order by expression that is usable inside an
+        array_agg() function
+
+        Note: ordering is slightly different from regular ordering of rows inside a DataFrame or Series, as
+        BigQuery does not support 'asc nulls last' inside aggregations. So we sort 'asc nulls first'.
```
thijs-obj, line 270: This is quite unfortunate, but didn't know a work-around

bach/tests/functional/bach/test_series_string.py
```diff
@@ -131,3 +131,75 @@ def test_string_replace(engine) -> None:
...
     assert_equals_data(result_df, expected_columns=expected_columns, expected_data=expected_data)
+
+
+def test_to_json_array(engine):
+    df = get_df_with_test_data(engine, full_data_set=True)
+    s_muni = df['municipality']
+    assert isinstance(s_muni, SeriesString)
+
+    series_json_array = s_muni.to_json_array()
+    print(series_json_array.view_sql())
```
vard-obj, line 142: guess the `print` is not needed.

vard-obj (approved):
Left minor comments, LGTM.

bach/bach/series/series_string.py
```diff
@@ -226,3 +228,66 @@ def __add__(self, other) -> 'Series':
 
     def _comparator_operation(self, other, comparator, other_dtypes=tuple(['string'])) -> 'SeriesBoolean':
         return super()._comparator_operation(other, comparator, other_dtypes)
+
+    def to_json_array(self, partition: WrappedPartition = None) -> 'SeriesJson':
+        """
+        Aggregate function: Concatenate the values of this Series into a json array
```
KathiaBarahona, line 234: IMHO `Concatenate` can get mixed up with "concatenating" arrays.
```suggestion
        Aggregate function: Group the values of this Series into a json array
```

bach/bach/series/series_string.py
```diff
@@ -226,3 +228,66 @@ def __add__(self, other) -> 'Series':
...
+
+        The order of the values in the array will be based of the order of the values in this Series. If
+        this Series does not have a deterministic sorting, then the values are sorted by the values
```
KathiaBarahona, line 237: **question** Shouldn't we also allow the user to decide if the values should be in ascending or descending order?

KathiaBarahona: Also I think the following sentence is a bit confusing ``If
        this Series does not have a deterministic sorting, then the values are sorted by the values themselves.``

bach/bach/series/series_string.py
KathiaBarahona: mm can we a single function that does this and maybe just add a parameter if `nulls last` statement should be added? I mean we already have this function in `DataFrame` and `Window` class. Much be easier for maintainability.

bach/bach/series/series_string.py
KathiaBarahona: Based on BigQuery's documentation, this is actually a window function. IMO, naming should be `window_array_agg`. Documentation can help telling user this actually returns a json series.

KathiaBarahona, line 232: Also, we should add a check to know if the partition is correct.
```suggestion
    def to_json_array(self, window: Optional[WrappedWindow] = None) -> 'SeriesJson':
```

KathiaBarahona (changes requested):
As I mentioned in the comments, this functionality is a window function in a nutshell. It seems to me there might be some code duplication, and all this can be handled by Window class (as we do for the rest of window functions)

bach/tests/functional/bach/test_series_string.py
thijs-obj: good one. I was a bit sloppy in looking over this before submitting, my bad.
```suggestion
```

bach/tests/functional/bach/test_series_string.py
```diff
@@ -131,3 +131,75 @@ def test_string_replace(engine) -> None:
...
+            ['De Friese Meren', 'Harlingen', 'Leeuwarden', 'Noardeast-Fryslân', 'Súdwest-Fryslân',
+             'Súdwest-Fryslân', 'Súdwest-Fryslân', 'Súdwest-Fryslân', 'Súdwest-Fryslân', 'Súdwest-Fryslân',
+             'Waadhoeke']
+        ]]
+    )
+
+
+def test_to_json_array_sorting(engine):
+    df = get_df_with_test_data(engine, full_data_set=True)
+    series_city = df['city']
```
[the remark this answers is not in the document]
thijs-obj:
```suggestion
```

bach/bach/series/series_string.py
thijs-obj: > **question** Shouldn't we also allow the user to decide if the values should be in ascending or descending order?

The user can decide whether the sorting is ascending/descending. For example to get the values sorted descending one can simply do: 
```python
series_object.sort_values(ascending=False).to_json_array()
```

thijs-obj: > Also I think the following sentence is a bit confusing `If this Series does not have a deterministic sorting, then the values are sorted by the values themselves.`

I am going to change this to
`If this Series does not have a deterministic sorting, then the values are additionally sorted by the values themselves.`
i.e. add an 'additionally' in there. Does that clarify things? An alternative could be just leaving this sentence out completely. Or do you have another textual suggestion?

bach/bach/series/series_string.py
thijs-obj: I don't think BigQuery considers this a window function, it is listed under "Aggregate functions": https://cloud.google.com/bigquery/docs/reference/standard-sql/aggregate_functions#array_agg

Same for Postgres: https://www.postgresql.org/docs/14/functions-aggregate.html

thijs-obj: Undid this change again, as I don't think this is a Window function. See other comment

bach/bach/series/series_string.py
thijs-obj: I created a separate PR that unites this function with the one in `Window`. I didn't include the one in `DataFrame` in that PR because there are some more differences between those, e.g. in the checks that DataFrame does.

https://github.com/objectiv/objectiv-analytics/pull/1070

bach/tests/functional/bach/test_series_string.py
```diff
@@ -131,3 +132,84 @@ def test_string_replace(engine) -> None:
...
+        # single row, single column, with a list of strings in that cell
+        expected_data=[[
+            ['De Friese Meren', 'Harlingen', 'Leeuwarden', 'Noardeast-Fryslân', 'Súdwest-Fryslân',
+             'Súdwest-Fryslân', 'Súdwest-Fryslân', 'Súdwest-Fryslân', 'Súdwest-Fryslân', 'Súdwest-Fryslân',
+             'Waadhoeke']
+        ]]
+    )
+
+
+def test_to_json_array_sorting_null(engine):
```
thijs-obj, line 157: Completely changed this test since the first review round to also include `None` values

bach/bach/series/series_string.py
KathiaBarahona: ah wait, I think I misunderstood the expression you are using.  As in the docs it seems its possible to define a window specification.

KathiaBarahona (approved):

Issue #1025, opened 2022-07-14
Fix BQ tests.functional.bach.test_df_merge.test_merge_right_join_shared_on


thijs-obj:
* Fix underlying issue in Bach
* Make test pass
* Remove `get_bt_with_railway_data()` when no longer used

thijs-obj:
Fixed by #1066

PR #1066, opened 2022-07-21
Do not assume None is a string - Bach BQ test fix

thijs-obj:
Changes:
* Do not assume that `None` is a string. This was not correct on BigQuery
* Fix test `test_merge_right_join_shared_on` and some clean-up

**Some background** In this query: `select x, None as y from some_table`, the value `y` will be of type `text` on Postgres, but of type `INT64` on BigQuery. In the past we just automatically made all `None` values `SeriesString`, that could lead to errors. E.g. in the merge test, the generated SQL compared a `STRING` with a `INT64`, which lead to an error.

This addresses issue https://github.com/objectiv/objectiv-analytics/issues/1025


bach/tests/functional/bach/test_df_merge.py
```diff
@@ -302,15 +301,18 @@ def test_merge_right_join(engine):
...
+    bt = bt.materialize()
+
     bt = bt.reset_index(drop=True)
 
-    mt = get_bt_with_railway_data()[['station', 'platforms']]
+    mt = get_df_with_railway_data(engine)[['station', 'platforms']]
     mt = mt.reset_index(drop=True)
 
     result = bt.merge(mt, how='right', on='station')
+    result = result.sort_values('station')
```
thijs-obj, line 315: Without this the results were non-deterministic

bach/tests/functional/bach/test_df_merge.py
```diff
@@ -302,15 +301,18 @@ def test_merge_right_join(engine):
     )
 
 
-def test_merge_right_join_shared_on() -> None:
-    bt = get_bt_with_test_data(full_data_set=False)[['skating_order', 'city']]
-    bt['station'] = None
+def test_merge_right_join_shared_on(engine) -> None:
```
thijs-obj, line 304: There were actually two issue that made this test fail:
1. `None` is an `INT64` type in BigQuery, and we compared it here against `station`, which is a `STRING`
2. BigQuery doesn't allow `join on left.stations = NULL`, which we generated.

This PR fixes issue one. Issue two is worked around by adding a `materialize()` below. I don't think this is really an urgent issue, as I imagine this would be quite a rare thing to trigger and takes a bit more work to fix. I'll create an issue for this

TODO

bach/bach/dataframe.py
```diff
@@ -3329,7 +3329,7 @@ def unstack(
                 new_column_name = f'{column}__{curr_col}'
                 new_columns.append(new_column_name)
 
-                df[new_column_name] = None
+                df[new_column_name] = df[curr_col].from_value(base=df, value=None)
```
thijs-obj, line 3332: I also considered to introduce some special object `UnassignedSeries` , that wouldn't require a `base_node` and that we would get from the DataFrame on assignment. Then we could write this perhaps as

```suggestion
                df[new_column_name] = df[curr_col].from_value(value=None)
```

But that change was a bit bigger and more impactful, and the only gain is that the user can drop the `base=df` part in function calls like this, so I threw that out.

KathiaBarahona: Can't we do the `from_value` in `DataFrame.__set_item__` when the value is None?

KathiaBarahona (commented):
Just a comment, though I am a bit confused regards the changes introduced with this PR. As they are more related on removing overwrites of `supported_literal_to_expression` method than actually just stop considering `None` as  a `SeriesString`

bach/bach/dataframe.py
thijs-obj: For new columns we wouldn't know the dtype. We wouldn't know whether to call `SeriesString.from_value()` or `SeriesInt64.from_value()`

thijs-obj:
> Just a comment, though I am a bit confused regards the changes introduced with this PR. As they are more related on removing overwrites of `supported_literal_to_expression` method than actually just stop considering `None` as a `SeriesString`

Yeah sorry this is a bit confusing. Before, there were some overwrites to `supported_literal_to_expression()` that did not do generate an `Expression` containing a `cast()`. However, now that `None` can be any type, we should always do a cast in the expression generated by `supported_literal_to_expression()`, because without a cast `None` will not be the correct type. So it didn't really make sense to not have a default implementation anymore, as all implementations should be the same. 
Then I actually made two exceptions to that again for readability of the generated SQL: `SeriestString` and `SeriesInt64` do some checks to not cast when it's not needed

jansentom (approved):
Dicsussed IRL. For now this is the quick way to do this. Let's go.

thijs-obj:
Had a look at all notebooks, and in none of them we use `None` as a value in a DataFrame/Series. So that should be good

KathiaBarahona:
> However, now that `None` can be any type, we should always do a cast in the expression generated by `supported_literal_to_expression()`, because without a cast `None` will not be the correct type. 

I feel this is not intuitive for users, I'll expect doing `df['x'] = None` will be enough

bach/tests/functional/bach/test_df_merge.py
thijs-obj: Actually, the second point got also solved by the solution to the first point. 

BQ doesn't like ` on (NULL = r.station)`, but doesn't mind ` on (cast(NULL as STRING) = r.station)`. So this is all good now

thijs-obj:
> > However, now that `None` can be any type, we should always do a cast in the expression generated by `supported_literal_to_expression()`, because without a cast `None` will not be the correct type.
> 
> I feel this is not intuitive for users, I'll expect doing `df['x'] = None` will be enough and the engine will "magically" know how to deal with it

What is you concrete proposal? What should we do if the user writes `df['x'] = None`?

I can make something, but if that's something else than what you had in mind then that might well be wasted effort.

thijs-obj:
As discussed in video chat: 462507977f2bbafaca9390d696a3c331cbe8a802
Some additional small changes after that: https://github.com/objectiv/objectiv-analytics/pull/1066/files/462507977f2bbafaca9390d696a3c331cbe8a802..3ddbee5

jansentom (approved):
LGTM!

KathiaBarahona (approved):

PR #1070, opened 2022-07-25
WIP: Bach - extracted order_by_expression function

thijs-obj:
Based on review comment https://github.com/objectiv/objectiv-analytics/pull/1062#discussion_r928295212

@KathiaBarahona does this match what you had intended?


bach/tests/unit/bach/test_series_window_functions.py
```diff
@@ -11,11 +11,11 @@ def test_window_row_number(dialect):
 
     result_sql = result.expression.to_sql(dialect)
     if is_bigquery(dialect):
-        assert result_sql == 'row_number() over ( order by `b` asc )'
+        assert result_sql == 'row_number() over ( order by `b` asc nulls first )'
```
thijs-obj, line 14: I think this actually changes behaviour of the window functions. Not sure if it's an improvement, but I do think that adding  things like 'nulls first' guarantees consistenty between BigQuery and Postgres, and that might not have been consistent before? :thinking:

KathiaBarahona: maybe adding a test comparing both behaviors? I think this change might introduce changes in modelhub without knowing

KathiaBarahona (approved):
lgtm, tho I'm not sure if `get_order_by_expression` should be in partitioning module.

thijs-obj:
> lgtm, tho I'm not sure if `get_order_by_expression` should be in partitioning module.

It's used by a window functions and an aggregate function, so this seemed like an okayish place to put it. Do you have a concrete proposal that you think is better?

bach/tests/unit/bach/test_series_window_functions.py
thijs-obj: I think having some tests is a good idea. But adding tests with nulls for all window functions is a quite a bit of work. The reason for this PR is to make maintainability [easier](https://github.com/objectiv/objectiv-analytics/pull/1062#discussion_r928295212), but I am not sure if that goal justifies me spending more time on this issue now to add a load of tests. So I propose that we then create a separate issue for this PR, and plan that in weekly planning. What do you think about that?

bach/bach/partitioning.py
```diff
@@ -498,3 +482,48 @@ def get_group_by_column_expression(self) -> Optional[Expression]:
...
+    # grouping on in an aggregate order by.
+    if not order_by:
+        return Expression.construct('')
+
+    if nulls_last:
+        asc_expr = Expression.construct('asc nulls last')
+        desc_expr = Expression.construct('desc nulls last')
+    else:
+        asc_expr = Expression.construct('asc nulls first')
+        desc_expr = Expression.construct('desc nulls last')
```
KathiaBarahona, line 515:
```suggestion
        desc_expr = Expression.construct('desc nulls first')
```

KathiaBarahona (changes requested):
desc + nulls first is wrong

thijs-obj:
> desc + nulls first is wrong

What do you mean by wrong? Is this different from what Pandas does, or how we did it in the past, or ... ?

KathiaBarahona:
> > desc + nulls first is wrong
> 
> What do you mean by wrong? Is this different from what Pandas does, or how we did it in the past, or ... ?

( I'm dumb, was replying on your comment)

See https://github.com/objectiv/objectiv-analytics/pull/1070#discussion_r934263974

when `nulls_last is False` and `order_by.asc is False`, the final expression in the order by clause is setting `ORDER BY COLUMN desc nulls last`

thijs-obj:
> > > desc + nulls first is wrong
> > 
> > 
> > What do you mean by wrong? Is this different from what Pandas does, or how we did it in the past, or ... ?
> 
> ( I'm dumb, was replying on your comment)
> 
> See [#1070 (comment)](https://github.com/objectiv/objectiv-analytics/pull/1070#discussion_r934263974)
> 
> when `nulls_last is False` and `order_by.asc is False`, the final expression in the order by clause is setting `ORDER BY COLUMN desc nulls last`
> 
> is yielding the same expression when having `nulls_last is True` and `order_by.asc is False`

Let's discuss tomorrow after stand-up. I am not quite following what you mean. To me it looks like the behaviour of the function matches what the docstring says it does, and the test for SeriesString keeps working. I agree that this might change things for window functions, but I believe your PR fixes that?

thijs-obj:
Closing this PR. PR https://github.com/objectiv/objectiv-analytics/pull/1089 contains all commits in this branch, with a lot more refinement added

Issue #1075, opened 2022-07-26
Investigate concurrency / sessions edge-cases


sdirosa:
When a user lands on an application triggering many non-interactive events at once (more than the batch size), the tracker instance will send batches of them concurrently to the collector. 

If the collector supports HTTP2, batches are going to be sent in parallel and will arrive, and be responded to, in random order and most likely at the same time. 

Since each of these batches has been created before the collector could assign a cookie to the client, each of them will receive a  different cookie and one of them will become the active one in FE.

PR #1076, opened 2022-07-26
Sync queue until session established

sdirosa:
Small change to TrackerQueue to send the very first batch synchronously.

I also bumped the prerelease version a couple of times to publish and test this up realistically.

**Review**. 
Filtered diff (no noise from version bumps) with the actual changes: 
- https://github.com/objectiv/objectiv-analytics/pull/1076/files?file-filters%5B%5D=.ts&show-viewed-files=true

**Testing**
This app is preconfigured with this branch:
- https://github.com/objectiv/objectiv-demo-apps/tree/main/sessions-playground 

**NPM**
This version is live on NPM as 0.0.23-next.1 on the next tag

Solves #1075


borft (commented):
.

borft (approved):
lgtm.

I think it's elegant, because it's simple. Only caveat is, that if the first request doesn't set the cookie for whatever reason, we still won't have the session. But to fix that, we probably need a much bigger and more complex change (if we can fix that at all).

sdirosa:
> But to fix that, we probably need a much bigger and more complex change (if we can fix that at all).

Yes. I think we would have to switch to some more explicit protocol between the tracker and the collector. 

Instead of relying on the clients to set up the cookie for us automatically, I would call a specific endpoint to establish a session and make sure to not send any events to remote until that happened successfully.

We could even reuse the endpoint we have now, e.g. call it without any events, just to get the response headers.

But, I agree with you, it's a lot more complicated change. It requires state and changes to the sending transports to do this extra work.

PR #1078, opened 2022-07-27
Bach/MH: stricter package requirements to avoid BQ deps clash

jansentom:
MH requires Bach (of course), and Bach requires `sqlalchemy-bigquery` which in turn has some pretty specific requirements. This PR makes sure pip knows about those before getting other dependencies, so in a clean env, everything installs without conflicts:

Observe the following to now be error-free.
```
python3 -m venv /tmp/venv-deps-test
. /tmp/venv-deps-test/bin/activate
pip install -e modelhub
```

### DIFF
--- a/modelhub/setup.cfg
+++ b/modelhub/setup.cfg
@@ -22,10 +22,15 @@ classifiers =
 
 [options]
 install_requires =
+    # sqlalchemy-bigquery (through objectiv-bach) is very specific about the following two deps. List them first to avoid pip getting confused
+    sqlalchemy<=1.4.27
+    pyarrow<7.0dev
     objectiv-bach ==0.0.17
     requests
     matplotlib
     seaborn
+    # sklearn requires wheel to properly install
+    wheel
     sklearn
     typing-extensions
 python_requires = >=3.7
--- a/bach/setup.cfg
+++ b/bach/setup.cfg
@@ -22,13 +22,14 @@ classifiers =
 
 [options]
 install_requires =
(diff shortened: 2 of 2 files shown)


modelhub/setup.cfg
```diff
@@ -22,10 +22,15 @@ classifiers =
...
 install_requires =
+    # sqlalchemy-bigquery (through objectiv-bach) is very specific about the following two deps. List them first to avoid pip getting confused
+    sqlalchemy<=1.4.27
+    pyarrow<7.0dev
     objectiv-bach ==0.0.17
     requests
     matplotlib
     seaborn
+    # sklearn requires wheel to properly install
+    wheel
```
thijs-obj, line 33: FYI, on my system it works fine without this line.

thijs-obj:
Only thing that worries me a bit is that for the automated tests we use different versions. e.g. `sqlalchemy>=1.4.32`. But I believe you are working on reconciling that too in future PR?

thijs-obj (commented):
Two comments. No objections to merging

jansentom:
> Only thing that worries me a bit is that for the automated tests we use different versions. e.g. `sqlalchemy>=1.4.32`. But I believe you are working on reconciling that too in future PR?

Yeah, that's what triggered that discussion. I'll tackle that in a next PR. Will not merge this until that's properly tested

modelhub/setup.cfg
jansentom: CI errors without it. You probably have it system wide.

KathiaBarahona (approved):

PR #1079, opened 2022-07-27
Client versions and session

sdirosa:
Changes:
- Each package now sets their name and version in a global `objectiv.versions` map, for debugging purposes
  - The snippet is repeated across al package because I don't want to introduce circular dependencies
- Tracker Core generates client session identifiers in a global `objectiv.clientSessionId`
  - The new identifier is added to all event batches, alongside the transport_time
- Modified the `event_list` schema to require `client_session_id`
  - Added a test to the collector to verify that event_list schema validation works

Extra:
- Renamed all monorepo `publish` tasks to `deploy` for consistency

**Testing**:
I've updated the react tracker in the sessions-playground app to test these two features:

- https://github.com/objectiv/objectiv-demo-apps/tree/main/sessions-playground

---

Screen Shot 2022-07-27 at 13 49 21


schema/event_list.json5
```diff
@@ -10,10 +10,14 @@
...
+      "type": "string",
+      "description": "A page/tab/screen session unique identifier generated on the client"
+    },
     "transport_time": {
       "type": "integer",
       "description": "Timestamp when the list of events was sent by the tracker"
     }
   },
-  "required": ["events", "transport_time"]
+  "required": ["events", "client_session_id", "transport_time"]
```
borft, line 22: if we require this, it means any events from a not yet updated tracker would be rejected by the collector. Is that what we want?

schema/event_list.json5
```diff
@@ -10,10 +10,14 @@
       },
       "description": "List/array of events. See the base_schema for the definition of an (abstract)event"
     },
+    "client_session_id": {
```
borft, line 13: maybe we should also bump the version here

schema/event_list.json5
sdirosa: Good point. Maybe not, I can remove this and leave it a soft-requirement, then perhaps make it obligatory only if we have some critical model running on this

schema/event_list.json5
sdirosa: Good one, I forgot to do that. Will fix

schema/event_list.json5
sdirosa: Fixed, I've made it optional for now

schema/event_list.json5
sdirosa: Patched the event_list schema to 0.0.2

borft (approved):
lgtm

PR #1088, opened 2022-07-29
Tracking SDKs v0.0.24: TS Schema interfaces, Angular SDK compatibility tweaks

sdirosa:
A user reported that on NPM 6, Node 10 and Angular 9.1.13 they get some "module not found" errors.
While debugging I found two issues and fixed them up:

**Made TS definitions compatible with TS 3.7**  
The previous TS Schema format had a few issues. Switching to simple interfaces makes validating the generated files easier and also they become compatible with stricter TS setups, like Angular 9.1.13 and Node 10.

**Fixed Angular package compatibility with NPM 6**  
NPM 6 is not smart enough to automatically load a package from a dist folder, if present. Instead it relies on module, types, files and other attributes to be set correctly in the package.json.  
I changed how the package is created to fill all the necessary fields and made sure what we have works also with Verdaccio.


tracker/core/schema/src/static.ts
```diff
@@ -2,7 +2,7 @@
  * Copyright 2021-2022 Objectiv B.V.
  */
 
-import { AbstractGlobalContext, AbstractLocationContext } from './abstracts';
+import { AbstractGlobalContext, AbstractLocationContext } from './generated';
```
sdirosa, line 5: This was broken, but it didn't matter because anything in .d.ts files becomes global, so it was just for IDEs. Now that we switched to TS files we had to fix it.

tracker/core/utilities/src/generateSchema.js
```diff
@@ -107,9 +107,9 @@ function createDefinition(
     if (params.properties[property]['type']) {
       p_list.push(`${property}: ${get_property_definition(params.properties[property])};\n`);
     } else if (params.properties[property]['discriminator']) {
-      p_list.push(`readonly ${property} = ${params.properties[property]['discriminator']};\n`);
+      p_list.push(`${property}: ${params.properties[property]['discriminator']};\n`);
     } else {
-      p_list.push(`readonly ${property}: ${params.properties[property]['value']};\n`);
+      p_list.push(`${property}: ${params.properties[property]['value']};\n`);
     }
   });
```
sdirosa, line 114: There's probably duplication and space for optimization here, now that there's basically no difference between abstracts and the other interfaces, but I didn't feel like refactoring everything so I just left it as is.

tracker/core/utilities/src/generateLanguageCodes.js
```diff
@@ -54,7 +54,7 @@ try {
 
     fs.appendFileSync(
       DESTINATION_FILENAME,
-      `export const LanguageCodes = [\n  '${filteredAndSortedLanguageCodes.join(`',\n  '`)}',\n];\n`
+      `export const LanguageCodes = [\n  '${uniqueSortedLanguageCodes.join(`',\n  '`)}',\n];\n`
     );
```
sdirosa, line 58: Small optimization here, unrelated.

tracker/trackers/angular/package.json
```diff
@@ -26,24 +26,34 @@
...
+  ],
+  "main": "./dist/bundles/objectiv-tracker-angular.umd.js",
+  "module": "./dist/fesm5/objectiv-tracker-angular.js",
+  "es2015": "./dist/fesm2015/objectiv-tracker-angular.js",
+  "esm5": "./dist/esm5/objectiv-tracker-angular.js",
+  "esm2015": "./dist/esm2015/objectiv-tracker-angular.js",
+  "fesm5": "./dist/fesm5/objectiv-tracker-angular.js",
+  "fesm2015": "./dist/fesm2015/objectiv-tracker-angular.js",
+  "typings": "./dist/objectiv-tracker-angular.d.ts",
+  "metadata": "./dist/objectiv-tracker-angular.metadata.json",
```
sdirosa, line 40: All of these properties are required by NPM 6

tracker/trackers/angular/package.json
```diff
@@ -26,24 +26,34 @@
...
     "deploy:verdaccio": "npm publish",
     "prettify": "prettier --write .",
     "tsc": "tsc --noEmit"
   },
   "peerDependencies": {
-    "@angular/common": "^9.0.0",
-    "@angular/core": "^9.0.0",
-    "@objectiv/tracker-browser": "^0.0.23"
+    "@angular/common": ">=9.0.0",
+    "@angular/core": ">=9.0.0",
```
sdirosa, line 49: Small fix here to remove warnings in console when installing on newer versions of Angular

tracker/trackers/angular/src/index.ts
```diff
@@ -2,6 +2,7 @@
  * Copyright 2021-2022 Objectiv B.V.
  */
 
+export * from './AngularTracker';
```
sdirosa, line 5: This is not necessary to be exported, but handy for testing and debugging, so I added it.

PR #1089, opened 2022-08-01
Bach: Simulate NULLS FIRST and NULLS LAST for BigQuery in window functions

KathiaBarahona:


bach/bach/series/series_string.py
```diff
@@ -246,7 +246,7 @@ def to_json_array(self, partition: Optional[WrappedPartition] = None) -> 'Series
         The order of the values in the array will be based of the order of the values in this Series. If
         this Series does not have a deterministic sorting, then the values are additionally sorted by the
         values themselves. A difference in sorting can occur between the resulting array and the values in a
-        Series when the Series contains None/NULL values. Null values will be sorted first when sorting
+        Series when the Series contains None/NULL values. Null values will be sorted last when sorting
         ascending, contrary to sorting inside a DataFrame or a Series where they come last.
```
thijs-obj, line 250: > Null values will be sorted last when sorting
        ascending, contrary to sorting inside a DataFrame or a Series where they come last.

This sentence doesn't make sense anymore. Perhaps the one before it doesn't make sense anymore too

bach/bach/partitioning.py
```diff
@@ -484,7 +491,9 @@ def get_group_by_column_expression(self) -> Optional[Expression]:
         return None
 
 
-def get_order_by_expression(order_by: List['SortColumn'], nulls_last: bool = False) -> Expression:
+def get_order_by_expression(
+    dialect: Dialect, order_by: List['SortColumn'], nulls_last: bool = False,
+) -> Expression:
     """
     INTERNAL: Convert order_by into an order by expression that is usable inside an aggregation or window
     function.
```
thijs-obj, line 499: There are some descriptions about the sorting in this docstring that are no longer correct

thijs-obj: e.g. it says `By default (=False) the sorting is 'asc null first' and 'desc nulls last'.` But if I am understanding this correct now the default (=False) is `asc nulls first` and `desc nulls first`

thijs-obj: There is also a paragraph on the default ordering. That says something is not supported by BigQuery, but I think you fixed that part now

bach/bach/dataframe.py
```diff
@@ -1825,12 +1828,15 @@ def rolling(self, window: int,
             end_boundary = WindowFrameBoundary.FOLLOWING
 
         index = list(self._group_by.index.values()) if self._group_by else []
-        group_by = Window(group_by_columns=index,
-                          order_by=self._order_by,
-                          mode=mode,
-                          start_boundary=start_boundary, start_value=start_value,
-                          end_boundary=end_boundary, end_value=end_value,
-                          min_values=min_periods)
+        group_by = Window(
```
thijs-obj, line 1831: `rolling` and `expanding` are the ways to use Windows according to our docstring (at least the docstring of `window()` tells us to use these two functions). Both of these functions don't set `nulls_last`, so this makes `nulls_last = False`.

**Question:** Does this PR change the behaviour on both Postgres and BigQuery for `rolling` and `expanding`?

I sort of get the feeling it does, but there are no tests for those directly.

thijs-obj (commented):
Have some comments on things that I think are not right, and a question.

bach/bach/dataframe.py
thijs-obj: Part of the issue is that `class Window`, and `df.window()` have no documentation on what the parameters do. So I am not really sure what the `nulls_last` was supposed to do if set to `False`. To be clear: this is NOT a criticism of this PR, that was already the case, but it is something that we were lax on reviewing in the past it seems.

jansentom: Good catch. Think it makes sense to document what nulls_last does if False, as it might not be natural for everyone that it means `nulls first` in this case. It makes sense to me to do this for every method that has it, or at least refer to window() where we properly explain what it means.

I don't think the change in default behaviour affects anything in MH, else we would have caught it in the big BQ review.

bach/bach/partitioning.py
```diff
@@ -523,7 +532,17 @@ def get_order_by_expression(order_by: List['SortColumn'], nulls_last: bool = Fal
         else:
             multi_lvl_exprs = [sc.expression]
         for level_expr in multi_lvl_exprs:
-            expr = Expression.construct('{} {}', level_expr, asc_expr if sc.asc else desc_expr)
+            expr = Expression.construct('{} {}', level_expr,  asc_expr if sc.asc else desc_expr)
+            if is_bigquery(dialect):
+                # Big Query does not support NULLS LAST and NULLS FIRST (might generate some errors)
+                # in window order by, therefore we require to simulate it.
```
jansentom, line 538: complex stuff, but very nicely written.

jansentom:
Obviously, a lot of work went into this one. Nice that you kept at it and found a solution, thanks.

bach/bach/partitioning.py
```diff
@@ -346,6 +347,9 @@ def __init__(self,
         if mode is None:
```
thijs-obj, line 347: I wanted to add a code-suggestion above, but github doesn't allow that on non-changed lines, so here:
```python

        :param na_position: Either 'first' or 'last'. When ordering values, put null values before all other
            values if 'first', and put null values after all other values if 'last'.
```

Let's start adding comments for new parameters at least.

thijs-obj (commented):
One comment. Otherwise looks good to me

bach/bach/partitioning.py
KathiaBarahona: done here https://github.com/objectiv/objectiv-analytics/pull/1089/commits/a7a9762bd3f5c51de62bd0c245628eecfe19247c

PR #1091, opened 2022-08-01
Angular 9.1.13 compatibility

sdirosa:
We cannot publish Angular 9 compatible packages with neither yarn workspace nor changesets because they don't support publishing from the /dist folder as the old angular packager expects. Or at least I haven't found a way that doesn't entirely compromise our Verdaccio setup.

~~Removed angular from automation, will manage deployments and patching manually for now, until a nicer solution is found.~~

Found a solution: we now map the development package.json to the /dist folder. This produces a package that, although being a bit weird with a single dist subfolder, works properly both for older and newer angular versions.

This is also compatible with Verdaccio registry switching, which is what we wanted to preserve.

Small downside: some warnings during the first build of the packages. Mainly because Angular 9 tries to backup some ivy_ngc_bak files, but it doesn't affect the build.


sdirosa:
Tested this with the help of a user, both the TS schema refactoring and these changes solved the issue on NPM 6 + node 10 + Angular 9.1.13.

PR #1092, opened 2022-08-02
Cleanup test data functions in test_data_and_utils.py

thijs-obj:
Bit of cleanup before introducing tests for Athena

Changes:
* Consistently use `DataFrame.from_pandas()` to create dataframes with test data
* No longer rely on tables existing in database for BigQuery (through `_get_big_query_data()`)

This does mean that we now have no tests anymore that query any actual data from a table. This was already the case for Postgres, but now also for BigQuery. Virtually all tests use `from_pandas()` that creates a CTE with the data. So we probably should add back at least one test that actually creates a table and queries the data, added a TODO for that and an [issue](https://github.com/objectiv/objectiv-analytics/issues/1093). This doesn't have the highest prio, as we do have such a test for modelhub so the behaviour is tested there.

### DIFF
--- a/bach/README.md
+++ b/bach/README.md
@@ -43,46 +43,7 @@ make tests
 ```
 
 ### Running BigQuery-only tests
-Before running tests for BigQuery, please make sure you have the following tables in your dataset:
-
-**Cities**
-```sql
-insert into `<YOUR_PROJECT>.<YOUR_DATASET>.cities`(skating_order, city, municipality, inhabitants, founding)
-values
-    (1, 'Ljouwert', 'Leeuwarden', 93485, 1285),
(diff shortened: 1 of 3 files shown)


KathiaBarahona (approved):
lgtm :rocket: nice improvement, but I think we also should remove this section https://github.com/objectiv/objectiv-analytics/blob/main/bach/README.md#running-bigquery-only-tests

thijs-obj:
> lgtm rocket nice improvement, but I think we also should remove this section https://github.com/objectiv/objectiv-analytics/blob/main/bach/README.md#running-bigquery-only-tests

Good point, fixed in [7bf5d7d](https://github.com/objectiv/objectiv-analytics/pull/1092/commits/7bf5d7d6faabb80d6177ff8e26c7bb34ea76b0fc)

PR #1098, opened 2022-08-02
Flatten contexts in SP/BQ

borft:
In this PR:
- restructure the mapping of Objectiv events on the Snowplow event
- instead of one big custom context, split up the event
- alle global contexts get their own specific context and iglu schema
- the location_stack is stored as 1 entity, with its own schema as well:
  - to guarantee / preserve the order
  - to support additional properties

In order for this to work, the new contexts should be added to the iglu server. After that, the snowplow-bigquery-mutator should automagically add columns for any new contexts encountered in an event.

The setup has been tested with the postgres-pubsub-loader. bigquery-pubsub-loader  and the s3-kinesis-loader.  

**NOTE** this is a breaking change for the Snowplow pipeline, as this will change the format in which events are persisted (and thus break any query that relies on the original objectiv taxonomy custom context).


backend/objectiv_backend/snowplow/snowplow_helper.py
```diff
@@ -26,35 +26,58 @@
...
+
+    # # original format
+    # schema = config.schema_objectiv_taxonomy
+    # _event = {'event_id' if k == 'id' else k: v for k, v in event.items()}
+    # try:
+    #     cookie_context = get_context(event, 'CookieIdContext')
+    # except ValueError:
+    #     cookie_context = {}
+    # _event['cookie_id'] = cookie_context.get('id', '')
+    # self_describing_contexts.append(make_snowplow_context(schema, _event))
```
borft, line 66: Uncomment these lines to also send the original event/structure

backend/objectiv_backend/end_points/collector.py
```diff
@@ -164,7 +164,14 @@ def set_time_in_events(events: EventDataList, current_millis: int, client_millis
         # here we correct the tracking time with the calculated offset
         # the assumption here is that transport time should be the same as the server time (current_millis)
         # to account for clients that have an out-of-sync clock
-        event['time'] = event['time'] + offset
```
borft, line 167: this changes the current behaviour, it means event.time is now the exact timestamp as sent by the tracker

backend/tests/collector/test_snowplow.py
```diff
@@ -76,6 +76,26 @@ def test_objectiv_event_to_snowplow_payload():
     assert json.loads(base64.b64decode(body['data'][0]['cx']))
 
 
+def test_objectiv_event_to_snowplow_mapping():
+    collector_payload = objectiv_event_to_snowplow_payload(event=event, config=config)
+
+    body = json.loads(collector_payload.body)
+    data = body['data'][0]
+
+    mapping = {
```
jansentom, line 85: Think some are missing here or are the other time fields populated in the SP pipeline?

jansentom (commented):
Pretty nice to have this. Some details to be fixed, but in general, think this is pretty close to prime-time

backend/objectiv_backend/snowplow/docs/iglu/io.objectiv.context/generate.py
```diff
@@ -0,0 +1,70 @@
...
+
+def get_sp_schema(schema, context_name):
+
+    context = schema['contexts'][context_name]
+
+    required_properties = ['id']
+    properties = {
+        "id": {
+            "type": "string",
+            "description": "some id"
```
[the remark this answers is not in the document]
borft: The descriptions on the web are all by hand (as is the schema), this one is generated. I think it could be nicer, but that requires the script to build the complete hierarchy to infer which of its parents defined the ID.

Actually, come to think  of it. this implementation is broken, because it doesn't include properties that may have been defined in intermediate classes. Implementing that will also fix this issue.

backend/objectiv_backend/snowplow/docs/iglu/io.objectiv/location_stack/jsonschema/1-0-0
```diff
@@ -0,0 +1,21 @@
+{
+    "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+    "description": "Objectiv Taxonomy location_stack schema",
```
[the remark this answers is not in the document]
borft: this was definitely manual. Not sure if we can generate this from the base_schema though.

backend/objectiv_backend/snowplow/snowplow_helper.py
```diff
@@ -83,28 +109,80 @@ def objectiv_event_to_snowplow_payload(event: EventData, config: SnowplowConfig)
...
     payload = {
         "schema": snowplow_payload_data_schema,
         "data": [{
             "e": "se",  # mandatory: event type: structured event
             "p": "web",  # mandatory: platform
-            "tv": "objectiv-tracker-0.0.5",  # mandatory: tracker version
-            "eid": event['id'],  # event_id
-            "url": path_context.get('id', ''),
-            "cx": snowplow_custom_context
+            "tv": "0.0.5",  # mandatory: tracker version
```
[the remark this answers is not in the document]
borft: no, ideally, we'd pull this from the EventList, to reflect the actual tracker version

backend/objectiv_backend/snowplow/snowplow_helper.py
```diff
@@ -83,28 +109,80 @@ def objectiv_event_to_snowplow_payload(event: EventData, config: SnowplowConfig)
     except ValueError:
         path_context = {}
 
+    try:
+        application_context = get_context(event, 'ApplicationContext')
+    except ValueError:
+        application_context = {}
     query_string = urlparse(str(path_context.get('id', ''))).query
```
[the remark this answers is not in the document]
borft: it gets send as part of the CollectorPayload

backend/objectiv_backend/common/config.py
```diff
@@ -50,6 +50,9 @@
 _SP_SCHEMA_COLLECTOR_PAYLOAD = 'iglu:com.snowplowanalytics.snowplow/CollectorPayload/thrift/1-0-0'
 _SP_SCHEMA_CONTEXTS = 'iglu:com.snowplowanalytics.snowplow/contexts/jsonschema/1-0-0'
 _SP_SCHEMA_OBJECTIV_TAXONOMY = 'iglu:io.objectiv/taxonomy/jsonschema/1-0-0'
+_SP_SCHEMA_OBJECTIV_LOCATION_STACK = 'iglu:io.objectiv/location_stack/jsonschema/1-0-0'
```
jansentom, line 53: use schema version var below

backend/tests/collector/test_snowplow.py
borft: only thing missing is `event.id`, added that, other time fields are populated in the SP pipeline.

backend/objectiv_backend/common/config.py
```diff
@@ -189,6 +195,9 @@ def get_config_output_snowplow() -> SnowplowConfig:
     else:
         aws_message_raw_type = 'kinesis'
 
+    schema = get_config_event_schema()
+    version = map_schema_version_to_snowplow(schema.version['base_schema'])
```
borft, line 199: @jansentom version used in iglu is now derived from a mapping method (maps schema['base_version'] onto a semver version), used both by the collector and the iglu generator

backend/objectiv_backend/snowplow/docs/iglu/io.objectiv.context/AbstractLocationContext/jsonschema/1-0-0
```diff
@@ -0,0 +1,21 @@
+{
+    "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+    "description": "AbstractLocationContext are the abstract parents of all Location Contexts.\n        Location Contexts are meant to describe where an event originated from in the visual UI.",
```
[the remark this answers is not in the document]
borft: trimmed whitespacing, and removed newlines.

backend/objectiv_backend/snowplow/docs/iglu/io.objectiv.context/generate.py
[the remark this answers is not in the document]
borft: This is now partially fixed. The description is derived from the base_schema. However, that only has a description for `id` at the top level: `AbstractContext`.

backend/objectiv_backend/snowplow/docs/iglu/io.objectiv.context/generate.py
```diff
@@ -0,0 +1,70 @@
...
+        }
+    }
+
+    if 'properties' in context:
+        for property_name in context['properties']:
+            property_type = ['string']
```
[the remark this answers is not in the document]
borft: sure

backend/objectiv_backend/snowplow/docs/iglu/io.objectiv.context/generate.py
```diff
@@ -0,0 +1,70 @@
+import os
+import json5  # type: ignore
+import json
+
+version = '1-0-0'
```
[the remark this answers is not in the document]
borft: Doesn't work with SemVer in Snowplow. It's now solved with a mapping between the versions, to make sure we keep it in sync.

backend/objectiv_backend/snowplow/docs/iglu/io.objectiv/location_stack/jsonschema/1-0-0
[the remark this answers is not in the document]
borft: this is now generated

backend/objectiv_backend/snowplow/generate_iglu_schema.py
```diff
@@ -0,0 +1,172 @@
...
+        return mapping[version]
+    else:
+        raise Exception(f'Version: {version} not in mapping')
+
+
+def get_properties_for_context(schema: Dict[str, Any], context_type: str, hierarchy: Dict[str, list]) -> \
+        Tuple[Dict[str, dict], List[str]]:
+    """
+    recursively get all properties for a context, by traversing the hierarchy until there are no parents
+    :param schema:
```
jansentom, line 34: well described :)

backend/objectiv_backend/common/config.py
```diff
@@ -11,6 +11,7 @@
 # below (e.g. get_config_output())
 from objectiv_backend.schema.event_schemas import EventSchema, get_event_schema, get_event_list_schema
 from objectiv_backend.common.types import EventListSchema
+from objectiv_backend.snowplow.generate_iglu_schema import map_schema_version_to_snowplow
```
jansentom, line 14: imho, this is the wrong way around. we're in `common` here, that's where the common stuff should be.

backend/objectiv_backend/common/config.py
jansentom: If `_SP_GCP_PROJECT != ''`, we can just not populate these fields. This will allow the collector to start without dieing if SP is not used.

jansentom (approved):

backend/objectiv_backend/common/config.py
borft: Fixed

PR #1097, opened 2022-08-02
Support for running test against AWS Athena

thijs-obj:
Changes:
1. Introduce `@pytest.mark.athena` to mark tests as Athena-compatible
    * In the future when all tests are converted we'll remove the markers again. 
2. Add `--athena` parameter for tests, similar to the `--big-query` parameter
3. Introduce `.env_test` file to set environment variables for connecting to databases
4. Made some Bach changes to make the first test pass: `bach/tests/functional/bach/test_df.py::test_basic`

Running the tests on CI is handled in a follow-up PR: https://github.com/objectiv/objectiv-analytics/pull/1101

PR #1101, opened 2022-08-03
Run bach Athena tests in CI

thijs-obj:
Run Bach Athena tests (1 so far) in CI.

Also added the secrets as repository level secrets in github. I used the account that is in 1password, which I created according to the doc here: https://github.com/objectiv/objectiv-analytics/pull/1097/files#diff-d396beb544e288b0ba60a663e94d78cc8f10496804421e2c9881d0acd8fbe98a

Please have a look @borft @KathiaBarahona

### DIFF
--- a/.github/workflows/bach-tests.yml
+++ b/.github/workflows/bach-tests.yml
@@ -49,6 +49,8 @@ jobs:
       GCP_SERVICE_ACCOUNT: ${{ secrets.GCP_SERVICE_ACCOUNT }}
       # Just the path to a temp file, the actual secret is fully contained in the env var above.
       OBJ_DB_BQ_CREDENTIALS_PATH: './sa.json'
+      OBJ_DB_ATHENA_AWS_ACCESS_KEY_ID: ${{ secrets.OBJ_DB_ATHENA_AWS_ACCESS_KEY_ID }}
+      OBJ_DB_ATHENA_AWS_SECRET_ACCESS_KEY: ${{ secrets.OBJ_DB_ATHENA_AWS_SECRET_ACCESS_KEY }}
     steps:
       - uses: actions/checkout@v2
       - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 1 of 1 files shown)


borft (approved):
lgtm

.github/workflows/bach-tests.yml
```diff
@@ -90,11 +92,25 @@ jobs:
       - name: Install bigquery dependencies
         run: |
           pip install -e bach[bigquery]
-      - name: Unit tests big-query
+      - name: Unit tests BigQuery
```
KathiaBarahona, line 95: or capitalize `postgres` in lines 80 and 84 :P just for following the same pattern
```suggestion
      - name: Unit tests bigquery
```

.github/workflows/bach-tests.yml
```diff
@@ -90,11 +92,25 @@ jobs:
...
+        run: |
+          pip install -e bach[athena]
+      - name: Unit tests Athena
+        run: |
+          cd bach
+          pytest --athena tests/unit
+      - name: Functional tests Athena
+        run: |
+          cd bach
+          pytest -n 4 --dist loadgroup --athena tests/functional
```
KathiaBarahona, line 116: maybe we can run without parallelization and when we progress with the tests support we change it. So, probably an issue for now is enough

KathiaBarahona (approved):
lgtm overall, just couple of comments

.github/workflows/bach-tests.yml
thijs-obj: 04c812b3496f9484445e9f6fca94a167ae0f15f3

.github/workflows/bach-tests.yml
thijs-obj: One worker would be OK for now, but we should get benefit from multiple workers quite soon.

**Question:** What do you mean by "an issue for now is enough"?

PR #1102, opened 2022-08-03
[Checklock] Extend Papermill

KathiaBarahona:
Implemented Papermill API and removed implementation of cell executions and wrapping as papermill also tries to track cell timings and exceptions (which is nice)


modelhub/tests_modelhub/checklock_holmes/checklock_holmes/nb_checker.py
```diff
@@ -1,45 +1,99 @@
...
+        self, cell: nbformat.NotebookNode, cell_index: str = None, **kwargs,
+    ) -> None:
+        super().cell_exception(cell, cell_index, **kwargs)
+
+        exc = kwargs['exception']
+
+        self.nb.metadata.papermill.error = CellError(
+            number=cell.metadata.papermill.index,
+            exc=f'{exc.ename}: {exc.evalue[:self.MAX_LOG_EXCEPTION_MESSAGE]}',
+        )
```
KathiaBarahona, line 35: sadly papermill does not store the exception object

jansentom (commented):
Nice to use papermill for this; I did not know it existed. Tried it out, and it works quite nicely. I did not review in great detail though as discussed, since I'm far from a papermill expert. Think for this stuff it's fine if it "just works" (TM) though. Ship it! :ship:

modelhub/tests_modelhub/checklock_holmes/requirements.txt
```diff
@@ -1,3 +1,5 @@
+papermill
```
jansentom, line 1: The last line of this file does not work anymore. Instead, the Makefile should `pip install -e modelhub` I guess? This will get you bach[bigquery] as well.

KathiaBarahona: Done here https://github.com/objectiv/objectiv-analytics/pull/1102/commits/674f1b05d3d2dcda471e5d6f9b11d05a8debfa00

Issue #1006, opened 2022-07-12
Bach BQ: migrate tests that use pg_engine as fixture


thijs-obj:
* Migrate tests that use the `pg_engine` fixture, to instead use `engine`
* When migration of a test is non-trivial create separate issues for those tests
* Remove `pg_engine` fixture all together afterwards

Similar to https://github.com/objectiv/objectiv-analytics/issues/811, but about other tests

thijs-obj:
Going to close this isseue. All tests now use `engine` instead of `pg_engine`.

Converted tests that are not yet working on all platforms are annotated with a skip_<db name>_todo marker. So we have visibility on what still needs to be done.

Issue #811, opened 2022-06-01
Bach BQ: migrate tests that use get_bt_with_test_data()


thijs-obj:
* Migrate tests that use `get_bt_with_test_data()` to use `get_df_with_test_data()` instead
* When migration of a test is non-trivial create separate issues for those tests

List of sub-issues and related issues:
* https://github.com/objectiv/objectiv-analytics/issues/1024
* https://github.com/objectiv/objectiv-analytics/issues/1025
* https://github.com/objectiv/objectiv-analytics/issues/1026
* https://github.com/objectiv/objectiv-analytics/issues/1027
* https://github.com/objectiv/objectiv-analytics/issues/833
* https://github.com/objectiv/objectiv-analytics/issues/973

thijs-obj:
PR https://github.com/objectiv/objectiv-analytics/pull/1105 replaces usage of `get_bt_with_test_data()` with `get_df_with_test_data()` in conjunction with `pg_engine`. The tests still need to be migrated to actually work with BigQuery tho

thijs-obj:
Going to close this issue: The main effort here has been done. All test tests use `get_df_with_test_data()`.

Converted tests that are not yet working on all platforms are annotated with a `skip_<db name>_todo` marker. So we have visibility on what still needs to be done.

PR #1105, opened 2022-08-03
replace `get_bt_with_test_data()` with `get_df_with_test_data()`

thijs-obj:
Instead of using `get_bt_with_test_data()` we use `get_df_with_test_data(engine=pg_engine)` in combination with the `pg_engine` fixture.

Without these change, the tests run against Postgres when running with `--big-query`, after this change they are skipped when running with `--big-query` (and in the future `--athena`)

This is something I noticed while working on `--athena`, I expected that to run one functional tests, but it ran actually more than that.

### DIFF
--- a/bach/tests/functional/bach/test_savepoints.py
+++ b/bach/tests/functional/bach/test_savepoints.py
@@ -4,16 +4,15 @@ Copyright 2021 Objectiv B.V.
 from typing import List
 
 import pytest
-from sqlalchemy.dialects.postgresql.base import PGDialect
 from sqlalchemy.future import Engine
 
 from bach.savepoints import Savepoints, CreatedObject
 from sql_models.model import Materialization
-from tests.functional.bach.test_data_and_utils import get_bt_with_test_data, assert_equals_data
+from tests.functional.bach.test_data_and_utils import get_df_with_test_data, assert_equals_data
 
 
-def test_add_savepoint():
-    df = get_bt_with_test_data()
+def test_add_savepoint(pg_engine):
+    df = get_df_with_test_data(engine=pg_engine)
     sps = Savepoints()
     sps.add_savepoint('test', df, Materialization.TABLE)
     df2 = df.groupby('municipality').min()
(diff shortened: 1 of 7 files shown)


bach/tests/functional/bach/test_savepoints.py
```diff
@@ -4,16 +4,15 @@
...
 
 from bach.savepoints import Savepoints, CreatedObject
 from sql_models.model import Materialization
-from tests.functional.bach.test_data_and_utils import get_bt_with_test_data, assert_equals_data
+from tests.functional.bach.test_data_and_utils import get_df_with_test_data, assert_equals_data
 
 
-def test_add_savepoint():
-    df = get_bt_with_test_data()
+def test_add_savepoint(pg_engine):
```
KathiaBarahona, line 14: is this test intended just to work with PG? I think I prefer avoid using `pg_engine` fixture and just use the marks and add a comment. As `pg_engine` is something temporary IMHO

KathiaBarahona (approved):
lgtm in overall, but as I mentioned in my prev comment. `engine` fixture + skip mark + comment > `pg_engine`

bach/tests/functional/bach/test_savepoints.py
thijs-obj: I think if we want to support this, we want to support this for all databases, but I am not sure.

Not really the goal of this PR tho, I just wanted to make sure we skip these tests when we run with `--big-query` or `--athena`. There are issue on the board to visit all tests and make them work with BigQuery, as part of those issues we can decide whether we should add a `skip` mark for certain databases. So I don't think it makes sense to consider that now

thijs-obj: (sub) issues are listed in these issues:
* https://github.com/objectiv/objectiv-analytics/issues/811
* https://github.com/objectiv/objectiv-analytics/issues/1006

thijs-obj:
I'll merge this, as there was an approval. Let me know if you want to further discuss the open discussion thread.

PR #1107, opened 2022-08-04
Simplify and fix steps for local development in modelhub readme

thijs-obj:


modelhub/README.md
```diff
@@ -45,17 +45,13 @@ virtualenv venv
 source venv/bin/activate
 export PYTHONPATH=.
 
-# You probably want to remove objectiv-modelhub if you did not just create a fresh venv
-pip uninstall objectiv-modelhub
-
```
thijs-obj, line 50: Removed this as it doesn't really relevant, we just told the user to create a fresh environment

thijs-obj: note that `pip freeze` might show that `objectiv-modelhub` is installed, even in a clean environment.
But that is then because the directory `objectiv_modelhub.egg-info` exists, and that's in the PYTHONPATH.

That directory might exists because a users has invoked `pip install -e .[dev]` before (e.g. in a different virtual environment)

modelhub/README.md
```diff
@@ -45,17 +45,13 @@ virtualenv venv
...
 # This will fail if the postgres lib development headers are not present if so, then on Ubuntu that can
 # be fixed with: sudo apt-get install libpq-dev
 pip install -e .[dev]
 
-# in case your notebook can not find all of the packages you just installed
-# you may also need to create a kernel config for this venv:
-ipython kernel install --user --name=objectiv_venv
```
thijs-obj, line 57: I have never seen an instance where this was needed

modelhub/README.md
```diff
@@ -45,17 +45,13 @@ virtualenv venv
...
-# now restart the notebook server, and from the kernel menu select 'objectiv_venv'
+# The above command installed objectiv-bach from PyPi. For development we need the local copy instead.
+pip uninstall objectiv-bach
+pip install -e ../bach/
```
jansentom, line 54: I can't suggest because gitlab, but we could write:
```
# install both Bach and ModelHub in local 'edit' mode
pip install -e ../bach
pip install -e .[dev]
```
Makes it even simpler

modelhub/README.md
jansentom: discussed offline: can go!

modelhub/README.md
thijs-obj: Changed in https://github.com/objectiv/objectiv-analytics/pull/1107/commits/d5ac969a9accafc9126b818a13def5f680423d90

thijs-obj: Doh, forgot to include the comment. Added that in 9e02ff5aa293b3592ae6f0ebb5aed01dbe76f52f and also moved and reworded the existing comment there

jansentom (approved):

thijs-obj:
Merging, if you have any comments @KathiaBarahona just leave them here and I'll address them in a follow-up PR

PR #1111, opened 2022-08-04
Test setup for athena

thijs-obj:
Test setup for Athena. 

Part 1/3 after splitting https://github.com/objectiv/objectiv-analytics/pull/1097

### DIFF
--- a/bach/tests/conftest.py
+++ b/bach/tests/conftest.py
@@ -43,32 +43,51 @@ import os
 from collections import defaultdict
 from enum import Enum
 from typing import Dict, List
+from urllib.parse import quote_plus
 
 import pytest
 from _pytest.fixtures import SubRequest
 from _pytest.main import Session
 from _pytest.python import Metafunc, Function
 from _pytest.config.argparsing import Parser
+from dotenv import dotenv_values
 from sqlalchemy import create_engine
 from sqlalchemy.engine import Engine
 
+# Load settings from .test_env file, but allow overrides from Environment variables
+_DOT_ENV_FILE = os.path.dirname(os.path.dirname(os.path.realpath(__file__))) + '/.secrets/.test_env'
+_ENV = {
+    **dotenv_values(_DOT_ENV_FILE),
+    **os.environ
+}
 
-DB_PG_TEST_URL = os.environ.get('OBJ_DB_PG_TEST_URL', 'postgresql://objectiv:@localhost:5432/objectiv')
-DB_BQ_TEST_URL = os.environ.get('OBJ_DB_BQ_TEST_URL', 'bigquery://objectiv-snowplow-test-2/bach_test')
-DB_BQ_CREDENTIALS_PATH = os.environ.get(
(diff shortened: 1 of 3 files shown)


bach/tests/conftest.py
```diff
@@ -43,32 +43,51 @@
...
+_DOT_ENV_FILE = os.path.dirname(os.path.dirname(os.path.realpath(__file__))) + '/.secrets/.test_env'
+_ENV = {
+    **dotenv_values(_DOT_ENV_FILE),
+    **os.environ
+}
 
-DB_PG_TEST_URL = os.environ.get('OBJ_DB_PG_TEST_URL', 'postgresql://objectiv:@localhost:5432/objectiv')
-DB_BQ_TEST_URL = os.environ.get('OBJ_DB_BQ_TEST_URL', 'bigquery://objectiv-snowplow-test-2/bach_test')
-DB_BQ_CREDENTIALS_PATH = os.environ.get(
+DB_PG_TEST_URL = _ENV.get('OBJ_DB_PG_TEST_URL', 'postgresql://objectiv:@localhost:5432/objectiv')
```
KathiaBarahona, line 64: is only used in this file
```suggestion
_DB_PG_TEST_URL = _ENV.get('OBJ_DB_PG_TEST_URL', 'postgresql://objectiv:@localhost:5432/objectiv')
```

bach/tests/conftest.py
```diff
@@ -223,4 +253,22 @@ def _get_postgres_engine() -> Engine:
...
 
 def _get_bigquery_engine() -> Engine:
-    return create_engine(DB_BQ_TEST_URL, credentials_path=DB_BQ_CREDENTIALS_PATH)
+    return create_engine(_DB_BQ_TEST_URL, credentials_path=_DB_BQ_CREDENTIALS_PATH)
+
+
+def _get_athena_engine() -> Engine:
+    if _DB_ATHENA_TEST_URL:
+        return create_engine(_DB_ATHENA_TEST_URL)
+    aws_access_key_id = quote_plus(_DB_ATHENA_AWS_ACCESS_KEY_ID)
```
KathiaBarahona, line 262: **question** why do we need `quote_plus`?

KathiaBarahona (approved):
lgtm overall

bach/tests/conftest.py
thijs-obj: As to why `quote_plus()` is needed: Some of these values might contain special characters. e.g. the secret access key for another account we have contains things like `/`, which breaks the url.

As to why not run `quote_plus()` on the final url: that would quote all the special character that do need to be in the url. e.g. the `://` would become `%3A%2F%2F` and the `&` between variables would become '%26'.

bach/tests/conftest.py
thijs-obj: Done here: [cedc294](https://github.com/objectiv/objectiv-analytics/pull/1111/commits/cedc294b6ecdd8f02791d5fb2a02f8ef625a292d)

PR #1113, opened 2022-08-04
Bach athena first test after split

thijs-obj:
Make first test work on Athena


bach/Makefile
```diff
@@ -6,9 +6,18 @@ tests:
 	pytest tests/unit tests/functional tests/
 
 
+tests-athena:
+# Similar to `tests`, but tests that can run against multiple databases will be run against Athena instead
+# of Postgres. Note that a lot of test are not yet supported for athena. The postgres-only tests will still run
```
KathiaBarahona, line 11:
```suggestion
# of Postgres. Note that a lot of tests are not yet supported for Athena. The postgres-only tests will still run
```

bach/bach/utils.py
```diff
@@ -68,6 +68,12 @@ def is_valid_column_name(dialect: Dialect, name: str) -> bool:
         # could lead to identifier collisions, so we just disallow it.
         # source: https://www.postgresql.org/docs/14/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS
         return len(name) < 64
+    if is_athena(dialect):
```
KathiaBarahona, line 71: Test case is missing in https://github.com/objectiv/objectiv-analytics/blob/80a1414cefd5507e39efdc501debbd26a2dee09c/bach/tests/unit/bach/test_utils.py#L24

KathiaBarahona (approved):
lgtm overall

PR #1116, opened 2022-08-04
Bach Support boolean, integer, and float for Athena

thijs-obj:
Changes:
- support boolean, int64, float64 for athena
- mark tests for those series as supporting Athena

### DIFF
--- a/bach/bach/series/series_numeric.py
+++ b/bach/bach/series/series_numeric.py
@@ -1,6 +1,7 @@
 """
 Copyright 2021 Objectiv B.V.
 """
+import math
 from abc import ABC
 from typing import cast, Union, TYPE_CHECKING, Optional, List, Tuple
 
@@ -55,6 +56,10 @@ class SeriesAbstractNumeric(Series, ABC):
 
         :param decimals: The amount of decimals to round to
         """
+        if is_athena(self.engine):
+            return self.copy_override(
+                expression=Expression.construct(f'round({{}}, {decimals})', self)
+            )
         return self.copy_override(
             expression=Expression.construct(f'round(cast({{}} as numeric), {decimals})', self)
         )
@@ -295,6 +300,13 @@ class SeriesInt64(SeriesAbstractNumeric):
             # Postgres expects the argument to a bitshift to be a regular int, not bigint.
             return self._arithmetic_operation(other, 'rshift', '({}) >> cast({} as int)',
                                               other_dtypes=tuple(['int64']))
+        if is_athena(self.engine):
+            # Latest version of prestodb supports bitwise_logical_shift_right() [1], but the version that
+            # Athena uses, does not [2]. So we do some hackish thing here
+            # [1] https://prestodb.io/docs/current/functions/bitwise.html
+            # [2] https://prestodb.io/docs/0.217/functions/bitwise.html
+            return self._arithmetic_operation(other, 'rshift', 'cast(({}) / power(2, {}) as bigint)',
+                                              other_dtypes=tuple(['int64']))
         return self._arithmetic_operation(other, 'rshift', '({}) >> ({})', other_dtypes=tuple(['int64']))
 
     def __lshift__(self, other):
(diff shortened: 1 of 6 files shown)


bach/tests/functional/bach/test_series_numeric.py
```diff
@@ -366,6 +373,7 @@ def test_series_minmax_scale(engine) -> None:
     )
 
 
+@pytest.mark.athena_supported()
```
jansentom, line 376: This one fails locally after I've installed Checklock requirements. :thinking: 
```FAILED tests/functional/bach/test_series_numeric.py::test_exp[engine0] - AssertionError: row 0 is not equal: [1, 2.718281828459045] != [1, 2.7182818284590455]```
Note that there is an extra 5 in the end. 

Work totally fine in a fresh venv without it. Did not investigate yet. @KathiaBarahona any idea?

**Review by jansentom — APPROVED** [review]
LGTM, works locally as well, with exception mentioned. I don't think that needs to be addressed here.

File: bach/tests/functional/bach/test_series_numeric.py (hunk 1 above)
Comment (jansentom) [reply bach/tests/functional/bach/test_series_numeric.py:376]: Nevermind, it's in `-e modelhub`'s requirements, and it only happens very once in a while. Hmm. Sometimes the extra precision is returned. Any ideas?

Comment (thijs-obj) [reply bach/tests/functional/bach/test_series_numeric.py:376]: interesting. I can reproduce this with your bash one liner. I don't yet understand quite what is going on. I am leaning towards blaming Athena tho :) Will look at this a bit more

File: bach/bach/series/series_numeric.py [hunk 2]
```diff
@@ -369,6 +387,16 @@ def supported_value_to_literal(
...
+        if is_athena(dialect) and not math.isfinite(value):
+            # see https://prestodb.io/docs/0.217/functions/math.html, under 'Floating Point functions'
+            if math.isnan(value):
+                return Expression.raw('nan()')
+            if math.isinf(value):
+                if value > 0:
+                    return Expression.raw('infinity()')
+                else:
+                    return Expression.raw('-infinity()')
+            raise Exception('value is not finite, but not nan or infinite')  # should never happen
```
KathiaBarahona, line 399: is it necessary to have this statement then?

KathiaBarahona (approved):
lgtm overall

bach/bach/series/series_numeric.py
thijs-obj: When I wrote this I was only like 99% sure what `math.isfinite` did, so figured I'd add an exception here in case I misunderstood or forgot to handle a case.

This line should be unreachable, so I can remove it if needed.

bach/tests/functional/bach/test_series_numeric.py
thijs-obj: As a fix I added rounding to the `assert_equals_data`. I think this behaviour is acceptable, as the difference are so small. `2.718281828459045` seems to be more correct, as `e` is `2.7182818284590452...` according to [Wikipedia](https://en.wikipedia.org/wiki/E_(mathematical_constant)). 



The two different outputs also show up when using the webinterface for Athena, and only ever show up for `exp(1)` never for any of the other exponentials.

I _think_ the difference is between Athena nodes that run on `x86` and on `ARM` architecture. Only thing I could find on the internet is [this](https://bytemeta.vip/repo/kframework/k/issues/2312) where somebody complains about different output for Java's `Math.exp(1.0)` on arm and x86, and Presto uses Java. Might actually have the same difference if we were to run Postgres on ARM, don't know.

bach/bach/series/series_numeric.py
KathiaBarahona: yeh, let's remove it.

thijs-obj: Read this too late
fixed here: https://github.com/objectiv/objectiv-analytics/pull/1123

bach/tests/functional/bach/test_series_numeric.py
jansentom: Agree, good enough work-around.

PR #1117, opened 2022-08-04
SeriesUUID support for Athena

thijs-obj:
Support 'uuid' type on Athena

PR #1122, opened 2022-08-08
Bach athena UUID

thijs-obj:


thijs-obj:
Reviewed here: https://github.com/objectiv/objectiv-analytics/pull/1117

PR #1124, opened 2022-08-09
Bach: SeriesString.str.upper and SeriesString.str.lower

KathiaBarahona:
Support for both `upper` and `lower` for string values.


bach/tests/functional/bach/test_series_string.py
```diff
@@ -134,6 +135,30 @@ def test_string_replace(engine) -> None:
     assert_equals_data(result_df, expected_columns=expected_columns, expected_data=expected_data)
 
 
+@pytest.mark.athena_supported
```
thijs-obj, line 138: nice :+1:

thijs-obj (approved):
LGTM

PR #1126, opened 2022-08-09
Add start_from_end parameter in navigation paths

vard-obj:


modelhub/modelhub/models/funnel_discovery.py
```diff
@@ -240,11 +241,21 @@ def get_navigation_paths(
             conversion location.
         :param n_examples: limit the amount of navigation paths.
                            if None - all the navigation paths are taken.
+        :param reverse: get the reversed navigation paths.
+                Having, `location_stack = ['a', 'b', 'c' , 'd']` and `steps` = 3
+                Will generate the following paths:
+                - `'b', 'c', 'd'`
+                - `'a', 'b', 'c'`
+                - `None, 'a', 'b'`
```
KathiaBarahona, line 249: I'm a bit confused, wouldn't `d` be the first step?

modelhub/modelhub/models/funnel_discovery.py
```diff
@@ -305,7 +321,10 @@ def get_navigation_paths(
             if step == 1:
                 next_step = root_step_series.copy_override(group_by=None)
             else:
-                next_step = root_step_series.window_lag(offset=step - 1)
+                if reverse:
+                    next_step = root_step_series.window_lead(offset=1 - step)
+                else:
+                    next_step = root_step_series.window_lag(offset=step - 1)
```
KathiaBarahona, line 327: Comment in line 311 makes no sense now:
``# since we are using lag, we should reverse the step order``.

Isn't it easier to just reverse the used in line 312?

vard-obj: done

modelhub/modelhub/models/funnel_discovery.py
```diff
@@ -238,13 +239,28 @@ def get_navigation_paths(
             per each navigation path and adds it as a column to the returned dataframe.
         :param only_converted_paths: if True filters each navigation path to first
             conversion location.
+        :param start_from_end: if True starts the construction of navigation paths from the end.
```
KathiaBarahona, line 242: maybe?
```suggestion
        :param start_from_end: if True starts the construction of navigation paths from the last context from the stack, otherwise it starts from the first.
```

modelhub/modelhub/models/funnel_discovery.py
```diff
@@ -336,6 +356,16 @@ def get_navigation_paths(
...
 
+        if start_from_end:
+            # need to reverse column order
+            column_old_order = result.data_columns
+            column_new_order = column_old_order[::-1]
+
+            new_columns_name = {}
+            for old, new in zip(column_old_order, column_new_order):
+                new_columns_name[old] = new
+            result = result.rename(columns=new_columns_name)[column_old_order]
```
KathiaBarahona, line 367: maybe a bit more comments of why we need to rename and reorder columns will be nice

KathiaBarahona (approved):
lgtm overall

PR #1130, opened 2022-08-09
Bach: Athena base Series `setitem` test

thijs-obj:
Changes:
1. Add the Athena database type to Series docstring. Had forgotten this earlier in #1116 and #1117
2. Converted some of the tests om `test_df_setitem.py` to work for Athena too. Specifically converted the tests for the data types that are already working.

This came up while working on SeriesJson support for Athena

### DIFF
--- a/bach/tests/functional/bach/test_data_and_utils.py
+++ b/bach/tests/functional/bach/test_data_and_utils.py
@@ -297,75 +297,38 @@ def assert_postgres_type(
     assert series_type == expected_series_type
 
 
-def assert_postgres_types(
+def assert_db_types(
         df: DataFrame,
         series_expected_db_type: Dict[str, str],
-):
-    """
-    Check that the given series in the DataFrame have the expected data types in Postgres.
-
-    Fails if df.engine is not a Postgres connection.
-
-    :param df: DataFrame object for which to check the database types
-    :param series_expected_db_type: mapping of series-names, to database types.
-    """
-    if not is_postgres(df.engine):
-        raise Exception(f'df.engine is not a postgres connection, but {df.engine.name}')
-    return _assert_db_types(
-        df=df,
-        series_expected_db_type=series_expected_db_type,
-        typeof_function_name='pg_typeof'
-    )
-
-
-def assert_athena_types(
-        df: DataFrame,
-        series_expected_db_type: Dict[str, str],
-):
-    """
-    Check that the given series in the DataFrame have the expected data types in Athena.
-
-    Fails if df.engine is not an Athena connection.
-
-    :param df: DataFrame object for which to check the database types
-    :param series_expected_db_type: mapping of series-names, to database types.
-    """
-    if not is_athena(df.engine):
-        raise Exception(f'df.engine is not an Athena connection, but {df.engine.name}')
-    return _assert_db_types(
-        df=df,
-        series_expected_db_type=series_expected_db_type,
-        typeof_function_name='typeof'
-    )
-
-
-def _assert_db_types(
-        df: DataFrame,
-        series_expected_db_type: Dict[str, str],
-        typeof_function_name: str
 ):
     """
     Check that the given series in the DataFrame have the expected data types in the database.
 
-    This uses df.engine as the connection.
+    ONLY supported for Postgres and Athena. Other database engines will raise an exception.
 
     :param df: DataFrame object for which to check the database types
     :param series_expected_db_type: mapping of series-names, to database types.
     """
+    engine = df.engine
+    if is_postgres(engine):
+        typeof_function_name = 'pg_typeof'
+    elif is_athena(engine):
+        typeof_function_name = 'typeof'
+    else:
+        raise Exception(f'Not supported: {engine.name}')
+
     df_sql = df.view_sql()
     types_sql = ', '.join(
         f'{typeof_function_name}("{series_name}")'for series_name in series_expected_db_type.keys()
     )
     sql = f'with check_type as ({df_sql}) select {types_sql} from check_type limit 1'
-    db_rows = run_query(engine=df.engine, sql=sql)
+    db_rows = run_query(engine=engine, sql=sql)
     db_values = [list(row) for row in db_rows]
     for i, series_name in enumerate(series_expected_db_type.keys()):
         expected_db_type = series_expected_db_type[series_name]
         db_type = db_values[0][i]
         msg = f"expected type {expected_db_type} for {series_name}, found {db_type}"
         assert db_type == expected_db_type, msg
-
-
 
 
 def convert_expected_data_timestamps(dialect: Dialect, data: List[List[Any]]) -> List[List[Any]]:
--- a/bach/tests/functional/bach/test_data_and_utils.py
+++ b/bach/tests/functional/bach/test_data_and_utils.py
@@ -9,7 +9,7 @@ as a test file. This makes pytest rewrite the asserts to give clearer errors.
 import datetime
 import uuid
 from decimal import Decimal
-from typing import List, Union, Type, Any
+from typing import List, Union, Type, Any, Dict
 
 import sqlalchemy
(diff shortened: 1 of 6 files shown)


bach/tests/functional/bach/test_df_setitem.py
```diff
@@ -42,40 +51,64 @@ def check_set_const(engine, constants: List[Any], expected_series: Type[Series],
         ]
     )
 
+    for column_name in column_names:
+        assert isinstance(bt[column_name], expected_series)
+
+    if is_postgres(engine):
+        pg_types = {column_name: expected_pg_db_type for column_name in column_names}
+        assert_postgres_types(bt, pg_types)
```
thijs-obj, line 59: I created a new function here rather than using the existing `assert_postgres_type`. This way we can do it in a similar way for Postgres and Athena, and check all types with one query. In the future we could then probably get rid of `assert_postgres_type` at some point.

Additional change is that we now check the type of the actual Series instance here, rather than the type associated with the database type. That seems more relevant for this test.

Now this function always does two queries, previously it would do `1 + number of columns added`.

jansentom: A thought:
you now double the `if is_postgres` in both this function as well as in assert_postgres_type.
Would it be nice to silently return from the assert if the wrong engine was given, but check the assertion whet the right one is given? That way, this function becomes simpler and there is less duplicate checks.

jansentom (approved):
Nice, LGTM. One small thought.

bach/tests/functional/bach/test_df_setitem.py
thijs-obj: I see your point. Reason I did it like this, is that last time I did something with `assert_postgres_type` I got the exact opposite feedback: https://github.com/objectiv/objectiv-analytics/pull/608#discussion_r843722206

Can't make everybody happy :confused:

jansentom: ha, makes sense. Let's keep it like this then

bach/tests/functional/bach/test_data_and_utils.py
```diff
@@ -297,6 +297,77 @@ def assert_postgres_type(
...
+    if not is_athena(df.engine):
+        raise Exception(f'df.engine is not an Athena connection, but {df.engine.name}')
+    return _assert_db_types(
+        df=df,
+        series_expected_db_type=series_expected_db_type,
+        typeof_function_name='typeof'
+    )
+
+
+def _assert_db_types(
```
KathiaBarahona, line 342: isn't it easier to make this function public and validate if the engine is supported? I find this a bit difficult to follow tbh

bach/tests/functional/bach/test_data_and_utils.py
```diff
@@ -297,6 +297,77 @@ def assert_postgres_type(
     assert series_type == expected_series_type
 
 
+def assert_postgres_types(
```
KathiaBarahona, line 300:
```suggestion
def assert_postgres_db_types(
```

bach/tests/functional/bach/test_data_and_utils.py
```diff
@@ -297,6 +297,77 @@ def assert_postgres_type(
...
+    if not is_postgres(df.engine):
+        raise Exception(f'df.engine is not a postgres connection, but {df.engine.name}')
+    return _assert_db_types(
+        df=df,
+        series_expected_db_type=series_expected_db_type,
+        typeof_function_name='pg_typeof'
+    )
+
+
+def assert_athena_types(
```
KathiaBarahona, line 321:
```suggestion
def assert_athena_db_types(
```

bach/tests/functional/bach/test_data_and_utils.py
```diff
@@ -297,6 +297,77 @@ def assert_postgres_type(
...
+    :param df: DataFrame object for which to check the database types
+    :param series_expected_db_type: mapping of series-names, to database types.
+    """
+    df_sql = df.view_sql()
+    types_sql = ', '.join(
+        f'{typeof_function_name}("{series_name}")'for series_name in series_expected_db_type.keys()
+    )
+    sql = f'with check_type as ({df_sql}) select {types_sql} from check_type limit 1'
+    db_rows = run_query(engine=df.engine, sql=sql)
+    db_values = [list(row) for row in db_rows]
```
KathiaBarahona, line 361: **question**: How about calling `bach.from_database.get_dtypes_from_model` and then getting the `db type` based on the dtype? I think that also validates that the casting is working

KathiaBarahona: Also, might work for BQ?...

bach/tests/functional/bach/test_df_setitem.py
```diff
@@ -42,40 +51,64 @@ def check_set_const(engine, constants: List[Any], expected_series: Type[Series],
...
         np.int64(4),
         5,
         2147483647,
         2147483648
     ]
-    check_set_const(engine, constants, SeriesInt64, 'bigint')
+    check_set_const(engine, constants, SeriesInt64,
+                    expected_pg_db_type='bigint',
+                    expected_athena_db_type='bigint'
+                    )
```
KathiaBarahona, line 77: **thought** I think we should check dbtypes per series for each engine in unit tests, I don't see any test doing that. This ways we could just get the db_type from the series class based on the engine, and remove those `expected_..._db_type` params from that function

KathiaBarahona: since we are already sure there's a test checking that each series class has the correct db types per engine

bach/tests/functional/bach/test_df_setitem.py
KathiaBarahona: https://github.com/objectiv/objectiv-analytics/pull/1130/files#r942236278 might help removing those doubled ifs ;)

bach/tests/functional/bach/test_df_setitem.py
thijs-obj: I might misunderstand what you are saying. But I don't think unittests (i.e. tests that don't query the database) really help here. Such tests would just duplicate in the test what is written in the `Series.supported_db_dtype` field. Querying the actual database makes sure that we actually have the database type that we think we have.

Example: in a very early version of bach we had `SeriesInt64` literals of the form `123`. If we'd do that now then this test would catch that in Postgres that is actually an `int` type, not a `bigint` type. I am not sure how we could catch something like that in a unittest.

KathiaBarahona: the functional test should still break. What I mean is just avoiding adding the expected db type per engine. Yes, the unit test is just duplicating what is already written, but at least we are sure that is what bach should always yield and if we did a random change in code we can also catch it. For example:

Having the following unit test:
```python
    def test_check_integer_supported_db_types():
        assert  SeriesInt64[DBDialect.POSTGRES] == 'bigint'
        assert SeriesInt64[DBDialect.BIGQUERY] == 'INT64'
        assert SeriesInt64[DBDialect.ATHENA] == 'bigint'
```

For me having the test from above, is that bach ALWAYS expects those db types per engine, therefore we can just do:
```python
   check_set_const(engine, constants, SeriesInt64)
```
And that function should just get the supported dbtype from the Series. Else we kinda "allow" ourselves to provide db types that bach actually does not support.

bach/tests/functional/bach/test_data_and_utils.py
thijs-obj: yes. I think that would be easier. Changed in 075af7d6

bach/tests/functional/bach/test_data_and_utils.py
thijs-obj: Removed this function in 075af7d6

KathiaBarahona (approved):
lgtm!

PR #1133, opened 2022-08-09
Simplify types

thijs-obj:
Small simplification of the typing in SeriesJson.
Before we had `TSeriesJson` as return type and parameter type for a lot of functions, which could be either `SeriesJson` or `SeriesJsonPostgres`. But we always only pass and return `SeriesJson`, so we could simplify that a bit. That's what this PR does


thijs-obj:
doh! I forgot about modelhub. Ignore this one

PR #1134, opened 2022-08-09
Bach: Add test for complex type to json conversion on BQ, and implement.

jansentom:
This is needed  to convert complex data types into JSON on BQ, in order to be able to keep them around in the dataframe at query time (like global_context). 

No rush, if it causes conflicts/trouble at the moment, totally okay to postpone as it's quite simple. 

Adds one query to the test suite.


bach/bach/series/series_json.py
```diff
@@ -196,7 +196,10 @@ def dtype_to_expression(cls, dialect: Dialect, source_dtype: str, expression: Ex
         if is_bigquery(dialect):
             if source_dtype in ('json', 'string'):
                 return expression
-            raise ValueError(f'cannot convert {source_dtype} to json')
+            else:
```
thijs-obj, line 199: I sort of have two concerns here, not really major, but still:
1. Now on BigQuery we support casting from, e.g. an integer to json. But on Postgres we do not. So something that works on BQ might not work on PG.
2. Casting a string to json results in us interpreting the string as a json, whereas casting any other type to json results in us creating a json to represent that type. e.g. casting `True` results in json of the form `true`, but casting `'test'` does not result in json of the form `"test"`, but rather broken json, and casting `'{"test": "test"}'` does not result in the json `"{\"test\": \"test\"}"'`, but just `{"test": "test"}`. To illustrate how different this behaviour is: on AWS Athena this would mean that for strings we do `json_parse(value)` and for all other values we do `cast(value as json)`, very different operations.

Now I think number two is very defendable. But would be good to explain that somewhere in the docs, e.g. the docstring of the SeriesJson class.
Number two, we can just accept that, or limit this to list/dict types, or also allow more casts on Postgres.

thijs-obj (commented):
Some concerns, nothing major. Up to you

bach/bach/series/series_json.py
jansentom: Very valid points. I've limited this to list/dict and made a comment on why that is.

Issue #1131, opened 2022-08-09
TrackedInput with type='checkbox' doesn't trigger events


Yankovsky:
```jsx
<TrackedInput
  id='myid'
  type='checkbox'
/>
```
won't fire any events, because the value doesn't change for the checkbox.
Either `TrackedInput` should track `checked` value change, or there should be a separate component `TrackedCheckbox`.
Currently, the problem could be solved by manually setting value property, but I think this should be integrated into the library.

Also, I think it would be a great addition to add a warning if someone tries to use TrackedInput with type=radio or type=checkbox.

Thank you!

sdirosa:
Checkboxes are a bit arbitrary because in some cases it may make sense to track the value property, while in other cases we only care whether the checkbox is selected or not. And sometimes both.

What would be the preferred behavior in your case?

---
FYI:

There's a new version of the SDK that allows to track custom event handlers and stateless inputs : https://github.com/objectiv/objectiv-analytics/pull/1127.

It's available on npm as `0.0.24-next.5`

So you could do something like this:

```tsx
<TrackedInput 
  type={"checkbox"} 
  id={'checkbox-1'} 
  stateless={true} 
  eventHandler={'onChange'}  // or 'onClick'
  trackValue={true} 
  value={'value-1'}
/>
```

This will track values only though. So an event will be triggered when the checkbox is checked or unchecked, and there's no way of distinguishing them currently.

I think a `TrackedInputCheckbox` makes sense. It should probably, automatically, gather the `checked` attribute as InputValueContext.

Would that cover your use-cases?

Yankovsky:
I think it should be either one very smart `TrackedInput` component, that will know how to handle any input types, or a set of separate components (radio, checkbox, date, file, etc.).
Personally, I prefer the latter option.

sdirosa:
@Yankovsky I went for a smart TrackedInputContext: https://github.com/objectiv/objectiv-analytics/pull/1136

Still working on it:
- Verifying the onChange weirdness while testing
- Adding some more validation messages, as you suggested

I plan to release a preview of this probably early afternoon and give it a round of testing

sdirosa:
[React SDK: easier tracking of checked based inputs + select](https://github.com/objectiv/objectiv-analytics/pull/1136) has been merged in the `next` tag. 

A prerelease has been pushed to NPM: https://www.npmjs.com/package/@objectiv/tracker-react/v/0.0.24-next.12

[...]

PR #1136, opened 2022-08-10
React SDK: easier tracking of `checked` based inputs + `select`

sdirosa:
This PR is inspired by and (hopefully) solves https://github.com/objectiv/objectiv-analytics/issues/1131

Many thanks to @Yankovsky for reporting and testing these changes up 🙌🏼 

---
These changes are live on NPM as [version 0.0.24-next.12](https://www.npmjs.com/package/@objectiv/tracker-react/v/0.0.24-next.12)

---

In this PR we add some new options and TrackedElements to support new Elements and improve how we track some existing ones. 

The main aim is to better support tracking `checked` based inputs, such as radio and checkbox. 

### TrackedInputContext (and inherently TrackedInput)
- Support for `defaultChecked` and `defaultValue`
- Support for Array `value`
- Normalization for `checked` tracking to our standard '0' and '1'
- Re-enabled triggering events via `onChange` event for checkboxes and radio (it was a testing fluke)
- Automatically track radio `name` as ContentContext in LocationStack of InputChangeEvents
- Extended type definition to support HTMLSelectElement + test case
- Refactored TrackedInputContext into a factory with multiple independent implementations

### New Components
Leveraging the new capabilities of TrackedInputContext I've improved / added:
- TrackedInputRadio (pimped it a bit)
- TrackedInputCheckbox (new)
- TrackedSelect with single and multiple selection (new)

### Examples

Change the default `eventHandler` from `onBlur` to `onChange`
```tsx 
<TrackedInputContext Component={'input'} type={'date'} id={'birthday'} eventHandler={'onChange'} />
// or
<TrackedInput type={'date'} id={'birthday'} eventHandler={'onChange'} />
```

Default values
```tsx 
<TrackedInput type={'number'} id={'quantity'} min={1} max={5} defaultValue={1} />
<TrackedInputCheckbox id={'check'} value={'abc'} defaultChecked />
```

Checkbox and option tracking
```tsx 
<TrackedInputCheckbox value={'a'} trackValue={true} />
<TrackedInputCheckbox value={'b'} trackValue={true} />
<TrackedInputCheckbox value={'c'} trackValue={true} />

<TrackedInputRadio name={'options'} value={'something'} trackValue={true} />
<TrackedInputRadio name={'options'} value={'something else'} trackValue={true} />
<TrackedInputRadio name={'options'} value={'something sensitive'} trackValue={true} />
````

Select and select multiple
```tsx
<TrackedSelect id={'user'} trackValue={true}>
  <option value={1}>User A</option>
  <option value={2}>User B</option>
  <option value={3}>User C</option>
</TrackedSelect>

<TrackedSelect id={'options'} multiple values={['a', 'c']} trackValue={true}>
  <option value={'a'}>Option A</option>
  <option value={'b'}>Option B</option>
  <option value={'c'}>Option C</option>
</TrackedSelect>
```


tracker/trackers/react/src/trackedElements/TrackedInputCheckbox.tsx
```diff
@@ -0,0 +1,43 @@
...
+export type TrackedInputCheckboxProps = Omit<TrackedInputContextProps, 'id' | 'type' | 'eventHandler'> & {
+  /**
+   * Optional. Defaults to the `value` attribute.
+   */
+  id?: string;
+
+  /**
+   * Optional. Defaults to `onClick`. Valid values: 'onBlur' | 'onClick'.
+   */
+  eventHandler?: 'onBlur' | 'onClick';
```
sdirosa, line 24: TODO: Apparently onChange works fine when testing in an actual browser. I need to investigate why it doesn't work as expected when unit-testing.

Regardless we should add `onChange` here and the same needs to be done for the Radio component.

sdirosa: It's definitely a bug with testing-library `fireEvent.change`. I found an alternative way of testing this and re-enable onChange events for radio and checkbox inputs.

tracker/trackers/react/src/trackedContexts/TrackedInputContext.tsx
```diff
@@ -56,89 +80,153 @@ const isClickEvent = (event: SyntheticEvent<HTMLInputElement>): event is React.M
...
-    ...otherProps
-  } = props;
-  const [previousValue, setPreviousValue] = useState<string>(defaultValue ? defaultValue.toString() : '');
-  const locationStack = useLocationStack();
-
-  let inputId: string | null = id;
-  if (normalizeId) {
-    inputId = makeIdFromString(inputId);
-  }
+export const TrackedInputContext = React.forwardRef<HTMLInputElement | HTMLSelectElement, TrackedInputContextProps>(
```
sdirosa, line 107: This component is becoming too bulky for my taste. 

While the code is rather simple, managing these many types in one place is a pain. Especially the event handlers have no overlap with each other and in some cases I had to cheat out TS by omitting a more specific type. 

I'll consider changing it into a factory and move the logic for fundamentally different elements in their own sub components. I can think of three: text based elements, checkboxes & radios and selects. I may split checkboxes & radios as well if the overlap is not high enough.

I will probably tackle this in a separate PR.

sdirosa: I've done a first round of cleanup. Each custom implementation is now its own component (default text/value, checkbox, radio, select with single selection) and the main TrackedInputContext component is a factory that, by inspecting the given props, picks the most appropriate implementation.

If we don't have a specific implementation, we revert to using the generic one we had before. It should work fairly well with most text based inputs that have a changing `value`.

This will allow us to easily add more custom implementations as we support more input Elements with specific requirements.

PR #1138, opened 2022-08-10
Support SeriesJson on Athena

thijs-obj:
Changes:
* Enable all tests in `test_series_json.py` for Athena
* Implemented functionality in `SeriesJson`
* Added Athena specific subclass of `JsonAccessor`
* Added Athena specific subclass of `ArrayFlattening`


bach/bach/series/series_json.py
```diff
@@ -225,6 +235,9 @@ def _comparator_operation(
         if is_postgres(self.engine):
             other_dtypes = ('json_postgres', 'json', 'string')
             fmt_str = f'cast({{}} as jsonb) {comparator} cast({{}} as jsonb)'
+        elif is_athena(self.engine):
+            other_dtypes = ('json', 'string')
+            fmt_str = f'{{}} {comparator} {{}}'
```
KathiaBarahona, line 240: **question** Is this block on purpose? We do the same for BQ

bach/bach/series/series_json.py
```diff
@@ -141,11 +142,12 @@ class SeriesJson(Series):
     dtype_aliases: Tuple[DtypeOrAlias, ...] = tuple()
     supported_db_dtype = {
         DBDialect.POSTGRES: 'jsonb',
+        DBDialect.ATHENA: 'json',
         # on BigQuery we use STRING for JSONs. Of course SeriesString is the default class for that, so
         # here we set None
         DBDialect.BIGQUERY: None
     }
-    supported_value_types = (dict, list)
+    supported_value_types = (dict, list, str, float, int, bool)
```
KathiaBarahona, line 150: **question** If we support those new types, why not supporting the casting of the `dtypes`?

bach/bach/series/series_json.py
```diff
@@ -736,3 +753,234 @@ def flatten_array(self) -> Tuple['TSeriesJson', 'SeriesInt64']:
...
+    def get_array_slice(self, key: slice) -> 'TSeriesJson':
+        """ See implementation in parent class :class:`JsonAccessor` """
+        if key.step:
+            raise NotImplementedError('slice steps not supported')
+
+        start_expression = self._get_slice_partial_expr(value=key.start, is_start=True)
+        stop_expression = self._get_slice_partial_expr(value=key.stop, is_start=False)
+
+        values_expression = Expression.construct(
+            "slice(cast({} as array(json)), {}, {})",
```
KathiaBarahona, line 775: nice that we don't need to simulate it :rocket:

bach/bach/series/series_json.py
```diff
@@ -736,3 +753,234 @@ def flatten_array(self) -> Tuple['TSeriesJson', 'SeriesInt64']:
...
+                # return index that is guaranteed to be beyond the last item in the array, but no bigger
+                # than a 32-bit integer. Experiments have shown that slice() will give empty results, if
+                # the length is larger than a 32 bit integer.
+                max_int = 2 ** 31 - 1
+                return Expression.construct(f'{max_int}')
+        elif value >= 0:
+            return Expression.construct(f'{value + 1}')
+        else:
+            array_len = self.get_array_length()
+            return Expression.construct(f'({{}} {value} + 1)', array_len)
```
KathiaBarahona, line 819: **question** Maybe making a helper function for both BQ and Athena? The differences are not that big

bach/bach/series/series_json.py
```diff
@@ -736,3 +753,234 @@ def flatten_array(self) -> Tuple['TSeriesJson', 'SeriesInt64']:
...
+        #   find first position of first element from the filtered array in the original json array
+        # 3.2 for not is_start:
+        #   find last position of last element from the filtered array in the original json array
+
+        if not isinstance(filtering_slice, dict):
+            raise TypeError(f'Key should be a dict, actual type: {type(filtering_slice)}')
+        if not all(isinstance(key, str) for key in filtering_slice.keys()):
+            raise TypeError(f'Keys in the key dict, should all be strings')
+        if not all(isinstance(value, str) for value in filtering_slice.values()):
+            raise TypeError(f'Values in the key dict, should all be strings')
```
KathiaBarahona, line 850: We should add these validations for both BQ and PG. IMHO is better to have the same behavior for all engines

bach/bach/series/series_json.py
```diff
@@ -736,3 +753,234 @@ def flatten_array(self) -> Tuple['TSeriesJson', 'SeriesInt64']:
...
+
+        if is_start:
+            # Find the position in the original array, of the first object that matches the filter criteria
+            first_element_pos_expr = Expression.construct(
+                'array_position({}, element_at({}, 1))',
+                array_expr,
+                filter_array_expr
+            )
+            return first_element_pos_expr
+        else:
```
KathiaBarahona, line 883: **nit-pick** this `else` block is not necessary since `if` block has a return statement

thijs-obj:
Added small fix for test that I initially forgot in 2e208b1918caf72bca83f23ad86d60f555a9e91f. Will have a look at the review comments in a bit

bach/tests/functional/bach/test_series_json.py
```diff
@@ -228,10 +231,10 @@ def test_json_getitem_slice_non_happy_mixed_data(engine, dtype):
     # But behaviour of Postgres and BigQuery is different. For now we just accept that's the way it is.
     bt = get_df_with_json_data(engine=engine, dtype=dtype)
     bts = bt.mixed_column.json[1:-1]
-    if is_postgres(engine):
+    if is_postgres(engine) or is_athena(engine):
```
KathiaBarahona, line 234: **nit-pick** Need to update comment in line 231

bach/bach/series/series_json.py
```diff
@@ -736,3 +753,234 @@ def flatten_array(self) -> Tuple['TSeriesJson', 'SeriesInt64']:
...
+        value_expression = Expression.construct(
+            f'regexp_replace({{}}, {{}}, {{}})',
+            str_expression,
+            Expression.raw('\'(^"|"$)\''),
+            Expression.string_value(''),
+        )
+        from bach.series import SeriesString
+        return self._series_object\
+            .copy_override(expression=value_expression)\
+            .copy_override_type(SeriesString)
```
KathiaBarahona, line 952: Did some tests here and seems there are a couple of inconsistencies between all engines:

```
def test_json_get_item_to_pandas_inconsistencies(engine, dtype):
    import pandas as pd
    import json
    pdf = pd.DataFrame(
        {
            'test_column': [
                json.dumps({'a': '"b"', 'b': 1, 'c': [0, 1, 2]}),
                json.dumps({}),
                json.dumps({'c': [3, 4, 5]})
            ]
        }
    )

    df = DataFrame.from_pandas(engine, df=pdf, convert_objects=True)
    df['test_column'] = df['test_column'].astype('json')
    df['example_1'] = df['test_column'].json.get_value('a')
    df['example_2'] = df['test_column'].json.get_value('a', as_str=True)
    df['example_3'] = df['test_column'].json.get_value('b')
    df['example_4'] = df['test_column'].json.get_value('b', as_str=True)
    df['example_5'] = df['test_column'].json.get_value('c')
    df['example_6'] = df['test_column'].json.get_value('c', as_str=True)

    example_cols = [f'example_{i}' for i in range(1, 7)]
    assert_equals_data(
        df[example_cols],
        expected_columns=['_index_0'] + example_cols,
        expected_data=[
            [0, '"b"',  '"b"',    1,  '1', [0, 1, 2], '[0, 1, 2]'],
            [1,  None,   None, None, None,      None,        None],
            [2,  None,   None, None, None, [3, 4, 5], '[3, 4, 5]'],
        ]
    )
```

The test from above passes for PG and failed for both BQ and Athena.

I guess for BQ is kinda expected because of the use of strings instead of jsons. But seems that with Athena is when getting value as str. Tried experimenting a bit with `json_extract_scalar` but does not work with arrays, so returns NULL in that case. Also, when stringifying an array, the final format is different from PG

bach/bach/series/series_json.py
```diff
@@ -736,3 +753,234 @@ def flatten_array(self) -> Tuple['TSeriesJson', 'SeriesInt64']:
...
+
+    def get_array_length(self) -> 'SeriesInt64':
+        """ For documentation, see function :meth:`JsonAccessor.get_array_length()` """
+        from bach.series import SeriesInt64
+        expression = Expression.construct('json_array_length({})', self._series_object)
+        return self._series_object \
+            .copy_override_type(SeriesInt64) \
+            .copy_override(expression=expression)
```
KathiaBarahona, line 960: FYI, for both BQ and Athena, array lengths of "non array" objects, is None. PG will raise an error

bach/bach/series/series_json.py
```diff
@@ -736,3 +753,234 @@ def flatten_array(self) -> Tuple['TSeriesJson', 'SeriesInt64']:
...
+                # the length is larger than a 32 bit integer.
+                max_int = 2 ** 31 - 1
+                return Expression.construct(f'{max_int}')
+        elif value >= 0:
+            return Expression.construct(f'{value + 1}')
+        else:
+            array_len = self.get_array_length()
+            return Expression.construct(f'({{}} {value} + 1)', array_len)
+
+    def _find_in_json_list(self, filtering_slice: Dict[str, str], is_start: bool):
```
KathiaBarahona, line 821: Failing for Athena:

```pytest
def test_json_dict_slicing_failing(engine, dtype):
    import pandas as pd
    import json
    pdf = pd.DataFrame(
        {
            'test_column': [
                json.dumps({'a': '"b"', 'b': 1, 'c': [0, 1, 2]}),
                json.dumps({}),
                json.dumps({'c': [3, {'a': '1'}, 4, 5, 6]})
            ]
        }
    )

    df = DataFrame.from_pandas(engine, df=pdf, convert_objects=True)
    df['test_column'] = df['test_column'].astype('json')
    df['example_1'] = df['test_column'].json.get_value('c')
    df['example_2'] = df['example_1'].json[{'a': '1'} : 3]

    assert_equals_data(
        df[['example_1', 'example_2']],
        expected_columns=['_index_0', 'example_1', 'example_2'],
        expected_data=[
            [0, [0, 1, 2], []],
            [1,  None, []],
            [2, [3, {'a': '1'}, 4, 5, 6], [{'a': '1'}, 4]],
        ],
        use_to_pandas=True,
    )

```

bach/bach/series/series_json.py
KathiaBarahona: **update** forgot to add `use_to_pandas=True`. Results for both BQ and Athena are actually the same for this test, but the result from `example_3` and `example_6` are different from PG.

KathiaBarahona (changes requested):
Went thoroughly each implementation and most are working as expected. Though in some scenarios results differ from engines, also it seems slicing using a dict is not working. Left a couple of tests to show this points

bach/bach/series/series_json.py
thijs-obj: I think the issue here is that part of the sql we generate is this:
```sql 
cast(json_extract(json_parse("test_column"), '$["c"]') as array(map(varchar, json)))
```
And that gives an error as `[3, {'a': '1'}, 4, 5, 6]` cannot be converted to `array(map(varchar, json))`. It doesn't kill the query as it is wrapped in a `try()`, so it just gives `null`.

Might be fixable by casting to just `array(json)`, and then in the filter method cast each item separately to `map(varchar, json)`  :thinking:

thijs-obj: Created this issue to track this: https://github.com/objectiv/objectiv-analytics/issues/1141

bach/bach/series/series_json.py
thijs-obj: Considered that, but as the functions were slightly different it seemed easier to just have two different functions for now. When we have many similar functions in the future we can always merge them

bach/bach/series/series_json.py
thijs-obj: This is not about what we can cast from, but what `supported_value_types()` supports. And since we just do `json.dumps` there, we actually already supported these values. 

Made this change at some point cause I thought I needed to use `supported_value_to_literal`, later changed the part where I used that, but kept this change as it was in fact correct

bach/bach/series/series_json.py
thijs-obj: This was wrong. Changed in e2d7ea83f81c6c6abfeb97895a08d255853718ca to only support comparisons between `json` left and `json` right, and added https://github.com/objectiv/objectiv-analytics/issues/1142 to fix this consistently in the future

bach/bach/series/series_json.py
thijs-obj: Changed in e2d7ea83f81c6c6abfeb97895a08d255853718ca

KathiaBarahona (approved):

PR #1144, opened 2022-08-12
Replace PNGs in the readme for vectors

jansenbob:
Fixes most of the image issues in the stack image related to the use of bitmap icons for the data cloud vendors. All have been replaced by vector shapes. The Bach icon cannot be fixed without a lot of work because it is not replaceable by a vector image.

The issue seems to be a bug on Github's side where PNG's that are included in SVGs have an incorrect path when viewing the file directly

before:
Screenshot 2022-08-11 at 19 50 56

after:
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,8 +39,15 @@ pip install objectiv-modelhub
 * **An open model hub** with pre-built product analytics models & operations.
 * **A modeling library** to create reusable models that run on your full dataset.
 
-![objectiv_stack](https://user-images.githubusercontent.com/82152911/184355849-ee1c59af-068e-48e0-bcd3-b064633d2fa7.svg#gh-light-mode-only "Objectiv Stack")
-![objectiv_stack_dark](https://user-images.githubusercontent.com/82152911/184355794-4b957f62-210c-4a66-bda6-6405ab8f8a3e.svg#gh-dark-mode-only "Objectiv Stack")
+<a href="https://objectiv.io">
+
+<picture>
+  <source media="(prefers-color-scheme: dark)" srcset="https://user-images.githubusercontent.com/82152911/184355794-4b957f62-210c-4a66-bda6-6405ab8f8a3e.svg">
+  <source media="(prefers-color-scheme: light)" srcset="https://user-images.githubusercontent.com/82152911/184355849-ee1c59af-068e-48e0-bcd3-b064633d2fa7.svg">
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -33,8 +33,8 @@ pip install objectiv-modelhub
 * **An open model hub** with pre-built product analytics models & operations.
 * **A modeling library** to create reusable models that run on your full dataset.
 
-![objectiv_stack](https://user-images.githubusercontent.com/82152911/184146312-dca08dfc-33b9-4f3a-8356-2f0ff8563b1d.svg#gh-light-mode-only "Objectiv Stack")
-![objectiv_stack_dark](https://user-images.githubusercontent.com/82152911/184146420-92ac1db7-4b09-476c-abc5-c67d5a970c75.svg#gh-dark-mode-only "Objectiv Stack")
+![objectiv_stack](https://user-images.githubusercontent.com/82152911/184355849-ee1c59af-068e-48e0-bcd3-b064633d2fa7.svg#gh-light-mode-only "Objectiv Stack")
```
vincenthoogsteder, line 36: @jansenbob should we maybe link this to the website like we do for the taxonomy pic as well? Image works now but does not really offer anything new to a user, the site as clickthrough might be cool?

ivarpruijn: Good idea!

ivarpruijn (commented):
Good improvement, agree with @vincenthoogsteder that it would be even better to link to the site instead of the full-size image though.

README.md
jansenbob: Changing the link is a pretty good idea. Not sure if it should link to the site though, but I can't think of a better alternative...

jansenbob: @ivarpruijn I tried adding links to the site in a similar way a link was added to the taxonomy image, but it seems to cancel out the dark/light mode switching, resulting in both images being showed simultaneously:

image

Here's what i tried. Any idea what i'm doing wrong?

``` [objectiv_stack](https://www.objectiv.io/) ```
``` [objectiv_stack_dark](https://www.objectiv.io/) ```

ivarpruijn: @jansenbob should be fixed now, think we can merge

ivarpruijn: See this documentation: https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#specifying-the-theme-an-image-is-shown-to

Issue #1147, opened 2022-08-12
React SDK: TrackedContext Component Injection overhaul


sdirosa:
This issue is on the line between being a refactoring and a bug. 

We have had in backlog for a while to try and improve our TS definitions for TrackedContexts. 

These components follow the Component Injection pattern and offer the ability to drop-in-replace any element with a tracked version of it. 

The idea is that developers shouldn't have to do anything more than:
- Replace the tag or component name with the Tracked*Context counterpart we offer
- Configure the `Component` and `id` properties to allow us to generate a LocationContext automatically
- Provide types for props and ref

In practice what we have now hits a wall in TS capabilities and if the given Component is too complex, types are not generated, which results in autocompletion and validation not working as intended.

As a simplification we narrowed down what these components can support to just JSX tag or a tag-like elements.

This is not ideal as most application rarely use plain HTML elements, instead they already enhance them with more props and logic. The same goes for third party components. Think of routing libraries. We have provided plugins for those so far, but it clearly doesn't scale.

To solve these problems we want to improve 4 aspects:
1. Element creation
2. Ref forwarding
3. TS Generics definition
4. Objectiv props

## Element creation
Out TrackedContext components internally use React.createElement to create the given Component. 

While this seemed necessary for HTML tags, such as 'input', 'div', etc, it's actually not if we leverage JSX to render.

## Ref forwarding
React.forwardRef doesn't support generics. There are a number of solutions (and some are close to TS hacks) but, since we want to achieve something both clean AND flexible, we should probably ditch using forwardRef and opt for a custom prop. E.g. componentRef. 

This approach has been recommended also by Dan Abramov (https://gist.github.com/gaearon/1a018a023347fe1c2476073330cc5509).

After testing all the options (see here for a recap: https://fettblog.eu/typescript-react-generic-forward-refs/), I've to agree it seems the least invasive and it offers the most configurability to developers.

## TS Generics definition
If we can avoid creating components, we should also be able to finally improve our TS definitions for TrackedContext components and made them support generic types. 

That means making TrackedContext components accept types for the props of the given Component and, optionally, for the expected Ref type.

## Objectiv prop
Reusing existing prop names (id, title) can cause issues in the long run for developers writing the own HOCs and wrappers. It's easy to miss that we don't forward those values automatically to the underlying component. And even when that's clear, it's still easy to set the forwarders in the wrong way.

We already have an experimental branch where we moved all Objectiv related props to an `objectiv` prop as an object.

[...]

PR #1149, opened 2022-08-14
React SDK:  TrackedContext Component Injection improvements

sdirosa:
More info on the issue: https://github.com/objectiv/objectiv-analytics/issues/1147

---

This version is available for testing as 0.0.25-next.0 on NPM

---

**In short**. 
Our TrackedContexts are designed to allow developers to inject either JSX Elements or their own components in them, resulting in a new Elements that is a combination of the original given `Component` and our tracking logic.

This PR introduces some changes to:
- Remove collisions between Objectiv's configuration attributes and the original Component's props
- Allow for injecting any JSX element or custom Element/Component
- Have both autocompletion for both the original Component's native props and Objectiv's configuration attributes
- Opt-in validation for both the original Component's native props and Objectiv's configuration attributes, via generics
- Fully typed props and ref support for both JSX Elements and custom Components'
- Auto-detecting native props, like id, title, href, value and others etc to automatically configure Context attributes
- Allow overriding native props with custom ones when auto detection doesn't work as expected or is not possible

Here is an example of the new syntax and what can be done now:
```tsx
type TestComponentProps = {
  id: number;
  abc: string;
};

const TestComponent = forwardRef((props: TestComponentProps, ref: Ref<HTMLDivElement>) => (
  <div ref={ref}>{props.abc}</div>
));

export const TestWrapper = () => {
  const divRef = createRef<HTMLDivElement>();

  return (
    <>
      {/* Render a custom component normally. */}
      <TestComponent id={123} abc={'test'} />

      {/* Input Element. Generic definitions can be omitted. Not really recommended, though. */}
      <TrackedContentContext ref={divRef} objectiv={{ Component: 'input', id: 'test' }} />

      {/* Input Element. Fully typed input and our objectiv prop via generics. */}
      <TrackedContentContext<ComponentProps<'input'>> 
        ref={divRef} 
        objectiv={{
          Component: 'input',
          id: 'test'
        }}
      />

      {/* Native props are still supported (id, title, href) but they will be applied to the DOM now! */}
      <TrackedContentContext<ComponentProps<'input'>> 
        id={'test'} // Previously this would not be applied to the DOM, unless `forwardId` was set to true
        ref={divRef} 
        objectiv={{
          Component: 'input',
        }}
      />

      {/* Inherently these changes apply to TrackedElements as well */}
      <TrackedDiv id={'test'} />
      <TrackedDiv objectiv={{ id: 'test' }} />

      {/* Custom Component with ref forwarding. Also fully typed and validated for both prop sets */}
      <TrackedContentContext<ComponentProps<typeof TestComponent>>
        id={123}
        abc={'asd'}
        ref={divRef}
        objectiv={{
          id: 'test',
          Component: TestComponent,
        }}
      />
    </>
  );
};

```

---
**Breaking changes**

- Core: `makeIdFromString(sourceString:string)`
  - Refactored to `makeId(source: string | number, normalize: boolean = true)`
  - Support for generating identifiers from numbers has been added
  - New `normalize` parameter, default true, to make this helper much easier to integrate with Tracked elements

- Angular: `MakeIdFromString(inputString: string)` PipeTransform
  -  Refactored to `MakeId(input: string | number, normalize: boolean = true)` to match Core changes

- React: TrackedContexts
  - All objectiv specific properties are now part of a new single prop called `objectiv`
  - Native properties are still auto-detected to attempt to automatically pre-fill `objectiv` attributes
  - We don't block `id`, `title` and `href` property anymore, so they behave as normal. This also means that the `forwardId`, `forwardTitle` and `forwardHref` props are gone, as they serve no purpose anymore.
  
  This change doesn't break the tracking, nor the component, but it may introduce failures in automated snapshot-based  tests. 
  Reason being that the properties that were exclusively added for objectiv, and we used to not forward to the underliying element, are now applied normally. 
  
  To fix these issues, all of those properties can simply be moved to the `objectiv` prop. 
  E.g. `<TrackedDiv id={'test'}>`, which used to produce `<div>` and now produces `<div id="test">` can be fixed by changing it to `<TrackedDiv objectiv={{ id: 'test' }}`. 

  Again, this affect only the DOM. Regardless of the change, the tracking still works the same. This is because we read both the native `id` attribute and our own `objectiv.id` prop.

- React Router Plugin  
  - This Plugin already used the new `objectiv` experimental prop. We only updated it to use the latest version. This means that the `objectiv.contextId` has been renamed to simply `objectiv.id`

---

closes https://github.com/objectiv/objectiv-analytics/issues/1147


Yankovsky:
> This also means that the forwardId, forwardTitle and forwardHref props are gone, as they serve no purpose anymore.

👍 
Overall, I think it is a really good change. Looking forward to it!

PR #1150, opened 2022-08-15
Update categories of model hub operations available

ivarpruijn:
- Update references to all the categories of model hub operations.
- Also fix a changed SQL result in an example notebook so the test runs against `main` again.


vincenthoogsteder (approved):
LGTM

PR #1151, opened 2022-08-15
Bach: SeriesAbstractDateTime support more literals in datetime series

KathiaBarahona:
Preparing datetime tests for Athena. Moved `dtype_to_expression` and `supported_value_to_literal` to parent class of all datetime Series. Also, now it validates if a constant string expression casted to date series has the right value (this is to avoid problems with Athena, as it is quite picky when casting strings to date)


bach/bach/series/series_datetime.py
```diff
@@ -561,25 +583,32 @@ class SeriesTimedelta(SeriesAbstractDateTime):
...
-        # https://www.postgresql.org/docs/8.4/datatype-datetime.html#:~:text=interval%20values%20can%20also%20be%20written%20as%20iso%208601%20time%20intervals%2C
-        return Expression.string_value(value_td.isoformat())
+        if isinstance(value, cls.supported_value_types):
+            # pandas.Timedelta checks already that the string has the correct format
+            # round it up to microseconds precision in order to avoid problems with BigQuery
+            # pandas by default uses nanoseconds precision
+            value_td = pandas.Timedelta(value).round(freq='us')
+
+            if value_td is pandas.NaT:
+                return Expression.construct('NULL')
```
jansentom, line 605: this is a bit nasty imho. Shouldn't we just raise?

bach/bach/series/series_datetime.py
```diff
@@ -311,6 +321,12 @@ class SeriesAbstractDateTime(Series, ABC):
 
     On any of the subtypes, you can access date operations through the `dt` accessor.
     """
+
+    _VALID_DATE_TIME_FORMATS = [
```
jansentom, line 325: What do both of these do? Could use some docstring

bach/bach/series/series_datetime.py
```diff
@@ -341,6 +370,78 @@ def _cast_to_date_if_dtype_date(cls, series: 'Series') -> 'Series':
...
+        dt_value: Union[datetime.datetime, datetime.date, datetime.time, None] = None
+
+        if isinstance(value, str):
+            # get the right date format of the string
+            str_format = cls._get_date_format_for_literal_value(value)
+            dt_value = datetime.datetime.strptime(value, str_format)
+
+        if isinstance(value, numpy.datetime64):
+            if numpy.isnat(value):
+                return Expression.raw('NULL')
```
jansentom, line 429: Isn't this really tricky behaviour?  (I know we did this before, but why?). Shouldn't we raise like in line 440?

bach/bach/series/series.py
```diff
@@ -117,6 +117,14 @@ class Series(ABC):
     by :meth:`supported_value_to_literal()`.
     """
 
+    supported_dtypes_to_cast: Tuple[str, ...] = tuple()
```
jansentom, line 120: I'd find it more logical if this was called `supported_source_dtypes`. Docstring can stay the same, but implementations become much more readable/logical imho. WDYT?

jansentom (commented):
Very nice new tests! All in all a nice improvement. I have one big remark (naming) and two things that I think are risky. Good to merge otherwise. Thnx!

bach/bach/series/series_datetime.py
KathiaBarahona: not really, might happen we are reading values from a pandas dataframe.

bach/bach/series/series_datetime.py
KathiaBarahona: sure. I just moved an array of date formats that `SeriesTimestamp` supports for string values, and `_FINAL_DATE_TIME_FORMAT` is pretty much the final format bach uses

bach/bach/series/series.py
KathiaBarahona: No problema for me

PR #1153, opened 2022-08-16
Bach: strict other dtypes in binary ops

### LINKED ISSUE #1142: Bach - Json - Athene - support comparison with string too
**Issues:**`SeriesJson` allows comparisons with other `SeriesJson` objects and `SeriesString` objects, except when running against Athena. Then only comparison between `SeriesJson` and another `SeriesJson` is allowed. See https://github.com/objectiv/objectiv-analytics/pull/1138/files#r942775258

**Tasks:**
* Add tests
* Make comparison between SeriesJson and SeriesString work on Athena.

**Consideration:** An alternative could be to NOT allow comparisons between string and json on other databases. Might make more sense

jansentom:
When doing binary operations, we sometimes need to cast RHS to a specific datatype before doing the operation. This suggest a way to do this, and shows two ways it is helpful.

Fixes #1142 in the process.

### DIFF
--- a/bach/bach/series/series.py
+++ b/bach/bach/series/series.py
@@ -1146,18 +1146,22 @@ class Series(ABC):
         """
         The standard way to perform a binary operation
 
-        :param self: SeriesSubTypehe left hand side expression (lhs) in the operation
+        :param self: SeriesSubType: the left hand side expression (lhs) in the operation
         :param other: The right hand side expression (rhs) in the operation
         :param operation: A user-readable representation of the operation
         :param fmt_str: An Expression.construct format string, accepting lhs and rhs as the only parameters,
             in that order.
-        :param other_dtypes: The acceptable dtypes for the rhs expression
+        :param other_dtypes: The acceptable dtypes for the rhs expression. There will be
+            an explicit cast/astype to lhs; we accept the type as rhs for this operation. If the DB
+            does not know what to do with the types in the operation, explicit type conversions can be done
+            using the `strict_other_dtypes` parameter.
         :param dtype: The new dtype for the Series that results from this operation. Leave None for same
             as lhs, pass a string with the new explicit dtype, or pass a dict that maps rhs.dtype to the
             resulting dtype. If the dict does not contain the rhs.dtype, None is assumed, using the lhs
             dtype.
         :param strict_other_dtypes: list of dtypes that are accepted directly into the rhs, all other ones
-            will be cast to lhs.dtype before the operation. Defaults to `other_dtypes` if none given.
+            (from `other_dtypes`) will be cast to lhs.dtype before the operation. Defaults to
+            `other_dtypes` if none given.
         """
         if len(other_dtypes) == 0:
             raise NotImplementedError(f'binary operation {operation} not supported '
--- a/bach/bach/series/series.py
+++ b/bach/bach/series/series.py
@@ -1140,7 +1140,8 @@ class Series(ABC):
         operation: str,
         fmt_str: str,
         other_dtypes: Tuple[str, ...] = (),
-        dtype: Optional[Union[str, Mapping[str, Optional[str]]]] = None
+        dtype: Optional[Union[str, Mapping[str, Optional[str]]]] = None,
+        strict_other_dtypes: Tuple[str, ...] = ()
     ) -> 'Series':
         """
         The standard way to perform a binary operation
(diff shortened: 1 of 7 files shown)


bach/bach/series/series.py
```diff
@@ -1155,13 +1156,22 @@ def _binary_operation(
             as lhs, pass a string with the new explicit dtype, or pass a dict that maps rhs.dtype to the
             resulting dtype. If the dict does not contain the rhs.dtype, None is assumed, using the lhs
             dtype.
+        :param strict_other_dtypes: list of dtypes that are accepted directly into the rhs, all other ones
```
jansentom, line 1159: This is the main change

bach/bach/series/series.py
```diff
@@ -1155,13 +1156,22 @@ def _binary_operation(
...
             raise NotImplementedError(f'binary operation {operation} not supported '
                                       f'for {self.__class__} and {other.__class__}')
 
+        if len(strict_other_dtypes) == 0:
+            strict_other_dtypes = other_dtypes
+
         other = value_to_series(base=self, value=other)
         self_modified, other = self._get_supported(operation, other_dtypes, other)
+
+        if other.dtype not in strict_other_dtypes:
```
jansentom, line 1172: This is the only true logic that is added generically.

bach/tests/functional/bach/test_series_json.py
```diff
@@ -70,6 +70,23 @@ def test_json_get_single_value(engine, dtype):
     assert a == {'a': 'b', 'c': {'a': 'c'}}
 
 
+def test_json_equals(engine, dtype):
```
jansentom, line 73: This was completly untested, even do we had a weird exception for PG and an issue for Athena. This catches both cases.

bach/bach/series/series_json.py
```diff
@@ -232,26 +232,24 @@ def _comparator_operation(
...
         other: Union['Series', AllSupportedLiteralTypes],
         comparator: str,
-        other_dtypes=tuple()
+        other_dtypes=tuple(),
+        strict_other_dtypes=tuple()
     ) -> 'SeriesBoolean':
         if is_postgres(self.engine):
             other_dtypes = ('json_postgres', 'json', 'string')
-            fmt_str = f'cast({{}} as jsonb) {comparator} cast({{}} as jsonb)'
+            strict_other_dtypes = tuple(['json_postgres'])
```
jansentom, line 240: Works directly through the new strict_other_dtypes change. Test now as well, see below.

bach/bach/series/series_json.py
```diff
@@ -232,26 +232,24 @@ def _comparator_operation(
...
         elif is_athena(self.engine):
-            # TODO: support string here to for compatibility with other databases
-            #  https://github.com/objectiv/objectiv-analytics/issues/1142
-            other_dtypes = ('json')
-            fmt_str = f'{{}} {comparator} {{}}'
+            other_dtypes = ('json', 'string')
+            strict_other_dtypes = tuple(['json'])
```
jansentom, line 243: Pretty happy with this, quite elegant ;)

bach/bach/series/series.py
KathiaBarahona: **question** what does rhs mean?

KathiaBarahona: ahhh nvm, saw line 1150 till now

KathiaBarahona, line 1159: I find the naming a bit confusing :thinking: I also got into this problem, but did you consider a boolean instead? like `cast_other_dtype`, smth like?

KathiaBarahona (commented):
nice! tho, I think the only problem I have is that it's used only in 2 cases. and maybe when defining `other_dtypes`  for binary operation that might implicitly mean that we can cast it to the LHS's dtype. So there might be a bit of redundancy between that parameter and `strict_other_dtypes`

jansentom:
> parameter and `strict_other_dtypes`

jansentom:
> I think the only problem I have is that it's used only in 2 cases. and maybe when defining `other_dtypes` for binary operation that might implicitly mean that we can cast it to the LHS's dtype

True, but the `other_dtypes` leaves it to the DB to handle the conversion, while `strict_other_dtypes` lets Bach do the conversion first. I will add that difference to the docstring.

KathiaBarahona:
> > I think the only problem I have is that it's used only in 2 cases. and maybe when defining `other_dtypes` for binary operation that might implicitly mean that we can cast it to the LHS's dtype
> 
> True, but the `other_dtypes` leaves it to the DB to handle the conversion, while `strict_other_dtypes` lets Bach do the conversion first. I will add that difference to the docstring.

:thinking: thought `other_dtypes` was more for us to know which series dtypes are supported for binary operations. Maybe the docstring might help, because right now it's hard to understand the difference between both params

jansentom:
> > > I think the only problem I have is that it's used only in 2 cases. and maybe when defining `other_dtypes` for binary operation that might implicitly mean that we can cast it to the LHS's dtype
> > 
> > 
> > True, but the `other_dtypes` leaves it to the DB to handle the conversion, while `strict_other_dtypes` lets Bach do the conversion first. I will add that difference to the docstring.
> 
> thinking thought `other_dtypes` was more for us to know which series dtypes are supported for binary operations. Maybe the docstring might help, because right now it's hard to understand the difference between both params

Updated in f07137637fbc43b5d29d56f5873fa9f712938665

KathiaBarahona (approved):

PR #1162, opened 2022-08-19
Modelhub: Refactoring of FunnelDiscovery model

KathiaBarahona:
Since we started to have issues with complex queries, did some optimizations required when generating navigation paths. Improvements:
- Removed unnecessary aggregation of nicenames to array and unflattering. Now, we just calculate the offset of the nice name using `row_number` based on the partition and moment of event.
- calculation of `_first_conversion_step_number` is now performed when generating steps based on nice name, so now we don't extract the step number from the tag and perform the `melt step`. The final expression of this series is based just on a` case when ` expression.

### DIFF
--- a/modelhub/modelhub/models/funnel_discovery.py
+++ b/modelhub/modelhub/models/funnel_discovery.py
@@ -4,12 +4,11 @@ Copyright 2022 Objectiv B.V.
 
 import bach
 from bach.series import Series
+from bach.expression import Expression, join_expressions
 from sql_models.constants import NotSet, not_set
-from typing import cast, List, Union, TYPE_CHECKING, Optional
+from typing import List, Union, TYPE_CHECKING
 
-from sql_models.util import is_bigquery, is_postgres
 
-from modelhub.decorators import use_only_required_objectiv_series
 from modelhub.util import check_groupby
 
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


modelhub/modelhub/models/funnel_discovery.py
```diff
@@ -312,147 +186,218 @@ def get_navigation_paths(
...
+        if only_converted_paths:
+            result = self._filter_navigation_paths_to_conversion(
+                result, step_series_name=final_result_series,
             )
 
-        flattened_lc = flattened_lc.astype('string')
+        if add_conversion_step_column:
+            final_result_series.append(self.CONVERSTION_STEP_COLUMN)
+
+        return result[final_result_series]
```
vard-obj, line 240: shall we add here `.sort_index()` ?
Otherwise, in the notebook we get the following error:
`ValueError: Sorting contains expressions referencing non-existent columns in current base node. ['__root_step_offset']. Please call DataFrame.sort_values([]) or DataFrame.sort_index() for removing sorting and try again.`

KathiaBarahona: afff true :/ bit annoying since it affects tests

KathiaBarahona: fixed here https://github.com/objectiv/objectiv-analytics/pull/1162/commits/bc6caed8b28cd759303b36edb8a5192adaa98f27

PR #1163, opened 2022-08-19
Support anonymous mode in the collector

borft:
adds /anonymous endpoint to collector, if used:
- no cookies are set
- some properties are hashed / anonimized, based on configuration


ivarpruijn:
@borft If I run the Collector following the instructions in https://github.com/objectiv/objectiv-analytics/blob/main/backend/CONTRIBUTING.md it starts on port 5000 instead of 8081. Am I missing an env variable, or should it be set before running `flask run`?

backend/objectiv_backend/common/config.py
```diff
@@ -73,16 +73,20 @@
 # default cookie duration is 1 year, can be overridden by setting `COOKIE_DURATION`
 _OBJ_COOKIE_DURATION = int(os.environ.get('COOKIE_DURATION', 60 * 60 * 24 * 365 * 1))
 # default cookie samesite is Lax, can be overridden by setting `COOKIE_SAMESITE`
-_OBJ_COOKIE_SAMESITE = str(os.environ.get('COOKIE_SAMESITE', 'Lax'))
+_OBJ_COOKIE_SAMESITE = str(os.environ.get('COOKIE_SAMESITE', 'None'))
```
borft, line 76: This is a bonus bugfix, to set proper defaults for the cookie, so it works ootb when collector and tracker are not on the same domain

backend/objectiv_backend/common/event_utils.py
```diff
@@ -34,6 +34,23 @@ def get_contexts(event: EventData, context_type: ContextType) -> List[ContextDat
     return result
 
 
+def remove_global_contexts(event: EventData, context_type: ContextType):
+    """
+    Remove all instances of context_Type from global contexts
```
jansentom, line 39:
```suggestion
    Remove all instances of context_type from global contexts
```

backend/objectiv_backend/common/event_utils.py
```diff
@@ -34,6 +34,23 @@ def get_contexts(event: EventData, context_type: ContextType) -> List[ContextDat
...
+    :param event:  EventData
+    :param context_type: name of context to remove
+    :return: copy of event with context removed
```
jansentom, line 42: this is not true, the original event is modified (booh?!)

backend/objectiv_backend/common/event_utils.py
```diff
@@ -34,6 +34,23 @@ def get_contexts(event: EventData, context_type: ContextType) -> List[ContextDat
...
+    """
+    # new_event = {k: v for k, v in event.items() if k != 'global_contexts'}
+
+    # new_event['global_contexts'] = []
+
```
jansentom, line 47:
```suggestion
```

backend/objectiv_backend/end_points/collector.py
```diff
@@ -41,21 +44,39 @@ def collect() -> Response:
         transport_time: int = event_data['transport_time']
     except ValueError as exc:
         print(f'Data problem: {exc}')  # todo: real error logging
-        return _get_collector_response(error_count=1, event_count=-1, data_error=exc.__str__())
+        return _get_collector_response(error_count=1, event_count=-1, data_error=exc.__str__(),
+                                       anonymous_mode=anonymous_mode)
+
+    # check for SessionContext to get client session id
+    if 'client_session_id' in event_data:
+        client_session_id = event_data['client_session_id'][7:]
```
jansentom, line 52: why the `[7:]`? Could use a comment, and looks quite hacky as well. Can you be more defensive here?

backend/objectiv_backend/end_points/collector.py
```diff
@@ -41,21 +44,39 @@ def collect() -> Response:
...
     # Do all the enrichment steps that can only be done in this phase
-    add_enriched_contexts(events)
+    add_enriched_contexts(events, anonymous_mode=anonymous_mode, client_session_id=client_session_id)
 
     set_time_in_events(events, current_millis, transport_time)
 
+    config = get_collector_config()
+    if anonymous_mode:
+        # in anonymous mode we hash certain properties, as defined in config.anonymous_mode.to_hash
+        for event in events:
```
jansentom, line 64: nice to spit this out into a function?

backend/objectiv_backend/end_points/collector.py
```diff
@@ -41,21 +44,39 @@ def collect() -> Response:
...
         ok_events, nok_events, event_errors = process_events_entry(events=events, current_millis=current_millis)
         print(f'ok_events: {len(ok_events)}, nok_events: {len(nok_events)}')
         write_sync_events(ok_events=ok_events, nok_events=nok_events, event_errors=event_errors)
-        return _get_collector_response(error_count=len(nok_events), event_count=len(events), event_errors=event_errors)
+        return _get_collector_response(error_count=len(nok_events), event_count=len(events), event_errors=event_errors,
+                                       anonymous_mode=anonymous_mode, client_session_id=client_session_id)
     else:
         write_async_events(events=events)
-        return _get_collector_response(error_count=0, event_count=len(events))
+        return _get_collector_response(error_count=0, event_count=len(events),
```
jansentom, line 78: don't we need to pass `anonymous_mode` here?

backend/objectiv_backend/end_points/common.py
```diff
@@ -5,26 +5,53 @@
...
     response = Response(mimetype='application/json', status=status, response=msg)
 
-    cookie_config = get_collector_config().cookie
-    if cookie_config:
-        cookie_id = get_cookie_id()
-        response.set_cookie(key=cookie_config.name, value=f'{cookie_id}',
-                            max_age=cookie_config.duration, samesite=cookie_config.samesite,
-                            secure=cookie_config.secure)
+    # No cookies if in anonymous mode
+    if not anonymous_mode:
```
jansentom, line 22: should we explicitely unset the cookie otherwise?

backend/objectiv_backend/end_points/common.py
```diff
@@ -5,26 +5,53 @@
...
+    if anonymous_mode:
+        cookie_id = client_session_id
+        cookie_source = 'client'
+    else:
+        cookie_id = get_cookie_id_from_cookie(generate_if_not_set=(client_session_id is None))
+        if not cookie_id:
+            cookie_id = client_session_id
+            cookie_source = 'client'
+        else:
+            cookie_source = 'backend'
```
jansentom, line 49: should we use `cookie_config.name`?

backend/objectiv_backend/snowplow/snowplow_helper.py
```diff
@@ -187,7 +194,9 @@ def objectiv_event_to_snowplow_payload(event: EventData, config: SnowplowConfig)
             "ip": http_context.get('remote_address', ''),  # IP address
             "dtm": str(event['corrected_time']),  # Timestamp when event occurred, as recorded by client device
             "stm": str(event['transport_time']),  # Timestamp when event was sent by client device to collector
-            "ttm": str(event['time'])  # User-set exact timestamp
+            "ttm": str(event['time']),  # User-set exact timestamp
+            "nuid": cookie_id,  # Unique identifier for a user, based on a cookie from the collector
+            "sid": client_session_id  # Unique identifier (UUID) for this visit of this user_id to this domain
```
jansentom, line 199: nice one

backend/objectiv_backend/snowplow/snowplow_helper.py
```diff
@@ -204,7 +213,7 @@ def objectiv_event_to_snowplow_payload(event: EventData, config: SnowplowConfig)
         headers=[],
         contentType='application/json',
         hostname='',
-        networkUserId=cookie_context.get('id', '')
+        networkUserId=cookie_id
```
jansentom, line 216: It looks weird that the client session id is nu in here too?

backend/tests/schema/test_schema.py
```diff
@@ -67,6 +68,20 @@ def order_dict(dictionary: Dict[str, Any]) -> Dict[str, Any]:
     return result
 
 
+def test_remove_global_context():
```
jansentom, line 71: Shouldn't there be a test (not necesarily in this file) for 
1) correct client session id promotion to cookie
2) correct obfuscation of PII (easier if split out into seperate method as suggested above)
3) CookieIdContext creation

jansentom (commented):
Did an initial pass, not fully there yet IMHO

backend/objectiv_backend/end_points/collector.py
borft: I think we should rather remove this code path, as it is not used anywhere

backend/objectiv_backend/end_points/common.py
borft: I don't think so, that is the name of the cookie stored on the client, and is not related to the source of the id here.

backend/objectiv_backend/end_points/common.py
borft: I think the only way to remove a cookie, is by setting it to an expired date (eg. return a setcookie header with an expiry date in the past). I'd say that in the situation where a client has a valid cookie, and consent is withdrawn, removing the cookie(s) should be part of that workflow.

backend/objectiv_backend/end_points/collector.py
borft: it is a hack, because we add the string `client-` in front of the ID, this should go when we update the tracker to no longer do that.

backend/objectiv_backend/end_points/collector.py
borft: I'll do that in a separate PR

backend/objectiv_backend/snowplow/snowplow_helper.py
jansentom: hmm, not sure I understand that answer. What I mean is that we explicitely add it to the payload, but the fields that are given to CollectorPayload constructor are a bit arbitrary? Why is for eaxmple networkUserId present, but not domain_sessionid?

backend/objectiv_backend/end_points/collector.py
jansentom: what's the plan for fixing that?

backend/objectiv_backend/common/config.py
```diff
@@ -148,13 +152,25 @@ class TimestampValidationConfig(NamedTuple):
...
     error_reporting: bool
     output: OutputConfig
     event_schema: EventSchema
     event_list_schema: EventListSchema
 
 
+def get_config_anonynous_mode() -> AnonymousModeConfig:
+    return AnonymousModeConfig({
+        'to_hash': {
+            'HttpContext': {
```
borft, line 166: @jansentom by default these are the fields to be hashed on anonymous mode. Maybe we should do referrer as well, any more?

backend/tests/schema/test_schema.py
borft: yes, please check test_collector.py above

backend/objectiv_backend/snowplow/snowplow_helper.py
borft: We explicitly set all parameters to the constructor, there aren't any more. 

I've changed it now, to not set it, if the session id doesn't come from a cookie. What I meant was, that now querying becomes more complicated, as you need to check 2 columns rather than 1.

borft:
> @borft If I run the Collector following the instructions in https://github.com/objectiv/objectiv-analytics/blob/main/backend/CONTRIBUTING.md it starts on port 5000 instead of 8081. Am I missing an env variable, or should it be set before running `flask run`?

@ivarpruijn 
The default port for flask apps is 5000. If you want a different port, you can add `-p <port number>` to the commandline

backend/objectiv_backend/end_points/collector.py
borft: done

backend/objectiv_backend/end_points/collector.py
borft: two-fold:
1. made the code a little more defensive, so it will be able to handle both cases (until we fix-up the tracker)
2. change the tracker behaviour, so we don't need the workaround

PR #1165, opened 2022-08-21
Model hub: Flatten context on `DataFrame` creation

jansentom:
In order to provide a nicer interface to interact with global contexts, we move from one global_context column, to preselected smaller ones. For the user, two things change (BREAKING).

1. You need to specify the contexts you'd like to see at modelhub instantiation time.
```
modelhub = ModelHub(time_aggregation='%Y-%m-%d', global_contexts=['http', 'application', 'marketing'])
```

2. You have to use the new accessor to acces the data:
```
df['app'] = df.application.context.id # use the accessor `context` on the application context, and retrieve `id`
```

Behind the scenes most of the work is accomplished in FlattenedContextPipeline. Furthermore, we've split the dtypes `objectiv_global_contexts` (plural) from `objectiv_global_context`. The latter is supposed to be used by user. The former acts as a container for all global contexts, and is currently only used in the `extracted_contexts` pipeline step. 

The PR will be follow by one that minimally updates all notebooks #1166, and this can not be merged until that one is merged into this branch.

This paves the way to support #1098, and will be followed up by a PR that adds support for the "flat" data format.


modelhub/modelhub/modelhub.py
```diff
@@ -51,15 +51,18 @@ class ModelHub:
...
         """
         Constructor
 
         :param time_aggregation: Time aggregation used for aggregation models.
+        :param global_contexts: The global contexts that should be made available for analysis
         """
 
         self._time_aggregation = time_aggregation
         self._conversion_events = cast(Dict[str, ConversionEventDefinitionType], {})
+        self._global_contexts = global_contexts
```
KathiaBarahona, line 65: shouldn't we validate if the provided global_contexts are supported? Maybe having an enum with all of them? I can imagine some mistyping the context name, executing the query and not understanding why there's no data.

modelhub/modelhub/pipelines/extracted_contexts.py
```diff
@@ -187,6 +189,14 @@ def _process_taxonomy_data(self, df: bach.DataFrame) -> bach.DataFrame:
             taxonomy_col = taxonomy_col.astype(dtype).copy_override(name=key)
             df_cp[key] = taxonomy_col
 
+        gc_series = df_cp['global_contexts'].astype('objectiv_global_contexts')
+        # Extract the requested global contexts
+        for gc in self._global_contexts:
+            context_name = gc.capitalize() + "Context"
+            df_cp[gc] = gc_series.obj.get_contexts(gc).astype('objectiv_global_context')
+
+        df_cp = df_cp.drop(columns=['global_contexts'])
```
KathiaBarahona, line 198: For keeping consistency.
```suggestion
        df_cp = df_cp.drop(columns=[ObjectivSupportedColumns.GLOBAL_CONTEXTS.value])
```

modelhub/modelhub/pipelines/extracted_contexts.py
```diff
@@ -187,6 +189,14 @@ def _process_taxonomy_data(self, df: bach.DataFrame) -> bach.DataFrame:
             taxonomy_col = taxonomy_col.astype(dtype).copy_override(name=key)
             df_cp[key] = taxonomy_col
 
+        gc_series = df_cp['global_contexts'].astype('objectiv_global_contexts')
```
KathiaBarahona, line 192:
```suggestion
        gc_series = df_cp[ObjectivSupportedColumns.GLOBAL_CONTEXTS.value].astype('objectiv_global_contexts')
```

modelhub/modelhub/pipelines/extracted_contexts.py
```diff
@@ -187,6 +189,14 @@ def _process_taxonomy_data(self, df: bach.DataFrame) -> bach.DataFrame:
...
+        # Extract the requested global contexts
+        for gc in self._global_contexts:
+            context_name = gc.capitalize() + "Context"
```
KathiaBarahona, line 195: Not used
```suggestion
```

modelhub/modelhub/pipelines/util.py
```diff
@@ -18,6 +18,7 @@ def get_objectiv_data(
     with_sessionized_data: bool = True,
     identity_resolution: Optional[str] = None,
     anonymize_unidentified_users: bool = True,
+    global_contexts: List[str] = []
```
KathiaBarahona, line 21: :P let's avoid mutable default arguments
```suggestion
    global_contexts: List[str] = None
```

modelhub/modelhub/series/series_objectiv.py
```diff
@@ -70,45 +70,67 @@ def _bigquery_get_from_context_with_type_series(self, type: str, key: str, dtype
...
 
+    def get_contexts(self, name: str) -> 'SeriesGlobalContext':
+        """
+        Returns an array of contexts from an Objectiv stack where `_type` matches the
+        UpperCamelCase context and created from `name` e.g. 'identity' becomes 'IdentityContext'.
+        """
+        upper_camel_name = name.capitalize() + "Context"
+        return self.get_contexts_with_type(upper_camel_name).astype('objectiv_global_context')
+
+    def get_contexts_with_type(self, type: str):
```
KathiaBarahona, line 81: maybe?
```suggestion
    def get_contexts_by_type(self, type: str):
```

modelhub/modelhub/modelhub.py
jansentom: Yeah, we should. We'll get that out of the box when moving to the new SP format, so it's driven by the data availability, instead of by us keeping a list. I prefer that. WDYT?

modelhub/modelhub/pipelines/util.py
jansentom: oh noes. :(

modelhub/modelhub/series/series_objectiv.py
jansentom: I fully agree, but that means I also need to update the other method, which breaks even more external stuff. I'll think about it some more.

modelhub/modelhub/modelhub.py
KathiaBarahona: mm yes, but we still use constant strings for the global contexts when checking if a df is a valid objectiv dataframe :thinking:

modelhub/modelhub/util.py
```diff
@@ -88,6 +89,7 @@ def get_all_columns(cls) -> List[str]:
 
 def get_supported_dtypes_per_objectiv_column(
     with_md_dtypes: bool = False, with_identity_resolution: bool = True,
+    global_contexts: List[str] = []
```
KathiaBarahona, line 92: :P https://defaultlist.readthedocs.io

modelhub/modelhub/pipelines/extracted_contexts.py
```diff
@@ -187,6 +189,14 @@ def _process_taxonomy_data(self, df: bach.DataFrame) -> bach.DataFrame:
...
+            df_cp[gc] = gc_series.obj.get_contexts(gc).astype('objectiv_global_context')
```
KathiaBarahona, line 196: **suggestion** It's better to extract all contexts at once, else per each context we will be iterating over the array

modelhub/modelhub/pipelines/extracted_contexts.py
```diff
@@ -187,6 +189,14 @@ def _process_taxonomy_data(self, df: bach.DataFrame) -> bach.DataFrame:
             taxonomy_col = taxonomy_col.astype(dtype).copy_override(name=key)
             df_cp[key] = taxonomy_col
 
+        gc_series = df_cp['global_contexts'].astype('objectiv_global_contexts')
+        # Extract the requested global contexts
+        for gc in self._global_contexts:
```
KathiaBarahona, line 194: maybe extracting requested global contexts till we remove event duplication?

modelhub/modelhub/pipelines/extracted_contexts.py
jansentom: thnx!

modelhub/modelhub/pipelines/extracted_contexts.py
jansentom: that's a nice idea! Will check if that's feasible.

modelhub/modelhub/pipelines/extracted_contexts.py
jansentom: I can't find a nice way to do that, suggestions welcome :)

modelhub/modelhub/pipelines/extracted_contexts.py
jansentom: It is, but requires some changes that are not trivial. Created issue #1170

modelhub/modelhub/series/series_objectiv.py
jansentom: Thought about it some more. It's only internal, so I'm keeping it like this for now, to match the horrible style of its sibling methods.

jansentom (commented):
Thanks for the review. Fixed all the constants directly from your suggestions and eradicated all the mutable default args in 4717d19f9ccf2a1b257cae7720868131c7311467. Hope you can live with the other answers.

modelhub/modelhub/pipelines/extracted_contexts.py
KathiaBarahona: mmmm maybe generating a dict where each key is the `_type` and the values are an array containing all values for that context? For example:
```SQL
`gc_mapping_nasty` as (
  select 
  `event_id`,
  '{' || ARRAY_TO_STRING( 

    ARRAY(
        SELECT DISTINCT JSON_QUERY(val, '$."_type"') || ' : ' || '[' || ARRAY_TO_STRING(ARRAY_AGG(val) over ( PARTITION BY JSON_QUERY(val, '$."_type"') order by pos), ', ') || ']'
        FROM UNNEST(JSON_QUERY_ARRAY(`global_contexts`)) val with offset as pos
    ), ', '
  ) || '}' as gc_mapping
  from `context_data___586b4ad6f99d35da6097445b203cdb63`
)
select `event_id`, JSON_QUERY(`gc_mapping`, '$."ApplicationContext"') as application, JSON_QUERY(`gc_mapping`, '$."HttpContext"') as http
from `gc_mapping_nasty`
```
A bit dirty, but we already do worse stuff already :joy_cat:

KathiaBarahona (approved):
lgtm overall

modelhub/modelhub/util.py
```diff
@@ -118,6 +124,7 @@ def check_objectiv_dataframe(
     :param df: bach DataFrame to be checked
     :param columns_to_check: list of columns to verify,
         if not provided, all expected objectiv columns will be used instead.
```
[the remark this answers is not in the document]
jansentom: fixed in c375528701d80a6668acb4c7a8433e1964606942

PR #1167, opened 2022-08-22
Update sankey plots in the docs

vard-obj:


hendrik-obj (commented):
graphs are good when plotted with `funnel.plot_sankey_diagram(df_steps, n_top_examples=15)`. In the example on the page `n_top_examples` is 20, while in the notebook it is 15. Maybe good to align.

hendrik-obj (approved):

PR #1173, opened 2022-08-24
ModelHub: Limit retrieved elements from array subquery

KathiaBarahona:
If having multiple contexts from the same type, subquery in `_bigquery_get_from_context_with_type_series` will return first element of the array for each element (generating multiple scalars)


vard-obj (approved):
lgtm

PR #1177, opened 2022-08-25
Model hub: Support for SP contexts for global contexts

jansentom:
After these changes, we support SPs native BQ format for global contexts. The user should not notice any differences, and the format is autmatically picked up on BQ when there is no global contexts column in the schema.


bach/bach/series/series_list.py
```diff
@@ -226,3 +230,19 @@ def len(self) -> 'SeriesInt64':
         return self._series \
             .copy_override_type(SeriesInt64) \
             .copy_override(expression=Expression.construct(expr_str, self._series))
+
+
+class SafeListAccessor(ListAccessor):
```
jansentom, line 235: Still needs a test.

modelhub/modelhub/pipelines/extracted_contexts.py
```diff
@@ -99,17 +100,65 @@ class ExtractedContextsPipeline(BaseDataPipeline):
...
+        self._taxonomy_column: Optional[TaxonomyColumnDefinition] = None
 
         # check if table has all required columns for pipeline
         dtypes = bach.from_database.get_dtypes_from_table(
             engine=self._engine,
             table_name=self._table_name,
         )
+
+        # If the taxonomy column is present, we need to use the "old" format
+        if _get_taxonomy_column_definition(engine).name in dtypes:
```
jansentom, line 115: this is the most important decision point: taxonomy column present? old format. Not present, new format

modelhub/modelhub/pipelines/extracted_contexts.py
```diff
@@ -99,17 +100,65 @@ class ExtractedContextsPipeline(BaseDataPipeline):
...
+                **self.required_context_columns_per_dialect[self._db_dialect],
+            }
+        else:
+            # No taxonomy column, we need to gather the contexts separately.
+            if not is_bigquery(engine):
+                raise Exception('only bigquery supports new data format for now')
+
+            base_dtypes: Dict[str, Any] = {
+                **self.required_context_columns_per_dialect[self._db_dialect],
+                'event_id': bach.SeriesString.dtype,
```
jansentom, line 128: since this shares nothing, found it more intuitive to define the extra fields here.

modelhub/modelhub/pipelines/extracted_contexts.py
```diff
@@ -166,44 +208,68 @@ def _process_taxonomy_data(self, df: bach.DataFrame) -> bach.DataFrame:
...
+
+            for gc, col in self._global_contexts_columns.items():
+                if gc == 'location_stack':
+                    df_cp[gc] = (
+                        df_cp[col]
+                        .elements_safe[0]
+                        .elements['location_stack']
+                    ).astype('objectiv_location_stack')
+                else:
+                    df_cp[gc] = df_cp[col].astype('objectiv_global_context')
```
jansentom, line 270: this is an implicit cast to json, making it a string from now. We lose the name of all the available fields here. This is definitely a todo to keep those around to give better errors to the user.

modelhub/modelhub/pipelines/extracted_contexts.py
```diff
@@ -99,17 +100,65 @@ class ExtractedContextsPipeline(BaseDataPipeline):
...
+                    context_name = "_".join(data_name[1:-1])
+                    if context_name in global_contexts:
+                        base_dtypes[col] = [{}]
+                        global_contexts_columns[context_name] = col
+
+                elif data_name[:2] == ['location', 'stack']:
+                    base_dtypes[col] = [{'location_stack': bach.SeriesString.dtype}]
+                    global_contexts_columns['location_stack'] = col
+
+            self._base_dtypes = base_dtypes
```
jansentom, line 157: We should probably raise here if some taxonomy column was not found.

modelhub/tests_modelhub/unit/modelhub/test_pipelines_extracted_contexts.py
```diff
@@ -3,18 +3,21 @@
...
+    DBParams
 
 
 @pytest.fixture(autouse=True)
 def patch_extracted_contexts_validations(monkeypatch):
     monkeypatch.setattr(
         'modelhub.pipelines.extracted_contexts.bach.from_database.get_dtypes_from_table',
-        lambda *args, **kwargs: {},
+        lambda engine, table_name: {
+            _get_taxonomy_column_definition(engine).name: _get_taxonomy_column_definition(engine).dtype
```
jansentom, line 19: we need a taxonomy column to detect the format used.

jansentom (commented):
@KathiaBarahona Could you have a go at reviewing this please? It's not fully complete, but far enough for a decent look from your side.

bach/bach/series/series_list.py
jansentom, line 235: this still needs a test.

modelhub/modelhub/pipelines/extracted_contexts.py
```diff
@@ -133,28 +182,21 @@ def _get_pipeline_result(self, **kwargs) -> bach.DataFrame:
     @property
     def result_series_dtypes(self) -> Dict[str, str]:
         context_columns = ObjectivSupportedColumns.get_extracted_context_columns()
-        supported_dtypes = get_supported_dtypes_per_objectiv_column(with_identity_resolution=False)
+        supported_dtypes = get_supported_dtypes_per_objectiv_column(
+            with_identity_resolution=False,
+            with_md_dtypes=True
```
jansentom, line 187: We want to return modelhub types immediately, so we can properly tag global context columns as such

bach/bach/series/series_list.py
KathiaBarahona: mmm why not using `SeriesJson` instead? I think the support is much better

modelhub/modelhub/pipelines/extracted_contexts.py
```diff
@@ -26,8 +27,8 @@ class TaxonomyColumnDefinition(NamedTuple):
 _DEFAULT_TAXONOMY_DTYPE_PER_FIELD = {
     '_type': bach.SeriesString.dtype,
     '_types': bach.SeriesJson.dtype,
-    'global_contexts': bach.SeriesJson.dtype,
-    'location_stack': bach.SeriesJson.dtype,
+    'global_contexts': modelhub.series.SeriesGlobalContexts.dtype,
+    'location_stack': modelhub.series.SeriesLocationStack.dtype,
```
KathiaBarahona, line 31: mmm is it necessary to use those dtypes? I actually used `json` on purpose :P didn't like the idea of mixing the pipelines with those series.

modelhub/modelhub/pipelines/extracted_contexts.py
KathiaBarahona, line 270: Same here, I think we should split this function. And maybe even perform it in `_apply_extra_processing` instead

modelhub/tests_modelhub/conftest.py
```diff
@@ -69,29 +70,37 @@ def pytest_generate_tests(metafunc: Metafunc):
...
-    return DBParams(
-        url=DB_BQ_TEST_URL,
-        credentials=DB_BQ_CREDENTIALS_PATH,
-        table_name='events',
-    )
+        format=DBParams.Format.OBJECTIV
+    )]
+
+
+def _get_bigquery_db_params() -> List[DBParams]:
```
KathiaBarahona, line 94: why not having a separate function for each? :P like:
`_get_bigquery_snowplow_db_params`
`_get_bigquery_objectiv_snowplow_db_params`.

Otherwise we will have other engines returning just a 1 single element array

modelhub/modelhub/pipelines/extracted_contexts.py
```diff
@@ -166,44 +208,68 @@ def _process_taxonomy_data(self, df: bach.DataFrame) -> bach.DataFrame:
...
+            df_cp = df_cp.rename(
+                columns={
+                    'se_action': 'event_type',
+                    'se_category': 'stack_event_types',
+                    'network_userid': 'user_id',
+                    'true_tstamp': 'time'
+                }
+            )
+
+            for gc, col in self._global_contexts_columns.items():
```
KathiaBarahona, line 262: `_global_contexts_columns` is a bit confusing since we have another variable called `_global_contexts` which in a nutshell are also names of other columns

modelhub/modelhub/pipelines/extracted_contexts.py
```diff
@@ -99,17 +100,65 @@ class ExtractedContextsPipeline(BaseDataPipeline):
...
+            global_contexts_columns = {}
+
+            for col, dtype in dtypes.items():
+                # match things like contexts_io_objectiv_context_cookie_id_context_1_0_0
+                # "contexts_io_objectiv_" + context_data_name + "_" + ve_rs_ion
+                if not col.startswith('contexts_io_objectiv_'):
+                    continue
+
+                parts = col.split('_')
+                data_name = parts[3:-3]
```
KathiaBarahona, line 144: why not using a regex instead? looks hackish IMHO

modelhub/modelhub/pipelines/extracted_contexts.py
```diff
@@ -99,17 +100,65 @@ class ExtractedContextsPipeline(BaseDataPipeline):
...
+
+                if data_name[0] == 'context':
+                    context_name = "_".join(data_name[1:-1])
+                    if context_name in global_contexts:
+                        base_dtypes[col] = [{}]
+                        global_contexts_columns[context_name] = col
+
+                elif data_name[:2] == ['location', 'stack']:
+                    base_dtypes[col] = [{'location_stack': bach.SeriesString.dtype}]
+                    global_contexts_columns['location_stack'] = col
```
KathiaBarahona, line 155: Isn't it better to treat `location_stack` separately from `global_contexts`?  I find a bit confusing mixing both

bach/bach/series/series_list.py
jansentom: because the native type is much faster and it's the SP format we want to follow

jansentom (commented):
Responded / fixed most of your comment @KathiaBarahona. Can you let me know whether you're happy with it?

modelhub/modelhub/pipelines/extracted_contexts.py
```diff
@@ -99,20 +101,81 @@ class ExtractedContextsPipeline(BaseDataPipeline):
...
 
         # check if table has all required columns for pipeline
         dtypes = bach.from_database.get_dtypes_from_table(
             engine=self._engine,
             table_name=self._table_name,
         )
+
+        # If the taxonomy column is present, we need to use the "old" format
+        if _get_taxonomy_column_definition(engine).name in dtypes:
+            self._taxonomy_column = _get_taxonomy_column_definition(engine)
```
KathiaBarahona, line 117: taxonomy_column = _get_taxonomy_column_definition(engine)
```suggestion
        if taxonomy_column.name in dtypes:
            self._taxonomy_column = taxonomy_column
```

PR #1182, opened 2022-08-26
Avoid building the cache twice for the Collector

borft:


backend/objectiv_backend/app.py
```diff
@@ -10,7 +10,7 @@ def create_app() -> Flask:
 
     # load config - this will raise an error if there are configuration problems, and will cache the
     # result for later calls.
-    init_collector_config()
+    get_collector_config()
```
thijs-obj, line 13: I don't really understand the change.

Basically `get_collector_config()` calls `init_collector_config()` and returns the resulting config. We don't use the return value of `get_collector_config()`, so this doesn't seem to change anything?

borft: init_collector_config() always builds the cache, get_collector_config() only does this when needed. This prevents the init to be run more than once, because the config was already generated prior to the call here (resulting in building the cache twice).

thijs-obj: Aah yes this makes total sense :ok_hand:  I only considered the path where this code is always the first place that calls `get_collector_config()`, but that might of course not always be the case, other code might already have called `get_collector_config()`, e.g. in tests.

thijs-obj: Thanks for explaining

thijs-obj (approved):

PR #1183, opened 2022-08-26
Add BI export how-to to notebooks

vincenthoogsteder:


ivarpruijn (approved):
Fixed the latest SQL issues (fyi @vard-obj, will probably fix your issues as well), so LGTM

PR #1184, opened 2022-08-26
Text fixes for example notebooks

vincenthoogsteder:


ivarpruijn (approved):
LGTM, added the change to the LR example notebook as well.

PR #1186, opened 2022-08-29
Set explicit start and end date for notebooks with doctests

ivarpruijn:
Set explicit start and end date for notebooks, which can be overridden in doctests to keep data results consistent.


vincenthoogsteder (approved):
LGTM

hendrik-obj (approved):

PR #1188, opened 2022-08-29
Make docstring and implementation of get_db_dtype consistent

thijs-obj:
Something I noticed in a review: What `get_db_dtype` does was not consistent with what the docstring said.

Changes:
1. Removed from the `:return:` explanation the part that said: `or None if this Series has no database type for which it is the standard Series for that database`
2. Added clarifcation to docstring 
3. Made implemenation of `SeriesJson` consistent

### DIFF
--- a/bach/bach/series/series.py
+++ b/bach/bach/series/series.py
@@ -383,15 +383,15 @@ class Series(ABC):
         """
         Give the database type used to store values of this Series, for the given database dialect.
 
-        ..note:
-            Multiple Series types might return the same db_dtype. For example on BigQuery
-        `SeriesUuid`, `SeriesJson`, and `SeriesString` will all return 'STRING'. As we use the STRING
-        database type to store all of those value types.
-
         :raises DatabaseNotSupportedException: If the Series subclass doesn't support the database dialect.
-        :return: database type as string, or None if this Series database type is a structural type whose
+        :return: Database type as string, or None if this Series database type is a structural type whose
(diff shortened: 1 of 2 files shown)


bach/bach/series/series.py
```diff
@@ -381,13 +381,16 @@ def get_class_instance(
     @classmethod
     def get_db_dtype(cls, dialect: Dialect) -> Optional[str]:
         """
-        Give the static db_dtype of this Series, for the given database dialect.
+        Give the static db_dtype used to store values of this Series, for the given database dialect.
+
+        Note: Multiple Series types might return the same db_dtype. For example on BigQuery
```
KathiaBarahona, line 386:
```suggestion
        ..note:
            Multiple Series types might return the same db_dtype. For example on BigQuery
```

bach/bach/series/series.py
```diff
@@ -381,13 +381,16 @@ def get_class_instance(
     @classmethod
     def get_db_dtype(cls, dialect: Dialect) -> Optional[str]:
         """
-        Give the static db_dtype of this Series, for the given database dialect.
+        Give the static db_dtype used to store values of this Series, for the given database dialect.
```
KathiaBarahona, line 384: **suggestion** For consistency with the text from the return section.
```suggestion
        Give the database type used to store values of this Series, for the given database dialect.
```

KathiaBarahona (approved):
lgtm overall

PR #1204, opened 2022-09-01
Bach athena date formatting

thijs-obj:
Changes:
* Support `strftime()` for Athena
  * Only support a subset of format codes
* Clearly define which format strings we support on all platforms
* Expand functional test: test all codes that we support on all platforms
* Fix issue on BigQuery where strftime() with time related format codes didn't work correctly `SeriesDate`. Fixed by converting to `SeriesTimestamp`

Missing here: Warnings for unsupported codes. See the follow-up PR https://github.com/objectiv/objectiv-analytics/pull/1205 that makes warnings consistent on all databases.

### DIFF
--- a/bach/bach/series/utils/datetime_formats.py
+++ b/bach/bach/series/utils/datetime_formats.py
@@ -2,9 +2,52 @@
 Copyright 2022 Objectiv B.V.
 """
 import re
+import string
 import warnings
 
-# https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes
+
+CODES_SUPPORTED_IN_ALL_DIALECTS = {
+    # These are the codes that we support for all database dialects.
+
+    # week codes
+    '%a',  # WEEKDAY_ABBREVIATED
+    '%A',  # WEEKDAY_FULL_NAME
+
+    # day codes
+    '%d',  # DAY_OF_MONTH
+    '%j',  # DAY_OF_YEAR
+
+    # month codes
+    '%b',  # MONTH_ABBREVIATED
+    '%B',  # MONTH_FULL_NAME
+    '%m',  # MONTH_NUMBER
+
+    # year codes
+    '%y',  # YEAR_WITHOUT_CENTURY
+    '%Y',  # YEAR_WITH_CENTURY
+
+    # time unit codes
+    '%H',  # HOUR24
+    '%I',  # HOUR12
+    '%M',  # MINUTE
+    '%S',  # SECOND
+
+    # format codes
+    '%F',  # YEAR_MONTH_DAY
+    '%R',  # HOUR_MINUTE
+    '%T',  # HOUR_MINUTE_SECOND
+
+    # special characters - TODO
+    # '%n',  # NEW_LINE
+    # '%t',  # TAB
+    # '%%',  # PERCENT_CHAR
+}
+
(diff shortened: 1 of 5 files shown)


thijs-obj:
Sorry, just pushed another commit. Hope nobody was reviewing yet. Initially forgot some fixes to the docstring, fixed in 9016a098d1d04b47044593ad1fefd93dd851e89b

bach/bach/series/utils/datetime_formats.py
```diff
@@ -2,9 +2,55 @@
 Copyright 2022 Objectiv B.V.
 """
 import re
+import string
 import warnings
 
-# https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes
+
+CODES_SUPPORTED_IN_ALL_DIALECTS = {
```
jansentom, line 9: I'm afraid we should figure out a way to support week numbers.

jansentom (approved):
Very nice PR, thanks. 1 small comment, but maybe big work.

bach/bach/series/utils/datetime_formats.py
thijs-obj: I couldn't find any usage of that. Grepped `modelhub/modelhub` and `notebooks` directories for `%U` `%V` and `%W`.

Questions:
1. Do we really need that?
2. If answer 1 is yes, do we care which one specifically? starting at 00 or at 01, starting on Monday or Sunday?
3. Do we need that now? Or can we implement that at some point in the future?

jansentom: I think it's a pretty common aggregation, so 1. Yes, 2. Select one that's most common / iso? 3. It's okay to make an issue for this I think.

jansentom: Maybe @hendrik-obj has a different opinion, or @vincenthoogsteder ?

vincenthoogsteder: yes we def need week numbers, weekly retention is super common. This is the most standard one that can be a great use in modeling I feel

`

%W | Week number of the year (Monday as the first day of the week) as a zero-padded decimal number. All days in a new year preceding the first Monday are considered to be in week 0. | 00, 01, …, 53 | (7), (9)
-- | -- | -- | --


`

https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes

thijs-obj: https://github.com/objectiv/objectiv-analytics/issues/1215

bach/bach/series/utils/datetime_formats.py
```diff
@@ -174,6 +220,61 @@ def parse_c_standard_code_to_postgres_code(date_format: str) -> str:
     return ''.join(new_date_format_tokens)
 
 
+def parse_c_code_to_athena_code(date_format: str) -> str:
+    """
+    Parses a date format string, and return a string that's compatible with Athena's date_format() function.
+
+    Some python format codes are different on Athena, or might even raise an error. Such codes are converted
+    or escaped; using the returned string with Athena's data_format will never raise an error.
```
KathiaBarahona, line 228:
```suggestion
    or escaped; using the returned string with Athena's date_format will never raise an error.
```

KathiaBarahona (approved):

PR #1205, opened 2022-09-01
Bach athena date formatting warnings

thijs-obj:
Refactor warnings about unsupported format codes in `strftime()`

Details:
* Removed generating warnings in database specific parse functions
* New generic function `warn_non_supported_format_codes`
  * Generate warning if format code is not supported on all databases
* Check warnings in functional test too

This is a follow-up to https://github.com/objectiv/objectiv-analytics/pull/1204

### DIFF
--- a/bach/bach/series/utils/datetime_formats.py
+++ b/bach/bach/series/utils/datetime_formats.py
@@ -43,9 +43,14 @@ CODES_SUPPORTED_IN_ALL_DIALECTS = {
     # '%%',  # PERCENT_CHAR
 }
 
-STRINGS_SUPPORTED_IN_ALL_DIALECTS = {
+STRINGS_SUPPORTED_IN_ALL_DIALECTS = [
+    # These are the combinations of codes that we support for all database dialects, even if some of the
+    # individual codes are not supported.
+    # When adding more strings here: Strings must be sorted longest to shortest, and all strings must
+    # start with a percentage sign.
+
     '%S.%f',  # <seconds>.<microsecnds>
-}
+]
 
 
 _SUPPORTED_C_STANDARD_CODES = {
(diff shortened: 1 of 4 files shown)


thijs-obj:
doh; please hold review. Small buggy

bach/bach/series/utils/datetime_formats.py
```diff
@@ -281,10 +275,50 @@ def parse_c_code_to_bigquery_code(date_format: str) -> str:
     """
     if '%S.%f' in date_format:
         date_format = re.sub(r'%S\.%f', '%E6S', date_format)
+    return date_format
+
+
+def warn_non_supported_format_codes(date_format: str):
```
KathiaBarahona, line 281: Is this actually necessary? IMHO it seems to me it's actually just duplicating `parse_c_code_to_athena_code`. In such a case, I think it's better just to raise the warning on the function in charge of parsing the date format.

thijs-obj: I think we need this, because:
1. `parse_c_code_to_athena_code()` only runs for Athena of course. But I think we need to have the same warnings for all databases.
2. There is this 'weird' exception where we support `%S.%f` but not `%f`. This is because of limitations on BigQuery. Detecting that exception correctly is not trivial, and not something that is build into `parse_c_code_to_athena_code()`

thijs-obj: IMHO it's really important that the interface is the same for all databases. And having one function check whether a format string is 'okay' for all databases, makes it easier to keep the interface consistent for all databases.

bach/bach/series/utils/datetime_formats.py
```diff
@@ -281,10 +275,50 @@ def parse_c_code_to_bigquery_code(date_format: str) -> str:
...
+            # See if the code that starts here is listed in CODES_SUPPORTED_IN_ALL_DIALECTS
+            if i < len(date_format):
+                current = current + date_format[i]
+                i += 1
+            if current not in CODES_SUPPORTED_IN_ALL_DIALECTS:
+                unsupported_c_codes.add(current)
+    if unsupported_c_codes:
+        message = f'These formatting codes are not generally supported: ' \
+                  f'{", ".join(sorted(unsupported_c_codes))}.' \
+                  f'They might not work reliably on some or all database platforms..'
```
jansentom, line 320: Pessimist ;)

jansentom (approved):
I quite like this very defensive approach. LGTM!

PR #1206, opened 2022-09-01
Bach athena date formatting cleanup

thijs-obj:
Changes:
* remove old `sql_format()` function
* remove test string format code `%g` from test
* small tweaks

@hendrik-obj can you double check that I am not cleaning up anything you rely on?

This is part of a series of PRs:
#1204
#1205
#1206 (this one)
#1207

### DIFF
--- a/bach/bach/series/series_datetime.py
+++ b/bach/bach/series/series_datetime.py
@@ -2,7 +2,6 @@
 Copyright 2021 Objectiv B.V.
 """
 import datetime
-import warnings
 from abc import ABC
 from enum import Enum
 from typing import Union, cast, List, Tuple, Optional, Any
--- a/bach/bach/series/series_datetime.py
+++ b/bach/bach/series/series_datetime.py
@@ -60,34 +60,6 @@ class DateTimeOperation:
     def __init__(self, series: 'SeriesAbstractDateTime'):
         self._series = series
 
-    def sql_format(self, format_str: str) -> SeriesString:
-        """
-        Allow formatting of this Series (to a string type).
-
-        :param format_str: The format to apply to the date/time column.
-            Currently, this uses Postgres' data format string syntax:
-            https://www.postgresql.org/docs/14/functions-formatting.html
-
-        .. warning::
-            This method is deprecated, we recommend using :meth:`SeriesAbstractDateTime.dt.strftime` instead.
-
-        .. code-block:: python
-
-            df['year'] = df.some_date_series.dt.sql_format('YYYY')  # return year
-            df['date'] = df.some_date_series.dt.sql_format('YYYYMMDD')  # return date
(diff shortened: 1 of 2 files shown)


bach/bach/series/series_datetime.py
```diff
@@ -60,34 +59,6 @@ class DateTimeOperation:
     def __init__(self, series: 'SeriesAbstractDateTime'):
         self._series = series
 
-    def sql_format(self, format_str: str) -> SeriesString:
```
thijs-obj, line 63: As far as I can tell, we never use this in either modelhub or notebooks. The function was deprecated too, so I think it is safe to remove.

bach/tests/functional/bach/test_series_date.py
```diff
@@ -71,7 +71,6 @@ def test_date_format(engine, recwarn):
         # tuple, types: (str, bool). Content: format, whether the format should raise a warning
         ('Year: %Y', False),
         ('%Y', False),
-        ('%g%Y', True),  # %g is not a supported code
```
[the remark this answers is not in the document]
thijs-obj: Replaced it with `%y%Y`. This way we still test that on Postgres we correctly place double quotes between consecutive codes, but using codes that are supported on all databases.

bach/tests/functional/bach/test_series_date.py
```diff
@@ -261,5 +255,5 @@ def test_date_arithmetic(engine):
 def test_to_pandas(engine):
     bt = get_df_with_test_data(engine)
     bt['d'] = datetime.date(2020, 3, 11)
-    bt[['d']].to_pandas()
-    assert bt[['d']].to_numpy()[0] == [datetime.date(2020, 3, 11)]
+    pdf_result = bt.to_pandas()
+    assert pdf_result[['d']].to_numpy()[0] == [datetime.date(2020, 3, 11)]
```
thijs-obj, line 259: Change so the test only does one query instead of two

bach/tests/functional/bach/test_series_date.py
```diff
@@ -99,18 +99,12 @@ def test_date_format(engine, recwarn):
 
     # Some columns do not contain correct results on some databases. For now, we accept that.
     # Here we define the correct result, below we define exceptions per database
-    iso_week_year_date = '212022'
-    iso_week_year_timestamp = '212021'
     percentage_format_date = '2022-%01-01'
     percentage_format_timestamp = '2021-%05-03'
 
     if is_postgres(engine):
```
thijs-obj, line 105: This lasts exception gets removed in the next PR: #1207

KathiaBarahona (approved):

hendrik-obj (approved):
lgtm

Issue #1194, opened 2022-08-30
Bach[Athena]: Support `df.head` for Athena


KathiaBarahona:
Currently failing due to syntax. 

Current `LIMIT X OFFSET Y`, for athena it must be `OFFSET Y LIMIT X`

thijs-obj:
Fixed by https://github.com/objectiv/objectiv-analytics/pull/1211

Issue #1022, opened 2022-07-14
Bach BQ: fix tests.functional.bach.test_df_getitem.test_positional_slicing


thijs-obj:
# Updated description
When using positional slicing: raise Exception if the DataFrame is not sorted

See `tests.functional.bach.test_df_getitem.test_positional_slicing` on what we mean by positional slicing

# Old description
Below is the old description. Issues 1 and 3 were handled in PR #1213.

Bach's test `tests.functional.bach.test_df_getitem.test_positional_slicing` does currently not work with BigQuery.
The issue is three-fold:
1. test efficiency: too many queries for BQ
 2. implementation: raise Exception if the DataFrame is not sorted - not BQ specific, but dangerous
 .3 implementation: make limit work well for BQ. Can generate postgres specific SQL

thijs-obj:
Only issue left is what to do when we use positional slicing on an unsorted DataFrame or Series. Raising an exception might break some workflows, so this is not that straight forward. I am sure the discussion of handling non-deterministic operations will pop-up again in the future, so for now I'll just close this. The tests work.

PR #1211, opened 2022-09-01
Bach[Athena] DataFrame.head()

thijs-obj:
* Add test for DataFrame.head()
* Factor out code responsible for generating limit expression. And make that work for all databases.

This fixes issue https://github.com/objectiv/objectiv-analytics/issues/1194

This is related to https://github.com/objectiv/objectiv-analytics/issues/1022. This makes the test from that issue passable, but more work is needed to make that test faster. Will do that in a later PR

### DIFF
--- a/bach/bach/dataframe.py
+++ b/bach/bach/dataframe.py
@@ -25,7 +25,8 @@ from sql_models.graph_operations import update_placeholders_in_graph, get_all_pl
 from sql_models.model import SqlModel, Materialization, CustomSqlModelBuilder, RefPath
 
 from sql_models.sql_generator import to_sql
-from sql_models.util import quote_identifier, is_bigquery, DatabaseNotSupportedException, is_postgres
+from sql_models.util import quote_identifier, is_bigquery, DatabaseNotSupportedException, is_postgres, \
+    is_athena
 
 if TYPE_CHECKING:
     from bach.partitioning import Window, GroupBy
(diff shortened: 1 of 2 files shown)


bach/bach/dataframe.py
```diff
@@ -2188,6 +2170,43 @@ def get_current_node(
...
 
+    def _get_limit_clause(self, limit: Optional[slice]) -> Expression:
+        """
+        Give a SQL limit expression based on the provided slice.
+        """
+        if limit is None:
+            return Expression.construct('')
+
+        # Validation
+        if limit.step is not None:
```
jansentom, line 2181: really nice!

bach/bach/dataframe.py
```diff
@@ -2188,6 +2170,43 @@ def get_current_node(
...
+            # BigQuery does not support 'limit all', and 'offset' has to come after 'limit'
+            # See https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax
+            #   #limit_and_offset_clause
+            if is_bigquery(self.engine):
+                if limit.stop is not None:
+                    return Expression.construct(f'limit {limit.stop - limit.start} offset {limit.start}')
+                big_number = 2**63-1  # use a big number, because 'all' is not supported
+                return Expression.construct(f'limit {big_number} offset {limit.start}')
+
+            # 'Normal' databases
```
jansentom, line 2202: haha
```suggestion
            # Other databases
```

jansentom (approved):
There's quite a bit of code left untested. I'm okay with that, since it's not currently used, but maybe we should at least make a note?

thijs-obj:
> There's quite a bit of code left untested. I'm okay with that, since it's not currently used, but maybe we should at least make a note?

https://github.com/objectiv/objectiv-analytics/pull/1213 contains a test that covers a lot of codepaths

jansentom:
> > There's quite a bit of code left untested. I'm okay with that, since it's not currently used, but maybe we should at least make a note?
> #1213 contains a test that covers a lot of codepaths

Ah, I overlooked that one. Will review as well!

thijs-obj:
> Ah, I overlooked that one. Will review as well!
I quickly created that PR after your comment ;)

KathiaBarahona (approved):

PR #1213, opened 2022-09-01
Bach optimize `test_positional_slicing()`, and enable for BigQuery and Athena

thijs-obj:
This addresses issue https://github.com/objectiv/objectiv-analytics/issues/1022 and makes sure the changes of PR https://github.com/objectiv/objectiv-analytics/pull/1211 are well tested on Athena and Postgres too.

Changes:
1. Make `test_positional_slicing()` use a single query
2. Enable `test_positional_slicing()` for Athena and BigQuery

Not changed, but something I wrote in the issue originally: We don't raise an exception if there is no sorting set. Even though limit doesn't make any sense if there is no order by, this might break `head()` in annoying ways. Maybe 'random' data is better then often. Let me know what you think

### DIFF
--- a/bach/tests/functional/bach/test_df_getitem.py
+++ b/bach/tests/functional/bach/test_df_getitem.py
@@ -1,11 +1,14 @@
 """
 Copyright 2021 Objectiv B.V.
 """
+from copy import deepcopy
+from typing import List
+
 import pytest
 
 from bach import DataFrame, Series, SeriesDict
 from tests.functional.bach.test_data_and_utils import assert_equals_data, df_to_list, \
-    get_df_with_test_data
+    get_df_with_test_data, TEST_DATA_CITIES_FULL
 
 
 def test_get_item_single(engine):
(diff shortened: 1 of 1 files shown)


bach/tests/functional/bach/test_df_getitem.py
```diff
@@ -84,19 +88,31 @@ def __getitem__(self, key):
...
+        bt_slice['slice'] = i
+        all_dfs.append(bt_slice)
+        expected_data = [row + [i] for row in base_expected_data][s]
+        all_expected_data.extend(expected_data)
+
+    df = all_dfs[0].append(all_dfs[1:])
+    df = df.reset_index(drop=True)
+    df = df.sort_values(by=['slice', 'skating_order'])
+
+    assert_equals_data(
```
KathiaBarahona, line 111: nice optimization!

KathiaBarahona (approved):

jansentom (approved):
Nice improvements

PR #1214, opened 2022-09-02
MH: Bugfix for context extraction

jansentom:
Fix two bugs
- cookie_id context overlapped cookie_id field on PG, causing the cookie to disappear and the field to have the wrong type
- multipart contexts like InputValueContext, got translated from input_value to Input_valueContext.

And fix broken test data, where cookie_id and CookieContextId values don't match.


modelhub/modelhub/series/series_objectiv.py
```diff
@@ -74,8 +74,11 @@ def get_contexts(self, name: str) -> 'SeriesGlobalContext':
         """
         Returns an array of contexts from an Objectiv stack where `_type` matches the
         UpperCamelCase context and created from `name` e.g. 'identity' becomes 'IdentityContext'.
+
+        @param name:    The name of the context-array to retrieve
```
thijs-obj, line 78:
```suggestion
        :param name:    The name of the context-array to retrieve
```

jansentom: whow. Thanks.

jansentom:
> lgtm, I added documentation, please check.

Very nice, thanks! Merging this.

modelhub/modelhub/series/series_objectiv.py
thijs-obj, line 78:
```suggestion
        Returns an array of contexts from an Objectiv stack where `_type` matches the
        UpperCamelCase context created from `name` e.g. if `name` is 'input_value', then this returns all
        contexts where the `_type` field has value 'InputValueContext'.

        :param name: The name of the context-array to retrieve, in snake_case
```

I just did not get that from the original docstring

bach/docs/source/example-notebooks/open-taxonomy.rst
```diff
@@ -313,12 +313,14 @@ From the marketing context, for example, we can therefore also get the 'source'
 
 	df['marketing_source'] = df.marketing.context.source
 
-When instantiating the model hub, global contexts are added using the name of the context in lower case, 
-without the word 'context', ie. 'HttpContext' becomes 'http':
+When instantiating the model hub, global contexts are added using the name of the context without the word
+'Context' and converted to 'snake_case' (the name of the context split before every capital letter and
+joining them with `_`). I.e. 'HttpContext' becomes 'http' and 'InputValueContext' becomes 'input_value' when
```
thijs-obj, line 318: This doesn't render correctly. Everything between `_` is made italic, so that's from the first `_` in 'snake_case' to the `_` just before the parenthesis. The fact that the second underscore is between backticks doesn't seem to matter. Not sure how to fix that.

thijs-obj (commented):
Some comments about the documentation. Otherwise OK for me

bach/docs/source/example-notebooks/open-taxonomy.rst
jansentom: Good catch @thijs-obj. @ivarpruijn can you please heeeeelp!

ivarpruijn: Can pick this up tomorrow. Quick solution should be to escape the `_` with a prepending backslash (`\_`) I think, but haven't tested that yet.

jansentom: Discussed with Ivar, he'll pick this up tmr with some other docs changes. We'll go ahead and merge without it.

ivarpruijn: Tried a few things, but what it comes down to, is that this is a Markdown parser issue. The exact same issue occurs on the MD parser on GitHub, e.g. see https://github.com/objectiv/objectiv-analytics/blob/main/notebooks/open-taxonomy-how-to.ipynb . 

Basically what happens is that the first underscore is seen as the start of an emphasis block and the single underscore in the literal (\`_`\) is parsed as the end of that emphasis block, even though it's between literal quotes. If you replace the literal part with a regular underscore, the same behavior occurs.

As this is so edge case, I'm going to 'fix' it by rewriting this paragraph slightly.

ivarpruijn: Workaround is here: https://github.com/objectiv/objectiv-analytics/pull/1231/commits/b59e25588a526458c971dd8451036609b2654afa

PR #1221, opened 2022-09-05
[Bach] Support Materialization.SOURCE 

jansentom:
Previously, every df.from_table('table_name') would create a first SQL node like 
```sql
WITH someCTE AS (SELECT fieldA, fieldB from `table_name`)
```
This wasted the partitioning on BQ, as well as did unnecessary processing. 
This PR aims to resolve that by making the first sql node just a reference to the source table, with `Materialization.SOURCE`

### DIFF
--- a/bach/sql_models/sql_generator.py
+++ b/bach/sql_models/sql_generator.py
@@ -128,9 +128,9 @@ def _to_sql_materialized_node(
     queries = _to_cte_sql(dialect=dialect, compiler_cache=compiler_cache, model=model)
     queries = _filter_duplicate_ctes(queries)
     if len(queries) == 0:
-        # _to_cte_sql only returns an empty list in synthetic test scenarios, where there is a source node,
-        # but it's not referenced.
-        return ''
+        # _to_cte_sql only returns an empty list when there is a source node,
+        # but it's not referenced. We can't generate anything in that scenario.
+        raise Exception('No models to compile')
 
     if len(queries) == 1:
         return _materialize(dialect=dialect, sql_query=queries[0].sql, model=model)
--- a/bach/sql_models/sql_generator.py
+++ b/bach/sql_models/sql_generator.py
@@ -292,7 +292,16 @@ def _single_model_to_sql(dialect: Dialect,
     """
     if model.hash in compiler_cache:
         return compiler_cache[model.hash]
+
+    result: List[SemiCompiledTuple] = []
     sql = model.sql
+    if sql == '':
+        # If the SQL is completely empty, this is the first node that we're selecting from.
+        # We only need to store it so we can select from the materialization name
+        compiler_cache[model.hash] = result
+        return result
+
+
     # If there are any format strings in the placeholder values that need escaping, they should have been
     # escaped by now.
     # Otherwise this will cause trouble the next time we call format() below for the references
@@ -304,7 +313,7 @@ def _single_model_to_sql(dialect: Dialect,
     _reference_names[REFERENCE_UNIQUE_FIELD] = model.hash
     sql = _format_sql(sql=sql, values=_reference_names, model=model)
     ctes = raw_sql_to_selects(sql)
-    result: List[SemiCompiledTuple] = []
+
     for cte in ctes[:-1]:
         # For all CTEs the name should be set. Only for the final select (== cte[-1]) it will be None.
         assert cte.name is not None
--- a/bach/sql_models/sql_generator.py
+++ b/bach/sql_models/sql_generator.py
@@ -261,8 +261,13 @@ def model_to_name(dialect: Dialect, model: SqlModel):
     """
     Get the name for the cte/table/view that will be generated from this model, quoted and escaped.
     """
-    # max length of an identifier name in Postgres is normally 63 characters. We'll use that as a cutoff
-    # if we don't know better
+    if model.materialization == Materialization.SOURCE:
+        if model.materialization_name is None:
+            raise Exception('SqlModel with SOURCE materialization must have materialization_name set.')
+        return model.materialization_name
+
+    # max length of an identifier name in Postgres is normally 63 characters.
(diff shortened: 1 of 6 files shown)


bach/sql_models/sql_generator.py
```diff
@@ -258,23 +257,28 @@ def _to_cte_sql(dialect: Dialect,
...
-def model_to_name(model: SqlModel):
+def model_to_name(dialect: Dialect, model: SqlModel):
     """
     Get the name for the cte/table/view that will be generated from this model, quoted and escaped.
     """
     # max length of an identifier name in Postgres is normally 63 characters. We'll use that as a cutoff
-    # here.
+    # if we don't know better
+    if is_bigquery(dialect):
+        max_length = 1023
```
jansentom, line 267: This was required (well, not the 1023, but that's just the max for BQ according to the docs) to support very_long_project_name.also_quite_a_long_dataset_name.table_name` references on BQ. Not uncommon, as it was triggered by one of the tests.

bach/sql_models/sql_generator.py
```diff
@@ -292,10 +296,18 @@ def _single_model_to_sql(dialect: Dialect,
...
+    if sql == '':
+        # If the SQL is completely empty, this is the first node that we're selecting from.
+        # We only need to store it in order to select from the materialization name
+        compiler_cache[model.hash] = result
+        return result
+
     # If there are any format strings in the placeholder values that need escaping, they should have been
     # escaped by now.
-    # Otherwise this will cause trouble the next time we call format() below for the references
+    # Otherwise, this will cause trouble the next time we call format() below for the references
```
jansentom, line 310: OCD, sorry

bach/sql_models/sql_generator.py
```diff
@@ -258,23 +257,28 @@ def _to_cte_sql(dialect: Dialect,
     return result
 
 
-def model_to_name(model: SqlModel):
+def model_to_name(dialect: Dialect, model: SqlModel):
     """
     Get the name for the cte/table/view that will be generated from this model, quoted and escaped.
     """
     # max length of an identifier name in Postgres is normally 63 characters. We'll use that as a cutoff
-    # here.
```
thijs-obj, line 266: I think we want to do something like this
```python
    if model.materialization == Materialization.CTE:
        # max length of an identifier name in Postgres is normally 63 characters. We'll use that as a cutoff
        # here.
        max_length = 63
        if model.materialization_name is not None:
            return model.materialization_name[0:max_length]
        name = f'{model.generic_name[0:(max_length - 35)]}___{model.hash}'
        return name

    if model.materialization_name is None:
        raise Exception('SqlModel with non-cte materialization must have materialization_name set.')
    return model.materialization_name
```

If materialization is not CTE, then we just want to give the name unchanged I guess. As even for postgres then it could be something like database_name.schema_name.table_name so can be more than 63 characters.
For the names of CTEs, I think it's sort of nice to cut names if they are too long.

thijs-obj (commented):
Didn't quite finish the review yet. But will leave the two comments I had so far here as I'll be out of office for a bit.

bach/sql_models/sql_generator.py
jansentom: Yeah, that's nicer. Maybe even give the full name only when we are Materialization.SOURCE?

bach/sql_models/sql_generator.py
```diff
@@ -128,8 +128,9 @@ def _to_sql_materialized_node(
     queries = _to_cte_sql(dialect=dialect, compiler_cache=compiler_cache, model=model)
     queries = _filter_duplicate_ctes(queries)
     if len(queries) == 0:
-        # _to_cte_sql should never return an empty list, but this make it clear we have a len > 0 below.
-        raise Exception('Internal error. No models to compile')
+        # _to_cte_sql only returns an empty list in synthetic test scenarios, where there is a source node,
+        # but it's not referenced.
+        return ''
```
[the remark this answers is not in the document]
jansentom: yes, you are totally right. This should not have made it here.  Fixed in 0f418bc4678b367d7ebc1fe71b1ebca1e5735503

bach/sql_models/sql_generator.py
jansentom: Did something in between: https://github.com/objectiv/objectiv-analytics/pull/1221/commits/e9fe355d657183d2b6f841e825f75456a920dfd7 
Your suggestion did not work for query and virtual materialization types. Felt that it's better to just have plain source tables as the exception, vs adding a list of exceptions even though we don't really need them.

thijs-obj (approved):

bach/tests/functional/bach/test_df_from.py
```diff
@@ -65,8 +66,9 @@ def test_from_table_basic(engine, unique_table_test_name):
     assert df.is_materialized
     assert df.base_node.columns == ('a', 'b', 'c', 'd', 'e', 'f')
     # there should only be a single model that selects from the table, not a whole tree
-    # todo: in the future introduce a special SqlModel type 'source', so we don't even need a first model
-    # with a query and we can just query directly from the source table.
+    assert df.base_node.materialization == Materialization.SOURCE
+    with pytest.raises(Exception, match="No models to compile"):
```
thijs-obj, line 70: Seems more like something to test under `tests/functional/sql_models/`. But don't feel too strongly about that

bach/tests/functional/bach/test_df_from.py
```diff
@@ -65,8 +66,9 @@ def test_from_table_basic(engine, unique_table_test_name):
     assert df.is_materialized
     assert df.base_node.columns == ('a', 'b', 'c', 'd', 'e', 'f')
     # there should only be a single model that selects from the table, not a whole tree
-    # todo: in the future introduce a special SqlModel type 'source', so we don't even need a first model
```
thijs-obj, line 68: noice :+1:

PR #1195, opened 2022-08-30
Bach[Athena] Removed `athena_supported` marker from tests

KathiaBarahona:
`athena_supported` marker is removed, now tests that don't work for athena are marked as`skip_athena`

Issue #1209, opened 2022-09-01
Bach: support arbitrary names for Series


thijs-obj:
Names for Series are currently limited by the column names that a database supports, because Series names map directly to column names.

We want to remove this limitation.

Discussed approach:
- `Series` is responsible for going from 'real-name' to 'column-name'
- Column name translation algorithm should work both ways
- Should be transparent to user (as long as user doesn't look at or use the generated SQL)

thijs-obj:
Make sure to enable the test `bach/tests/functional/bach/test_df_setitem.py::test_set_series_column_name_with_spaces()` for all databases afterwards

ivarpruijn:
Is this issue resolved @thijs-obj ?

thijs-obj:
Yes. There is still one sub-issue open https://github.com/objectiv/objectiv-analytics/issues/1317, but it's low prio and the main issue can be considered to be solved.

PR #1223, opened 2022-09-06
Bach tests athena 2 - Annotate tests with permanent skip marks

thijs-obj:
Cutting up https://github.com/objectiv/objectiv-analytics/pull/1195/ into smaller pieces, and adding some stuff. This is step 2.

Change: Add marks to all tests that we want to skip for Athena, and that we are not going to fix later on.

### DIFF
--- a/bach/tests/functional/bach/test_df_setitem.py
+++ b/bach/tests/functional/bach/test_df_setitem.py
@@ -204,8 +204,8 @@ def test_set_series_column(engine):
     assert filtered_bt.town == filtered_bt['town']
 
 
-@pytest.mark.skip_bigquery_todo('We do not yet support spaces in column names on BigQuery')
-@pytest.mark.skip_athena_todo('We do not yet support spaces in column names on Athena')
+@pytest.mark.skip_bigquery_todo('#1209 We do not yet support spaces in column names on BigQuery')
+@pytest.mark.skip_athena_todo('#1209 We do not yet support spaces in column names on Athena')
 def test_set_series_column_name_with_spaces(engine):
     bt = get_df_with_test_data(engine)
     bt['spaces in column'] = bt['founding']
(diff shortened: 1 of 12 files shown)


bach/tests/functional/bach/test_df_setitem.py
```diff
@@ -204,7 +204,8 @@ def test_set_series_column(engine):
     assert filtered_bt.town == filtered_bt['town']
 
 
-@pytest.mark.skip_bigquery("Bigquery doesn't support spaces in column names")
+@pytest.mark.skip_bigquery_todo('We do not yet support spaces in column names on BigQuery')
```
thijs-obj, line 207: This change and the next one are a bit a-typical: using the `_todo` mark here instead of the paremanent skip mark. Included these in this PR because this was initially a permanent skip. But with issue https://github.com/objectiv/objectiv-analytics/issues/1209 these are on the road map to be cross-database tests

jansentom: like it. Something like this is even nicer?
```suggestion
@pytest.mark.skip_bigquery_todo('#1209 We do not yet support spaces in column names on BigQuery')
```

bach/tests/functional/bach/test_injection_column_name_escaping.py
```diff
@@ -31,13 +31,10 @@ def test_column_names(engine):
     assert_equals_data(bt, expected_columns=expected_columns, expected_data=expected_data)
 
 
+@pytest.mark.skip_bigquery_todo('We do not yet support special characters in column names on BigQuery')
```
jansentom, line 34: Same here

jansentom (approved):

bach/tests/functional/bach/test_injection_column_name_escaping.py
thijs-obj: Changed in e81b9381defc3dc2e33973d30f1dc8a473cc0ee3

bach/tests/functional/bach/test_df_setitem.py
thijs-obj: Changed in e81b9381defc3dc2e33973d30f1dc8a473cc0ee3

PR #1226, opened 2022-09-06
Bach tests athena 5 - Run all tests against Athena

thijs-obj:
Cutting up https://github.com/objectiv/objectiv-analytics/pull/1195 into smaller pieces, and adding some stuff. This is step 5.

Changes:
* Marked all not-yet passing tests as `pytest.mark.skip_athena_todo()`
  * Different from PR [#1195](https://github.com/objectiv/objectiv-analytics/pull/1195): also marked tests that don't pass because of non-supported characters in column names with `skip_athena_todo()`, instead of changes the column names
* Invert logic: instead of only running tests against Athena that are marked as `athena_supported` now run all tests unless marked as `athena_skip` or `athena_skip_todo`

### DIFF
--- a/bach/tests/functional/bach/test_df_rename.py
+++ b/bach/tests/functional/bach/test_df_rename.py
@@ -1,7 +1,7 @@
 """
 Copyright 2021 Objectiv B.V.
 """
-from sql_models.util import is_bigquery
+from sql_models.util import is_bigquery, is_athena
 from tests.functional.bach.test_data_and_utils import assert_equals_data, get_df_with_test_data
 
 EXPECTED_DATA = [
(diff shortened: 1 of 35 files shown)


bach/tests/functional/bach/test_df_rename.py
```diff
@@ -30,8 +30,8 @@ def test_rename_complex(engine):
     bt = get_df_with_test_data(engine)
 
     new_name = 'stêd'
-    if is_bigquery(engine):
-        # BigQuery limits the allowed column names.
+    if is_athena(engine) or is_bigquery(engine):
+        # Athena and BigQuery limit the allowed column names.
```
thijs-obj, line 34: This change is a bit out of character for this PR. But I needed it to get to all tests passing

jansentom (approved):

PR #1232, opened 2022-09-07
Correct some wrongly classified tests as db_independent and db specific

thijs-obj:
### Background and PR
After all PRs mentioned in this comment https://github.com/objectiv/objectiv-analytics/pull/1195#issuecomment-1238586678 were merged. I had a look at what we still needed to do to get BigQuery and Athena to fully-supported, by looking at the tests that we still need to fix. I noticed that the numbers didn't quite match up, because some tests weren't classified correctly. This PR fixes that

### Some Numbers on tests
* `154` tests are not yet working for Athena: `pytest tests/ -m 'skip_athena_todo' --athena --collect-only`
* `32` tests are not yet working for BigQuery: `pytest tests/ -m 'skip_bigquery_todo' --big-query --collect-only`
* `597` tests in total.
* When running with `--all`, we run `1432` tests
  * `55` are database  independent
  * `525` run against Postgres or use the Postgres Dialect
  * `493` run against BigQuery or use the BigQuery Dialect
  * `359` run against Athena or use the Athena Dialect

### DIFF
--- a/bach/tests/unit/bach/test_series_multi_level.py
+++ b/bach/tests/unit/bach/test_series_multi_level.py
@@ -5,9 +5,8 @@ from bach.expression import Expression, MultiLevelExpression
 from sql_models.util import is_postgres, is_bigquery
 from tests.unit.bach.util import get_fake_df_test_data
 
-pytestmark = pytest.mark.skip_athena_todo()  # TODO: Athena
 
-
+@pytest.mark.db_independent('tested function does not care about engine or dialect')
 def test_series_numeric_interval_levels_dtypes() -> None:
     supported_dtypes = SeriesNumericInterval.get_supported_level_dtypes()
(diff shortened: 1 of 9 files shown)


bach/tests/unit/bach/test_series_multi_level.py
```diff
@@ -5,9 +5,8 @@
 from sql_models.util import is_postgres, is_bigquery
 from tests.unit.bach.util import get_fake_df_test_data
 
-pytestmark = pytest.mark.skip_athena_todo()  # TODO: Athena
-
 
+@pytest.mark.db_independent('tested function does not care about engine or dialect')
```
jansentom, line 9: This does not add anything for me
```suggestion
@pytest.mark.db_independent
```

jansentom (approved):
Nice one, cool to have stats now!

PR #1243, opened 2022-09-14
Fix property types in generated iglu schema

borft:
- use actual property types from base schema
- regenerate session context iglu schema, hit_number is now an actual number


jansentom (approved):
Good fix

PR #1259, opened 2022-09-19
[MH/BQ] Allow secrets to be used directly from environment

jansentom:
Next to `bq_credentials_path` which reads BQ credentials from a file, we now also support `bq_credentials_env` which reads from the specified env variable, so less boilerplate code is required when using secrets in Hex/Deepnote, etc.
```    
df = modelhub.get_objectiv_dataframe(start_date=start_date, end_date=end_date,
            db_url='bigquery://objectiv-production/snowplow',
            bq_credentials_env='SECRET_ENV_VAR'
)
```

### DIFF
--- a/bach/setup.cfg
+++ b/bach/setup.cfg
@@ -35,7 +35,7 @@ include_package_data = True
 
 [options.extras_require]
 bigquery = 
-    sqlalchemy-bigquery
+    sqlalchemy-bigquery>=1.4.0
     google-cloud-bigquery-storage
 athena =
     pyathena>=2.10.0
--- a/modelhub/modelhub/modelhub.py
+++ b/modelhub/modelhub/modelhub.py
@@ -1,6 +1,8 @@
 """
 Copyright 2021 Objectiv B.V.
 """
+import base64
+import os
 import re
 from typing import List, Union, Dict, Tuple, Optional, cast, Any
 from typing import TYPE_CHECKING
@@ -92,7 +94,6 @@ class ModelHub:
         If db_url is for BigQuery, bq_credentials_path or bq_credentials must be provided.
         When both are given, bq_credentials wins.
         """
-        import os
         kwargs: Dict[str, Any] = {}
 
         if db_url and re.match(r'^bigquery://.+', db_url):
@@ -100,7 +101,6 @@ class ModelHub:
                 raise ValueError('BigQuery credentials or path is required for engine creation.')
 
             if bq_credentials:
-                import base64
                 credentials_base64 = base64.b64encode(bq_credentials.encode('utf-8'))
                 kwargs['credentials_base64'] = credentials_base64
             else:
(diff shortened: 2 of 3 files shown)


modelhub/modelhub/modelhub.py
```diff
@@ -84,18 +84,35 @@ def conversion_events(self):
...
+
+            credentials = os.environ.get(bq_credentials_env)
+            if not credentials:
+                raise ValueError('BigQuery credentials environment variable is empty.')
+
+            from tempfile import NamedTemporaryFile
+            with NamedTemporaryFile(mode='w') as creds:
+                creds.write(credentials)
+                creds.flush()
+                return create_engine(db_url, credentials_path=creds.name)
```
thijs-obj, line 114: Dug around the sqlalchemy_bigquery code a bit, I think there is an alternative that's nicer:

```python
import base64
bsecret = base64.b64encode(credentials.encode('utf-8'))
return create_engine(db_url, credentials_base64=bsecret)
```

I have to do some testing on this tomorrow, but seems this should work

modelhub/modelhub/modelhub.py
```diff
@@ -129,7 +147,10 @@ def get_objectiv_dataframe(
         :param end_date: last date for which data is loaded to the DataFrame. If None, data is loaded up to
             and including the last date in the sql table. Format as 'YYYY-MM-DD'.
         :param bq_credentials_path: path for BigQuery credentials. If db_url is for BigQuery engine, this
-            parameter is required.
+            parameter or `bq_credentials_env` is required.  When both are given, bq_credentials_path wins.
+        :param bq_credentials_env: Name of the environment variable that contains the json for the credentials
+            file. If db_url is for BigQuery engine, this parameter or `bq_credentials_path` is required.
```
thijs-obj, line 152: I think passing the name of the environment variable is not really elegant. Just passing the credentials itself seems cleaner and more generally usable.

If the user happens to already have the data in an environment variable, then it can just do `bq_credentials=os.environ['SECRET_ENV_VAR']`

modelhub/modelhub/modelhub.py
jansentom: Fixed in 98442b47f4fef5e16e2f0946def6bade653b7aaf

modelhub/modelhub/modelhub.py
jansentom: Fixed in 98442b47f4fef5e16e2f0946def6bade653b7aaf

bach/setup.cfg
```diff
@@ -35,7 +35,7 @@ include_package_data = True
 
 [options.extras_require]
 bigquery = 
-    sqlalchemy-bigquery
+    sqlalchemy-bigquery>=1.4.0
```
jansentom, line 38: earlier versions don't support base64 encoded credentials

jansentom (commented):
@thijs-obj can you have another look plz?

thijs-obj (approved):

PR #1260, opened 2022-09-20
Bach[Athena]: support from_table for athena

KathiaBarahona:
Added `catalog_name` to aws athena URL, otherwise sqlalchemy raises an error when trying SELECT statements. I'm still missing adding a mapping of integer db types for athena, as we currently support only `bigint` and we should also support:
`tinyint`, `smallint`, `int`/`integer`. Sort of the same for `double`, `float`, `decimal` and `char` + `var`. It's better for us to discuss if we should change `supported_db_dtype` (from Series class) as `Dict[DBDialect, Tuple[str, ...]]` or we have a separate dict for athena.

### DIFF
--- a/bach/.secrets/.test_env.sample
+++ b/bach/.secrets/.test_env.sample
@@ -32,7 +32,7 @@ OBJ_DB_ATHENA_REGION_NAME="eu-west-1"
 OBJ_DB_ATHENA_SCHEMA_NAME="bach_test"
 OBJ_DB_ATHENA_CATALOG_NAME="automated_tests"
 OBJ_DB_ATHENA_S3_STAGING_DIR="s3://obj-automated-tests/athena_query_results/"
-OBJ_DB_ATHENA_LOCATION="s3://obj-automated-tests/bach_test/staging/"
+OBJ_DB_ATHENA_LOCATION="s3://obj-automated-tests/bach_test/"
 OBJ_DB_ATHENA_WORK_GROUP="automated_tests_work_group"
 
 
--- a/bach/.secrets/.test_env.sample
+++ b/bach/.secrets/.test_env.sample
@@ -29,8 +29,10 @@ OBJ_DB_ATHENA_AWS_ACCESS_KEY_ID=
 OBJ_DB_ATHENA_AWS_SECRET_ACCESS_KEY=
 
 OBJ_DB_ATHENA_REGION_NAME="eu-west-1"
-OBJ_DB_ATHENA_SCHEMA_NAME="automated_tests.bach_test"
-OBJ_DB_ATHENA_S3_STAGING_DIR="s3://obj-automated-tests/bach_test/staging/"
+OBJ_DB_ATHENA_SCHEMA_NAME="bach_test"
+OBJ_DB_ATHENA_CATALOG_NAME="automated_tests"
+OBJ_DB_ATHENA_S3_STAGING_DIR="s3://obj-automated-tests/athena_query_results/"
+OBJ_DB_ATHENA_LOCATION="s3://obj-automated-tests/bach_test/staging/"
 OBJ_DB_ATHENA_WORK_GROUP="automated_tests_work_group"
 
 
--- a/bach/bach/from_database.py
+++ b/bach/bach/from_database.py
@@ -49,7 +49,6 @@ def get_dtypes_from_table(
         meta_data_table = 'INFORMATION_SCHEMA.COLUMNS'
     elif is_bigquery(engine):
         meta_data_table, table_name = _get_bq_meta_data_table_from_table_name(table_name)
-        filters_expr.append(Expression.construct(f"table_name='{table_name}'"))
     elif is_athena(engine):
         catalog_name = (
             f"{engine.url.query['catalog_name']}." if 'catalog_name' in engine.url.query else ''
--- a/bach/bach/from_database.py
+++ b/bach/bach/from_database.py
@@ -54,6 +54,7 @@ def get_dtypes_from_table(
             f"{engine.url.query['catalog_name']}." if 'catalog_name' in engine.url.query else ''
         )
         meta_data_table = f'{catalog_name}INFORMATION_SCHEMA.COLUMNS'
+        # This filter is needed because athena does not limit to data from the current schema
         filters_expr.append(Expression.construct(f"table_schema='{engine.url.database}'"))
     else:
         raise DatabaseNotSupportedException(engine)
--- a/bach/bach/from_database.py
+++ b/bach/bach/from_database.py
@@ -1,16 +1,17 @@
 """
 Copyright 2022 Objectiv B.V.
 """
-from typing import Dict, Optional, Tuple
+from typing import Dict, Tuple
 
 from sqlalchemy.engine import Engine
 
+from bach.expression import Expression, join_expressions
 from bach.types import get_dtype_from_db_dtype, StructuredDtype
 from bach.utils import escape_parameter_characters
 from sql_models.constants import DBDialect
 from sql_models.model import SqlModel, CustomSqlModelBuilder
 from sql_models.sql_generator import to_sql
-from sql_models.util import is_postgres, DatabaseNotSupportedException, is_bigquery
+from sql_models.util import is_postgres, DatabaseNotSupportedException, is_bigquery, is_athena
 
 
 def get_dtypes_from_model(engine: Engine, node: SqlModel) -> Dict[str, StructuredDtype]:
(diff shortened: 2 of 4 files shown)


bach/bach/from_database.py
```diff
@@ -43,16 +44,27 @@ def get_dtypes_from_table(
...
         meta_data_table = 'INFORMATION_SCHEMA.COLUMNS'
     elif is_bigquery(engine):
         meta_data_table, table_name = _get_bq_meta_data_table_from_table_name(table_name)
+        filters_expr.append(Expression.construct(f"table_name='{table_name}'"))
+    elif is_athena(engine):
+        catalog_name = (
+            f"{engine.url.query['catalog_name']}." if 'catalog_name' in engine.url.query else ''
+        )
+        meta_data_table = f'{catalog_name}INFORMATION_SCHEMA.COLUMNS'
+        filters_expr.append(Expression.construct(f"table_schema='{engine.url.database}'"))
```
KathiaBarahona, line 58: thought that the engine was smart enough to know to fetch only information from the defined schema, but nope. Filter is needed as multiple schemas can have the same table

bach/tests/conftest.py
```diff
@@ -67,8 +67,10 @@
 _DB_ATHENA_AWS_ACCESS_KEY_ID = _ENV.get('OBJ_DB_ATHENA_AWS_ACCESS_KEY_ID')
 _DB_ATHENA_AWS_SECRET_ACCESS_KEY = _ENV.get('OBJ_DB_ATHENA_AWS_SECRET_ACCESS_KEY')
 _DB_ATHENA_REGION_NAME = _ENV.get('OBJ_DB_ATHENA_REGION_NAME', 'eu-west-1')
-_DB_ATHENA_SCHEMA_NAME = _ENV.get('OBJ_DB_ATHENA_SCHEMA_NAME', 'automated_tests.bach_test')
-_DB_ATHENA_S3_STAGING_DIR = _ENV.get('OBJ_DB_ATHENA_S3_STAGING_DIR', 's3://obj-automated-tests/bach_test/staging/')
+_DB_ATHENA_CATALOG_NAME = _ENV.get('OBJ_DB_ATHENA_CATALOG_NAME', 'automated_tests')
+_DB_ATHENA_SCHEMA_NAME = _ENV.get('OBJ_DB_ATHENA_SCHEMA_NAME', 'bach_test')
+_DB_ATHENA_S3_STAGING_DIR = _ENV.get('OBJ_DB_ATHENA_S3_STAGING_DIR', 's3://obj-automated-tests/athena_query_results/')
+DB_ATHENA_LOCATION = _ENV.get('OBJ_DB_ATHENA_LOCATION', 's3://obj-automated-tests/bach_test/staging/')
```
KathiaBarahona, line 73: changed both of this, as is better to have a folder for output results only for both BQ and MH.

KathiaBarahona: Also, I think there might be a bit of struggle when working more on writable db tests for Athena, as there might be a bit of conflict when writing data on the DB location... might affect other tests that read from this folder. Maybe partitioning will solve this?

bach/bach/from_database.py
```diff
@@ -43,16 +44,27 @@ def get_dtypes_from_table(
         BigQuery, e.g. 'project_id.dataset.table_name'
     :return: Dictionary with as key the column names of the table, and as values the dtype of the column.
     """
+    filters_expr = []
     if is_postgres(engine):
         meta_data_table = 'INFORMATION_SCHEMA.COLUMNS'
     elif is_bigquery(engine):
         meta_data_table, table_name = _get_bq_meta_data_table_from_table_name(table_name)
+        filters_expr.append(Expression.construct(f"table_name='{table_name}'"))
```
thijs-obj, line 52:
```suggestion
```

This same expression is added below on line 62

bach/.secrets/.test_env.sample
```diff
@@ -29,8 +29,10 @@ OBJ_DB_ATHENA_AWS_ACCESS_KEY_ID=
 OBJ_DB_ATHENA_AWS_SECRET_ACCESS_KEY=
 
 OBJ_DB_ATHENA_REGION_NAME="eu-west-1"
-OBJ_DB_ATHENA_SCHEMA_NAME="automated_tests.bach_test"
-OBJ_DB_ATHENA_S3_STAGING_DIR="s3://obj-automated-tests/bach_test/staging/"
+OBJ_DB_ATHENA_SCHEMA_NAME="bach_test"
+OBJ_DB_ATHENA_CATALOG_NAME="automated_tests"
+OBJ_DB_ATHENA_S3_STAGING_DIR="s3://obj-automated-tests/athena_query_results/"
+OBJ_DB_ATHENA_LOCATION="s3://obj-automated-tests/bach_test/staging/"
```
thijs-obj, line 35: Having 'staging' in the path name of the location is very confusing to me, because there is also a `staging_dir` but that is a different directory in this case.

bach/tests/functional/bach/test_df_from.py
```diff
@@ -12,14 +12,24 @@
...
     # TODO: insert data too, and check in the tests below that we get that data back in the DataFrame
     #  https://github.com/objectiv/objectiv-analytics/issues/1093
+    if is_athena(engine):
+        with engine.connect() as conn:
+            conn.execute(f'drop table if exists {table_name};')
+            conn.execute(
+                (
+                    f'create external table {table_name}'
+                    '(a bigint, b string, c double, d date, e timestamp, f boolean) \n'
+                    f"location '{DB_ATHENA_LOCATION}';"
```
thijs-obj, line 29: Should we not have a location per test?
Maybe something like
```suggestion
                    f"location '{DB_ATHENA_LOCATION}/{table_name}';"
```

Of course this test doesn't actually write any data, so I guess this might not yet matter :thinking:

bach/bach/from_database.py
thijs-obj: How we use expressions here. We might just as well use plain strings I think, but don't feel too strongly about that

thijs-obj:
> I'm still missing adding a mapping of integer db types for athena, as we currently support only `bigint` and we should also support:
> `tinyint`, `smallint`, `int`/`integer`.

I am not sure if that's really needed, at least not any time soon. Postgres has similar integer types, and there we also only support `bigint`. So far that has been enough I believe.

thijs-obj (commented):
Left a code suggestion and some questions/comments

KathiaBarahona:
> > I'm still missing adding a mapping of integer db types for athena, as we currently support only `bigint` and we should also support:
> > `tinyint`, `smallint`, `int`/`integer`.
> 
> I am not sure if that's really needed, at least not any time soon. Postgres has similar integer types, and there we also only support `bigint`. So far that has been enough I believe.

:/ sadly it is, when getting dtypes from snowplow staging data it will raise an error, there are some columns with `integer` type.

bach/bach/from_database.py
KathiaBarahona: whoops, forgot to remove it.

bach/.secrets/.test_env.sample
KathiaBarahona: can change it to read from `bach_test` directly. Left it because I thought it meant some other thing. To be fair, AWS query param naming was a bit misleading.

bach/tests/functional/bach/test_df_from.py
KathiaBarahona: huh that actually might help when avoiding multiple tests reading from the same location. will try

KathiaBarahona: Added here https://github.com/objectiv/objectiv-analytics/pull/1260/commits/c25e6d8f1b4eff48a9c887e52f8fce79ec9273f4

bach/.secrets/.test_env.sample
KathiaBarahona: removed `staging` from path here https://github.com/objectiv/objectiv-analytics/pull/1260/commits/c25e6d8f1b4eff48a9c887e52f8fce79ec9273f4

bach/bach/from_database.py
KathiaBarahona: https://github.com/objectiv/objectiv-analytics/pull/1260/commits/182d65c4ee25ea852eefec77075c1e48a6290e58

jansentom (approved):
I agree with Thijs on the "small" data types. Small comment but LGTM in general

bach/bach/from_database.py
```diff
@@ -43,16 +44,26 @@ def get_dtypes_from_table(
...
+    filters_expr.append(Expression.construct(f"table_name='{table_name}'"))
+    filters_stmt = join_expressions(filters_expr, join_str=' AND ').to_sql(engine.dialect)
     sql = f"""
         select column_name, data_type
         from {meta_data_table}
-        where table_name = '{table_name}'
+        where {filters_stmt}
         order by ordinal_position;
     """
     return _get_dtypes_from_information_schema_query(engine=engine, query=sql)
```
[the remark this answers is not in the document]
KathiaBarahona: I think I asked the same a long time ago, but we decided to let `from_table` handle the error.

PR #1269, opened 2022-09-20
Bump Bach and Modelhub version to v0.0.23

jansentom:


ivarpruijn:
Tested all notebooks on PG & BQ and made one fix. LGTM (with the known limitation on BQ for marketing-analytics).

ivarpruijn (approved):

PR #1270, opened 2022-09-20
Doctests for user intent example notebook

ivarpruijn:
Reformats the feature-importance example notebook, and adds doctests.

### DIFF
--- a/notebooks/basic-user-intent.ipynb
+++ b/notebooks/basic-user-intent.ipynb
@@ -5,10 +5,7 @@
    "id": "afb5dddf",
    "metadata": {},
    "source": [
-    "This is one of the Objectiv example notebooks. For more examples visit the \n",
-    "[example notebooks](https://objectiv.io/docs/modeling/example-notebooks/) section of our docs. The notebooks can run with the demo data set that comes with the our [quickstart](https://objectiv.io/docs/home/try-the-demo/), but can be used to run on your own collected data as well.\n",
-    "\n",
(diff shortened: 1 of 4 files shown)


notebooks/basic-user-intent.ipynb
```diff
@@ -160,33 +154,50 @@
...
+    "- Users that spent most time on the site (the 90th percentile), and specifically in our documentation sections, are in the Implement phase.\n",
+    "- As there's a jump beyond the one minute mark at the 70th percentile, we assume that users in the 70th to 90th percentile duration in our documentation sections are in the Explore phase.\n",
+    "- The remaining users are Informing themselves about the product. Those users are spending less than 1:30 in the docs and/or spend any amount of time on our main website.\n",
+    "\n",
+    "Summarizing:\n",
     "\n",
     "| User intent | Root locations | Duration |\n",
     "| :--- | :--- | :--- |\n",
     "| 1 - Inform | *all sections other than the ones mentioned below* | *any time spent* |\n",
```
vard-obj, line 198: this line is redundant because the 3 other cases below are covering all the possibilities.

vard-obj, line 198: the first row in the table is redundant because the 3 other cases below are covering all the possibilities.

ivarpruijn: Isn't `all sections other than the ones mentioned below`  a different case than the other ones, which specifically look at `Docs: modeling, taxonomy, tracking, home`? Though I agree there's likely no matching data currently, as it should all fall within those root_locations if it's instrumented properly.

vard-obj: Right, let's leave it as it is.

vard-obj (approved):

PR #1273, opened 2022-09-21
ModelHub: Refactor of `ObjectivStack.get_contexts_with_type`

KathiaBarahona:
Moved code from `ObjectivStack.get_contexts_with_type` to new class `ContextFlattening`. Since we require it for flattening global contexts and location stack for Athena, which caused code duplication with current implementation.

### DIFF
--- a/modelhub/modelhub/pipelines/extracted_contexts.py
+++ b/modelhub/modelhub/pipelines/extracted_contexts.py
@@ -15,6 +15,7 @@ import modelhub
 from bach.types import StructuredDtype
 from sqlalchemy.engine import Engine
 
+from modelhub.series.operations.context_flattening import ContextFlattening
 from modelhub.util import (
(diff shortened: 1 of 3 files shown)


jansentom (commented):
As we discussed IRL:
I don't like moving/abstracting this even further away from where it's used: the extracted context pipeline. We already have a codepath per Dialect there, so let's incorporate this logic there as well (maybe as private functions in the per-dialect/format classes there?).

Okay to do this after merging this chain of PRs to not create a huge mess.

KathiaBarahona:
> As we discussed IRL: I don't like moving/abstracting this even further away from where it's used: the extracted context pipeline. We already have a codepath per Dialect there, so let's incorporate this logic there as well (maybe as private functions in the per-dialect/format classes there?).
> 
> Okay to do this after merging this chain of PRs to not create a huge mess.

Sounds good to me

PR #1274, opened 2022-09-21
ModelHub[Athena]: Support SessionizedDataPipeline for Athena

KathiaBarahona:
Seems that we don't require any changes for this


jansentom (approved):
NIIIICE

PR #1275, opened 2022-09-21
ModelHub[Athena]: Support IdentityResolutionPipeline

KathiaBarahona:
Seems that everything is working, just needed to change the `_FAKE_DATA` format


jansentom (approved):
Also really nice :)

PR #1277, opened 2022-09-21
Bach[Athena]: Fix for JSON slicing

KathiaBarahona:
Current implementation was raising errors when no element contains filtering.

### DIFF
--- a/bach/bach/series/series_json.py
+++ b/bach/bach/series/series_json.py
@@ -787,7 +787,7 @@ class JsonAthenaAccessorImpl(Generic[TSeriesJson]):
         stop_expression = self._get_slice_partial_expr(value=key.stop, is_start=False)
 
         values_expression = Expression.construct(
-            "slice(cast({} as array(json)), {}, {})",
+            "try(slice(cast({} as array(json)), {}, {}))",
             self._series_object,
             start_expression,  # startpoint of slice
             Expression.construct('({} - {})', stop_expression, start_expression),  # length of slice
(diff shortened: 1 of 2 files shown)


bach/bach/series/series_json.py
```diff
@@ -864,56 +864,57 @@ def _find_in_json_list(self, filtering_slice: Dict[str, str], is_start: bool):
...
             )
             for key, value in filtering_slice.items()
         ]
-        # filter_array_map_expr: array of maps, but only with the maps that match the filter criteria
-        filter_array_map_expr = Expression.construct(
-            'filter({}, x -> ({}))',
+
+        # extracts only the keys to match
+        transform_array = Expression.construct(
+            'transform({}, x-> cast(map_from_entries(array[{}]) as json))',
```
jansentom, line 890: Should we use `__x` instead, since this is basically an internal name that could shadow a field previously defined and used (and thus colliding) within the expression?

jansentom (commented):
This fixes something that was not tested before since the tests where green? Do we need more tests for this or does recategorizing it as "same as BQ" in the only touched test this appropriately?

bach/bach/series/series_json.py
KathiaBarahona: sure :+1:

KathiaBarahona:
> This fixes something that was not tested before since the tests where green? Do we need more tests for this or does recategorizing it as "same as BQ" in the only touched test this appropriately?

Yeah, the main problem was providing `0` (value returned when an element does not exist in the array) as an index, the main fix was just not raising this exception

PR #1281, opened 2022-09-21
Doctests for machine-learning example notebook

ivarpruijn:
Reformats the machine-learning example notebook, and adds doctests.

### DIFF
--- a/bach/docs/source/example-notebooks/machine-learning.rst
+++ b/bach/docs/source/example-notebooks/machine-learning.rst
@@ -8,73 +8,345 @@
 Bach and sklearn
 ================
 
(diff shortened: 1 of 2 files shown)


vard-obj (approved):
Shall we rename `machine-learning.rst ` -> `sklearn-example.rst `? But the rest lgtm.

ivarpruijn:
> Shall we rename `machine-learning.rst ` -> `sklearn-example.rst `? But the rest lgtm.

Good idea, but it will also change the URL in the docs. Though we could put a redirect for that in place of course, I think we should keep it as-is for now, and review the types of notebooks we have later on. Merging this PR for now.

PR #1287, opened 2022-09-22
[DO NOT MERGE] Experiment: Validation libraries

sdirosa:


schema/validator-cli.js
```diff
@@ -0,0 +1,22 @@
...
+const { success, error } = validate(JSON.parse(event));
+
+if (!success) {
+  console.log(error);
+  console.log("\nNOPE: The given event is not valid\n");
+  process.exit(1);
+}
+
+console.log("\nOK: The given event is valid\n");
+process.exit(0);
```
sdirosa, line 22: Super basic cli that only validates events. If we like we can introduce many parameters to specify which validator to run against and customize the outputs, for now this works nicely to test things from the command line when needed.

schema/validator-test.js
```diff
@@ -0,0 +1,129 @@
+/*
+ * Copyright 2022 Objectiv B.V.
+ */
```
sdirosa, line 3: Everything in here has been manually written. We should generate this whole file from the base_schema.

schema/validator.js
```diff
@@ -0,0 +1,266 @@
+/*
+ * Copyright 2022 Objectiv B.V.
+ */
```
sdirosa, line 3: Everything in here has been manually written. We should generate this whole file from the base_schema.

tracker/core/utilities/src/generateSchema.js
```diff
@@ -386,7 +386,7 @@ const files = fs.readdirSync(schema_dir);
 // read all schema files
 const all_schema = {};
 files.forEach((fn) => {
-  if (fn.match(/[a-z0-9_]+\.json5?$/) && !fn.match('event_list.json5')) {
+  if (fn.match(/[a-z0-9_]+\.json5$/) && !fn.match('event_list.json5')) {
```
sdirosa, line 389: Unrelated bugfix. Our schemas are all json5; this was loading the typescript config as schema erroneously.

schema/validator.js
```diff
@@ -0,0 +1,266 @@
...
+        "PressEvent requires PressableContext or LinkContext in its LocationStack",
+    }
+  ),
+}).strict();
+
+// TODO we can use a discriminated union for much nicer error messages after this gets merged:
+//  - https://github.com/colinhacks/zod/issues/1171
+// z.discriminatedUnion('_type', [InputChangeEvent, PressEvent])
+
+export const validate = z.union([InputChangeEvent, PressEvent]).safeParse;
```
sdirosa, line 266: `validate` is a union of all available event validations. It accepts a json object as parameter and returns an object with two keys: `{success: true, data}` in case of valid event or `{success: false, error}` in case of failure.

tracker/core/utilities/src/generateSchema.js
borft: ah yes, this was intentional, to be able to load extensions that were allowed to be .json or .json5

sdirosa: 👍🏼  I'll skip this change when finalizing the new generator and move it out of here in its own clean branch/PR.

PR #1293, opened 2022-09-26
ModelHub[Athena]: Aggregate model support for athena

KathiaBarahona:
Seems that most of the functions work, needed some small changes


modelhub/modelhub/aggregate.py
```diff
@@ -507,19 +512,15 @@ def drop_off_locations(data: bach.DataFrame,
...
-        series = data.groupby(groupby)['__feature_nice_name'].sort_by_series(by=by,
-                                                                             ascending=True)
-        series_json_array = cast(bach.SeriesString, series).to_json_array()
-
-        drop_loc = series_json_array.json[-1].materialize()
+        window = data.sort_values(by='moment', ascending=True).groupby(groupby).window(
+            end_boundary=bach.partitioning.WindowFrameBoundary.FOLLOWING,
+        )
+        drop_loc = window['__feature_nice_name'].window_last_value()
+        drop_loc = drop_loc.materialize(distinct=True)
```
KathiaBarahona, line 519: replaced `to_json_array`, as for PG and BQ adds extra quotes to the scalar string values, this does not happen for athena. That's why slicing in tests was odd. Not sure if we should have the same behavior for Athena or remove extra quotes for PG and BQ, what do you think @thijs-obj ?

modelhub/modelhub/aggregate.py
```diff
@@ -507,19 +512,15 @@ def drop_off_locations(data: bach.DataFrame,
...
-        by = [data['moment']]
-        series = data.groupby(groupby)['__feature_nice_name'].sort_by_series(by=by,
-                                                                             ascending=True)
-        series_json_array = cast(bach.SeriesString, series).to_json_array()
-
-        drop_loc = series_json_array.json[-1].materialize()
+        window = data.sort_values(by='moment', ascending=True).groupby(groupby).window(
+            end_boundary=bach.partitioning.WindowFrameBoundary.FOLLOWING,
+        )
+        drop_loc = window['__feature_nice_name'].window_last_value()
```
thijs-obj, line 518: Completely not related to this PR, and no action required here. But general comment, as I think we should do this differently for future functionality:

IMHO, if a column name is returned in the result, like `__feature_nice_name` here, then we shouldn't prepend it with `__`, we should just call it `feature_nice_name`. We should only prepend columns that are internal to the function. This is inconsistent, and now the end-user will actually have to change the column name before he can feed it to another function, as the end-user is 'not allowed' to pass dataframes with column names that start with `__`

modelhub/tests_modelhub/functional/modelhub/test_model_hub_drop_off_locations.py
```diff
@@ -20,34 +19,38 @@ def test_drop_off_locations(db_params):
     # adding sorting column
     bts = bts.reset_index()
     bts['sort_str'] = bts['__feature_nice_name'].astype(dtype=str).str[8:10]
```
thijs-obj, line 21: Now I sort of understand what you meant above with the "slicing in tests was odd". But I don't quite understand WHY the slicing was odd. If I look at the data for the `__feature_nice_name` column in the expected data, then I don't see any extra quotes nor any changes, but still this slice changed. Can you help me understand what's going on here?

thijs-obj (commented):
Some questions. Code looks good to me

modelhub/modelhub/aggregate.py
KathiaBarahona: I agree we should follow Athena behaviour, I'll open an issue for this.

PR #1301, opened 2022-09-27
Bach map names pre tests

thijs-obj:
Preparation for issue #1209 

* Add tests that use column names with special characters and/or lower and upper case characters
* Add `use_pandas=True` to some tests that will need that, as they use special character in column names
* other preparatory changes to tests

Note that not all tests work for BigQuery currently. Specific issue is that BigQuery column names are case-insensitive. So for example `a` and `A` are actually considered to be the same column name by BQ, but Bach is not aware of that.  #1209 will fix that.

### DIFF
--- a/bach/tests/functional/bach/test_df_append.py
+++ b/bach/tests/functional/bach/test_df_append.py
@@ -1,10 +1,12 @@
 import numpy as np
 import pandas as pd
+import pytest
 
 from bach import DataFrame
 from tests.functional.bach.test_data_and_utils import assert_equals_data
 
 
+@pytest.mark.skip_athena_todo('https://github.com/objectiv/objectiv-analytics/issues/1209')
 def test_append_w_aligned_columns(engine) -> None:
     caller_pdf = pd.DataFrame({'A': [1, 2, 3, 4, 5], 'B': ['a', 'b', 'c', 'd', 'e']})
     other_pdf = pd.DataFrame({'A': [6, 7, 8, 9], 'B': ['f', 'g', 'h', 'i']})
@@ -17,6 +19,8 @@ def test_append_w_aligned_columns(engine) -> None:
     np.testing.assert_equal(expected.to_numpy(), result.to_numpy())
 
 
+@pytest.mark.skip_bigquery_todo('https://github.com/objectiv/objectiv-analytics/issues/1209')
+@pytest.mark.skip_athena_todo('https://github.com/objectiv/objectiv-analytics/issues/1209')
 def test_append_w_non_aligned_columns(engine) -> None:
     # Column names 'A#' and 'a#' should be considered as two different columns, i.e. they don't align
     # By adding the '#' to the column names we also test the correct handling of special characters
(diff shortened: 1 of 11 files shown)


bach/tests/unit/bach/test_series.py
```diff
@@ -51,22 +52,22 @@ def get_df(index_names: List[str], data_names: List[str]):
     expr_test = Expression.construct('test')
     expr_other = Expression.construct('test::text')
 
-    sleft = int_type(engine=engine, base_node=None, index={}, name='test',
+    sleft = int_type(engine=engine, base_node=base_node, index={}, name='test',
```
thijs-obj, line 55: Not actually passing a base_node was okay, because the tested code didn't need a real base_node. But a follow-up PR changes that

bach/tests/functional/bach/test_df_indexing.py
```diff
@@ -80,11 +79,6 @@ def test_basic_indexing_column_based(indexing_dfs: Tuple[pd.DataFrame, DataFrame
 def test_index_slicing(indexing_dfs: Tuple[pd.DataFrame, DataFrame]) -> None:
     pdf, df = indexing_dfs
 
-    if is_bigquery(df.engine):
```
thijs-obj, line 83: We silently didn't run these tests for BigQuery. But they pass

bach/tests/functional/bach/test_df_append.py
```diff
@@ -1,39 +1,46 @@
...
 
+@pytest.mark.skip_bigquery_todo('https://github.com/objectiv/objectiv-analytics/issues/1209')
+@pytest.mark.skip_athena_todo('https://github.com/objectiv/objectiv-analytics/issues/1209')
 def test_append_w_non_aligned_columns(engine) -> None:
-    caller_pdf = pd.DataFrame({'a': [1, 2, 3, 4, 5], 'b': ['a', 'b', 'c', 'd', 'e']})
-    other_pdf = pd.DataFrame({'d': [6, 7, 8, 9], 'c': ['f', 'g', 'h', 'i']})
+    # Column names 'A#' and 'a#' should be considered as two different columns, i.e. they don't align
+    # By adding the '#' to the column names we also test the correct handling of special characters
+    caller_pdf = pd.DataFrame({'A#': [1, 2, 3, 4, 5], 'b': ['a', 'b', 'c', 'd', 'e']})
+    other_pdf = pd.DataFrame({'a#': [6, 7, 8, 9], 'c': ['f', 'g', 'h', 'i']})
```
thijs-obj, line 28: BigQuery thinks `a#` and `A#` is the same. So this change actually broke this test for BigQuery, hence the `skip_bigquery_todo`. This will be fixed in a follow-up PR.

KathiaBarahona: Are we planning to merge these issues to the main after both 1209 are merged? Else this will mean that `main` will have some untested functionalities for both BQ and Athena, which are non-related to arbitrary names.

KathiaBarahona (commented):
lgtm, tho just one concern

thijs-obj:
Discussed out of band. Agreed to merge

PR #1298, opened 2022-09-26
Up notebooks & containers

ivarpruijn:
To do:

- [x] Update modelhub-basics notebook as soon as https://github.com/objectiv/objectiv-analytics/pull/1278 is approved.

Merge only when:

- [x] https://github.com/objectiv/objectiv.io/pull/567 is live.
- [x] [Docker image update PR](https://github.com/objectiv/objectiv-analytics/pull/1299) is ready.

PR #1305, opened 2022-09-28
Fix feature-engineering docs

vard-obj:


ivarpruijn:
Thanks for fixing this. I also fixed this in https://github.com/objectiv/objectiv-analytics/pull/1298, which will be merged today and used for the release, so it's okay to close this PR.

Issue #1309, opened 2022-09-29
Bach: Converting empty spaces to UUID


KathiaBarahona:
For both BigQuery and Athena, we use string dbtype for representing UUID. The problem with this is that the column might contain empty strings, resulting in errors when doing `to_pandas()`

PR #1310, opened 2022-09-29
ModelHub: Replace empty spaces with None for user_id and event_id

KathiaBarahona:
We should avoid having empty spaces in those columns, else it will raise an error when trying to do `to_pandas()`. 

Added an issue for Bach to handle better this situation: https://github.com/objectiv/objectiv-analytics/issues/1309

### DIFF
--- a/modelhub/modelhub/pipelines/extracted_contexts.py
+++ b/modelhub/modelhub/pipelines/extracted_contexts.py
@@ -322,16 +322,19 @@ class SnowplowExtractedContextsPipeline(BaseExtractedContextsPipeline, ABC):
                 'se_category': ObjectivSupportedColumns.STACK_EVENT_TYPES.value,
             }
         )
+
+        # avoid having empty string in series that fill user_id series, otherwise
+        # there is a risk of raising an exception when trying to parse string values to UUID
+        uuid_series = ['network_userid', 'domain_sessionid']
+        for series in uuid_series:
+            df_cp.loc[df_cp[series] == '', series] = None
+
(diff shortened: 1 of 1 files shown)


modelhub/modelhub/pipelines/extracted_contexts.py
```diff
@@ -328,6 +328,11 @@ def _process_data(self, df: bach.DataFrame) -> bach.DataFrame:
         )
         df_cp = df_cp.drop(columns=['network_userid', 'domain_sessionid'])
 
+        # avoid having empty string in series that will be parsed as UUID
+        uuid_series = [ObjectivSupportedColumns.EVENT_ID, ObjectivSupportedColumns.USER_ID]
```
jansentom, line 332: Bigger question is, how is this possible? These columns should always be set to something reasonable. `event_id` for sure, but I can't think of a way `user_id` could be an empty string either.  Not a big fan of fix therefore.

KathiaBarahona: This is because both columns are defined as "string". I tried telling Hive to interpret empty string as NULL just for those columns, but is not possible :/. Can only do it for the entire table. So, when receiving empty string and then calling `to_pandas()`, it will raise an error since it's not a valid uuid string.  The way to change this, is having "\N" instead of empty string in the csv files, but cannot change that at this point.

KathiaBarahona: as discussed, changed this statement, and replace values only for `'network_userid', 'domain_sessionid'` (before generating user_id series.)

jansentom (approved):
nice, thanks!

PR #1314, opened 2022-09-29
Checklock: Enable checks for Athena

KathiaBarahona:
Added athena as supported engine in Checklock.


jansentom (approved):
worked out of the box (after setting up .env correctly):
`make requirements-env checklock-inspection`

PR #1315, opened 2022-09-29
Modify location_stack type in drop_off_locations function

vard-obj:


hendrik-obj (commented):
lgtm, only small comment about the docstring

Issue #1317, opened 2022-09-29
Bach - support arbitrary column names in database_create_table()


thijs-obj:
Follow up issue of https://github.com/objectiv/objectiv-analytics/issues/1209

See this TODO: <add link>

tasks:
* If the DataFrame has columns with names that we normally encode, but are valid. Then make sure we use the original column names in the created table. Typical: upper case characters, we lower case them.
* add tests
* make sure that the usage in `DataFrame.get_sample()` is tested too

ivarpruijn:
Is this issue resolved @thijs-obj ?

thijs-obj:
Yes

PR #1306, opened 2022-09-28
Bach map names; preparation changes

thijs-obj:
Assorted small changes that I made while working on issue #1209.

Changes:
* docstrings
* variable naming
* code structure (but not semantics).
* some small other changes

Generally these changes have no effect on the outcome of any functions, and no tests are changed. Main purpose of this PR is to make the follow-up PR smaller and more focussed.

PR #1320, opened 2022-09-29
Bach: support arbitrary Series names on BQ and Athena

thijs-obj:
Basic implementation of supporting arbitrary column names on Athena and BigQuery.

### Improving method
We might switch the algorithm to determine the column name later to an algorithm that gives clearer names in more cases. Currently the names are not good. But merging this will unblock working on other things, and then we can work on generating nicer names.

### Main ideas implemented in this PR:
* `get_sql_column_name()` translates a series-name to a sql-column name
* There is no new state that we track.
  * `DataFrame` and `Series` still track the series-name, it's just that the sql-expression generated is slightly different.
  * When creating a new DataFrame with `get_instance()` or `copy_override()` we need to provide `name_to_column_mapping` so we can generate the correct sql-expressions.
* `name_to_column_mapping()` does the reverse of  `get_sql_column_name()`: it translate from sql-column name to series-name.
  * We only need this in for `revert_merge()`.
  * We could in the future use a non-reversible algorithm in `get_sql_column_name()`. Then we just need to track the name to column name mapping the `BachSqlModel` too.

### Other:
* BigQuery column names are case-insensitive, just like Athena's. In the past we allowed both lower case and upper case characters for BigQuery, which lead to problems. e.g. `a` and `A` could be different series in our DataFrame, but the same column in BigQuery.
  * This PR enables some tests again for BQ that were disabled in https://github.com/objectiv/objectiv-analytics/pull/1301 because the case-insensitivity caused them to fail.
* Generally enabled a lot of tests for Athena.


bach/bach/expression.py
```diff
@@ -337,13 +341,14 @@ def replace_column_references(self, old_column_name: str, new_column_name: str)
             replaced_tokens.append(ColumnReferenceToken(new_column_name))
         return self.__class__(replaced_tokens)
 
-    def remove_table_column_references(self) -> Tuple[str, str, 'Expression']:
+    def remove_table_column_references(self) -> Tuple[Optional[str], Optional[str], 'Expression']:
```
thijs-obj, line 344: This is an unrelated change, but so small I figured I just leave it in.

As part of this PR, I was working on some code that called this function, and found it very unintuitive so changed this. Couldn't put this in the #1306 PR as it actually changes some behaviour.

bach/tests/functional/bach/test_series_numeric.py
```diff
@@ -121,7 +121,7 @@ def test_aggregations_sum_mincount(engine):
         assert (math.isnan(pd_agg) and bt_agg_value is None) or bt_agg_value == pd_agg
 
 
-@pytest.mark.skip_athena_todo('https://github.com/objectiv/objectiv-analytics/issues/1209')
+@pytest.mark.skip_athena_todo()  # TODO: Athena
```
thijs-obj, line 124: These tests are still failing, but for different reasons. So still marked as `skip_athena_todo`

bach/tests/unit/bach/test_utils.py
```diff
@@ -51,8 +54,133 @@ def test_is_valid_column_name(dialect):
...
-@pytest.mark.db_independent
-def test_validate_sorting_expressions() -> None:
+class ColNameEncoded(NamedTuple):
+    name: str
+    postgresql: str
+    awsathena: str
+    bigquery: str
+
+
+def test_get_sql_column_name(dialect):
```
thijs-obj, line 64: This test gives an idea of how series-names get converted to column names.

For each test-case, the first string is the series-name, and the following strings are the column names for Postgres, Athena, and BigQuery

bach/bach/dataframe.py
```diff
@@ -1994,18 +2023,24 @@ def to_pandas(self, limit: Union[int, slice] = None) -> pandas.DataFrame:
...
-                series_name_to_dtype[series.name] = pandas_info.dtype
+                sql_column_name = get_sql_column_name(dialect=dialect, name=name)
+                series_name_to_dtype[sql_column_name] = pandas_info.dtype
 
         with self.engine.connect() as conn:
             # read_sql_query expects a parameterized query, so we need to escape the parameter characters
             sql = escape_parameter_characters(conn, sql)
             pandas_df = pandas.read_sql_query(sql, conn, dtype=series_name_to_dtype)
 
+        # Rename the columns that the query gives, to the actual column names of the DataFrame
```
KathiaBarahona, line 2040: IMHO we should avoid mixing `columns` and `series` words (even for Pandas)
```suggestion
        # Rename the columns that the query gives, to the actual series names of the DataFrame
```

bach/bach/utils.py
```diff
@@ -67,18 +68,23 @@ def is_valid_column_name(dialect: Dialect, name: str) -> bool:
         # Identifiers longer than 63 characters are not necessarily wrong, but they will be truncated which
         # could lead to identifier collisions, so we just disallow it.
         # source: https://www.postgresql.org/docs/14/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS
-        return len(name) < 64
+        return 0 < len(name) < 64
```
KathiaBarahona, line 71: **question** Is it possible that `len` might return a negative value? :thinking:

KathiaBarahona (commented):
left a couple of questions, still need to check tests. Though, will be also nice to create a test for `df.unstack`, unless its already done in another PR.

thijs-obj:
>  will be also nice to create a test for `df.unstack`, unless its already done in another PR.

Good one. Enabled and improved the tests in 7099c356fe3550cb2a5d10fbca34d40e447d1694

bach/bach/utils.py
thijs-obj: Don't thinks so. But it could return `0`

bach/bach/dataframe.py
```diff
@@ -537,6 +539,7 @@ def from_table(
             table_name: str,
             index: List[str],
             all_dtypes: Optional[Mapping[str, StructuredDtype]] = None,
+            name_to_column_mapping: Optional[Mapping[str, str]] = None
```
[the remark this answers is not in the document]
thijs-obj: I do think the mapping is needed. e.g. in `sample()` (will be fixed in later issue https://github.com/objectiv/objectiv-analytics/issues/1317) we want to write a table. If there are series with non-allowed characters, then we'll have to write the table with encoded column names. Then when we read the table back, we do want to use the original series names again, so we want to set this mapping to map the actual series names to the encoded column names.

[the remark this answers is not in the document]
thijs-obj: Don't think if we need to do more validation. if there are wrong keys in the name_to_column_mapping they will simply be ignored. That might give different column names, than the user expected, but he'll notice that quickly enough I think.

bach/bach/dataframe.py
```diff
@@ -595,6 +603,9 @@ def from_model(
         :param all_dtypes: Optional. Mapping series name to dtype.
             Must contain all index and data series.
             Must be in same order as the columns appear in the sql-model.
+        :param name_to_column_mapping: Optional mapping from series names to the actual column names in the
```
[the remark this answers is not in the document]
thijs-obj: See https://github.com/objectiv/objectiv-analytics/pull/1320#discussion_r989178755

bach/bach/dataframe.py
thijs-obj: Fixed in 51398ace5bada07f9cd84b7a2546e65afb2f7327

bach/bach/dataframe.py
```diff
@@ -741,12 +760,14 @@ def get_instance(
...
         savepoints: 'Savepoints',
-        variables: Dict['DtypeNamePair', Hashable]
+        variables: Dict['DtypeNamePair', Hashable],
+        name_to_column_mapping: Mapping[str, str],
     ) -> 'DataFrame':
         """
         INTERNAL: Get an instance with the right series instantiated based on the dtypes array.
 
         This assumes that base_node has a column for all names in index_dtypes and dtypes.
+        If name_to_column_mapping is set, it maps series names to the sql column names in base_node.
```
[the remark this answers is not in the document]
thijs-obj: I changed the wording in 51398ace5bada07f9cd84b7a2546e65afb2f7327

bach/bach/dataframe.py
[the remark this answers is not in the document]
KathiaBarahona: alright!

bach/bach/utils.py
KathiaBarahona: duh,nvm

KathiaBarahona (approved):

PR #1323, opened 2022-10-04
Checklock: Store kernel execution ouputs as pickle files

KathiaBarahona:
- Extends IPython Kernel, KernelSpec and Shell for storing cell outputs as pickled dataframes. Will later reads from stored outputs and compare between engines.


KathiaBarahona:
splitted this PR in multiple branches.

Issue #1093, opened 2022-08-02
Bach: add test that reads data from a table


thijs-obj:
Background: All our functional bach tests create their own data using `DataFrame.from_pandas(materialization='cte')`, which is efficient but doesn't create an actual table with data.

Goals:
1. Have a test that creates a table with data, and checks that that data is accessible with `DataFrame` and that operations can be done on it.
2. [Remove this comment afterwards](https://github.com/objectiv/objectiv-analytics/pull/1092/files#diff-41499784e40b2f31082ec6339dc2521db271362f27507e0cc2f2d698843e49f6R19-R20)

thijs-obj:
This was achieved by #1324

Issue #1318, opened 2022-09-29
Bach - support arbitrary column with from_*()


thijs-obj:
Follow up issue of https://github.com/objectiv/objectiv-analytics/issues/1209

See this [TODO](https://github.com/objectiv/objectiv-analytics/pull/1320/files#diff-41499784e40b2f31082ec6339dc2521db271362f27507e0cc2f2d698843e49f6R233)

tasks:
* Support tables and models that have column names with e.g. Upper case characters
* add tests

thijs-obj:
Done, see https://github.com/objectiv/objectiv-analytics/pull/1324 and  https://github.com/objectiv/objectiv-analytics/pull/1331

PR #1324, opened 2022-10-04
Bach - Add test data in tests for from_table() and from_model()

thijs-obj:
This resolves issue https://github.com/objectiv/objectiv-analytics/issues/1093.

Fixed this now, as this is needed for the issue https://github.com/objectiv/objectiv-analytics/issues/1318, which is a follow-up issue from PR https://github.com/objectiv/objectiv-analytics/pull/1320

Changes:
* in `test_df_from.py` add test data to the created test tables.
  * Verify that the resulting DataFrames work correctly with actual data.
* other small changes

### DIFF
--- a/bach/bach/testing/utils.py
+++ b/bach/bach/testing/utils.py
@@ -102,8 +102,8 @@ def assert_equals_data(
     else:
         column_names, db_values = _get_to_pandas_data(bt)
 
-    assert len(db_values) == len(expected_data)
-    assert column_names == expected_columns
+    assert len(db_values) == len(expected_data), f'{len(db_values)} != {len(expected_data)}'
+    assert column_names == expected_columns, f'{column_names} != {expected_columns}'
 
     _date_freq = 'ms' if is_athena(bt.engine) else 'us'
     for i, df_row in enumerate(db_values):
--- a/bach/tests/functional/bach/test_df_from.py
+++ b/bach/tests/functional/bach/test_df_from.py
@@ -6,40 +6,82 @@ tests for:
  * DataFrame.from_model()
 
 """
+from datetime import date, datetime
+
 import pytest
 from sqlalchemy.engine import Engine
 
 from bach import DataFrame
+from sql_models.constants import DBDialect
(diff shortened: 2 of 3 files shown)


bach/bach/testing/utils.py
```diff
@@ -102,8 +102,8 @@ def assert_equals_data(
     else:
         column_names, db_values = _get_to_pandas_data(bt)
 
-    assert len(db_values) == len(expected_data)
-    assert column_names == expected_columns
+    assert len(db_values) == len(expected_data), f'{len(db_values)} != {len(expected_data)}'
+    assert column_names == expected_columns, f'{column_names} != {expected_columns}'
```
thijs-obj, line 106: PyTest doesn't rewrite the asserts of this function anymore, because this function is no longer in a file that is named `test_*.py`. As a result the AssertionErrors that these statements gave didn't contain any useful information. This change makes it a bit easier to fix tests if they fail on these asserts.

bach/tests/functional/bach/test_df_from.py
```diff
@@ -6,40 +6,82 @@
...
+
+                insert into {table_name}(a, b, c, d, e, f) values
+                (123, 'test', 1.2345, '2022-01-01', '2000-03-04 05:43:21', true);
+            '''
+    }
+    sql_strings = all_sql_strings[DBDialect.from_engine(engine)]
+    sql_statements = [s for s in sql_strings.split(';') if s.strip()]
+    assert len(sql_statements) == 3
+
+    if not add_data:  # Only keep the 'drop table' and 'create table' statements
```
KathiaBarahona, line 73: **question** Is there a reason why not to insert it always?

thijs-obj: just to save a query. Mainly matters for Athena, where this could save a second or two

thijs-obj: I added a small comment in [dab5abc](https://github.com/objectiv/objectiv-analytics/pull/1324/commits/dab5abc24df5655fd00b70e4fc916e1a31672a55).

That also eliminates an additional query that I overlooked initially

PR #1329, opened 2022-10-06
Validation refactoring

sdirosa:
New validation setup based on [Zod](https://zod.dev/) and [Yellicode](https://www.yellicode.com/introduction). 

## Next steps
If testing is successful we have a few extra steps to take:
- [x] Adjust our schema definition to express all the missing validation rules we currently execute ad-hoc
- [x] Generate the validator script from the schema
- [ ] Generate the SDKs TS schema from the validator
- [ ] Generate the Factories for Core Tracker from the validator via
- [ ] Use the new validator in the SDKs Developer Tools
- [x] Use the new validator for our Collector
- [x] Use the new validator in Snowplow


schema/base_schema.json
```diff
@@ -0,0 +1,540 @@
...
+        "id": {
+          "description": "Unique identifier for a specific instance of an event.",
+          "type": "uuid"
+        },
+        "time": {
+          "description": "Timestamp indicating when the event was generated.",
+          "type": "integer"
+        }
+      },
+      "validation": {
```
ivarpruijn, line 197: According to the old schema, the `AbstractEvent` requires an `ApplicationContext`, did we miss that here @sdirosa @borft ?

borft: No, it is a requirement of the `global_contexts` array

ivarpruijn: Right. Then I'll have to traverse all properties' types to get their requirements and show those in the Taxonomy docs as well (e.g. like for [InputChangeEvent](https://objectiv.io/docs/taxonomy/reference/events/InputChangeEvent)).

borft: yes, but that was already the case, as the whole schema is hierarchical. But I guess in this case it also depends on how you want the docs to be structured. In this particular case, an alternative could be to simply require the GlobalContexts type, and link to a page that describes that in more detail.

btw, I think traversing all objects is something that should already be in the code somewhere, as that is also required for generating the validators, etc

sdirosa: As Mike said, this is a UI choice. Personally I'm not a fan of traversing and exploding the whole hierarchy into a single documentation page. Mainly because it doesn't scale, but also because one "loses" where the requirement is coming from. 

If, instead, we link the GlobalContext property, developers can navigate there and read what a Global Context actually is and what it requires to be valid.

That said, this is my personal preference when reading docs, probably influenced heavily by being used to type systems and lots of nesting and type composition.

sdirosa: About traversing the tree, there's already a method called `getEntityProperties` in the common helpers for the templates which performs this task for the `properties` attribute. 

Let me know if I should extend this to become generically about any property. It's fairly easy to add a parameter to specify the attribute to fetch, instead of the hardcoded properties (e.g. rules, documentation, etc)

schema/base_schema.json
```diff
@@ -0,0 +1,540 @@
+{
+  "name": "base_schema",
```
borft, line 2: @sdirosa should we also remove the old base_schema.json5 in this PR to avoid any confusion?

schema/base_schema.json
```diff
@@ -0,0 +1,540 @@
...
+              "includeContexts": ["InputValueContext"],
+              "by": ["_type", "id", "value"]
+            }
+          ]
+        }
+      ]
+    }
+  },
+
+  "events": {
```
borft, line 169: @sdirosa do you think we need this extra level of nesting? it feels a bit odd, and inconsistent with the toplevel types

sdirosa: Definitely not, I think we should flatten the schema. Every entity can be determined by its parent abstract anyway.

borft (approved):

sdirosa:
Merging this, as it doesn't affect any existing code. I will open a second PR to continue with the leftover generators and their finalization.

PR #1338, opened 2022-10-10
Bach[Athena]: Quantile Support

KathiaBarahona:
As previously discussed, the quantile support is based on a simulation of PERCENTILE_CONT. Also, refactored a bit the previous implementation, as calculation where performed per Series and later concatenated. Now it does all calculations in DataFrame level.


bach/tests/functional/bach/test_series_numeric.py
```diff
@@ -131,15 +130,14 @@ def test_aggregations_quantile(engine):
     for column, quantile in zip(pdf.columns, quantiles):
         expected = pdf[column].quantile(q=quantile)
         result = bt[column].quantile(q=quantile).to_numpy()[0]
-        assert expected == result
+        assert round(expected, 4) == round(result, 4)
```
thijs-obj, line 133: Question: is there a reason to round to `4` and not for example `14` or even more?

The `4` sort of suggest that we are not that accurate, but checking on this the difference is tiny, only like 0.00000000000000000004 (not exact)

bach/bach/quantile.py
```diff
@@ -1,87 +1,281 @@
...
             )
-        else:
-            agg_result = series.copy_override(name=str(qt))._derived_agg_func(
-                partition=partition,
-                expression=AggregateFunctionExpression.construct(
-                    f'percentile_cont({qt}) within group (order by {{}})', series,
-                ),
+            if is_bigquery(df.engine):
+                if window is None or not isinstance(window, Window):
+                    raise Exception('Invalid window object.')
```
thijs-obj, line 84:
```suggestion
```
This can never happen I think, because we set the window on line 63. If this does happen, we get an Exception anyway on the next line.

bach/bach/quantile.py
```diff
@@ -1,87 +1,281 @@
...
+    """
+    Helper function that simulates quantile calculation using linear interpolation based
+    on method 7 of Hyndman & Fan.
+    https://www.amherst.edu/media/view/129116/original/Sample+Quantiles.pdf
+
+    Main reason to use this method is because Numpy's `quantile` and SQL's `percentile_cont` functions are
+    based on it. Therefore, the expression generated by this function MUST yield exact results.
+
+    Formula:
+    𝑄(𝑝)=(1−𝜸)∗𝑋𝑗+𝜸∗𝑋𝑗+1
```
thijs-obj, line 129: I find the naming here and in the code confusing.

What is called `X_j` in the code is not the same as `𝑋𝑗` here, but actually `𝑋𝑗+1`

thijs-obj: Could we rename `X_j_series` to something like `X_j_plus_one_series` or `X_j_next_series` 
And `X_i_series` to `X_j_series`?

thijs-obj (commented):
very nice solution! 

Have one comment about naming, but looks good to me

bach/bach/quantile.py
```diff
@@ -1,87 +1,281 @@
...
+    return quantiles_df[final_series]
+
+
+def _calculate_quantiles_with_linear_interpolation(
+    df: DataFrame, q: Union[float, List[float]], group_by: List[str],
+) -> DataFrame:
+    """
+    Helper function that simulates quantile calculation using linear interpolation based
+    on method 7 of Hyndman & Fan.
+    https://www.amherst.edu/media/view/129116/original/Sample+Quantiles.pdf
```
thijs-obj, line 123: :+1:

bach/bach/quantile.py
KathiaBarahona: yup, will add a comment for it

bach/tests/functional/bach/test_series_numeric.py
KathiaBarahona: not really, can change it to round till 10

bach/bach/quantile.py
KathiaBarahona: sure

KathiaBarahona: done here https://github.com/objectiv/objectiv-analytics/pull/1338/commits/6f8b04c8fb03813f9131deab8f5e69446400b6e8

bach/bach/quantile.py
KathiaBarahona: https://github.com/objectiv/objectiv-analytics/pull/1338/commits/6f8b04c8fb03813f9131deab8f5e69446400b6e8

bach/tests/functional/bach/test_series_numeric.py
KathiaBarahona: https://github.com/objectiv/objectiv-analytics/pull/1338/commits/6f8b04c8fb03813f9131deab8f5e69446400b6e8

thijs-obj (approved):
lgtm

PR #1343, opened 2022-10-12
Bach athena create table simplified

thijs-obj:
Changes:
* Support `DataFrame.database_create_table()` on Athena
  * introduced `ddl_quote_identifier()`
  * split sql in separate statements for Athena
* This also makes `DataFrame.sampe().get_sample()` work, which relies on `database_create_table()`


bach/bach/utils.py
```diff
@@ -177,3 +177,14 @@ def validate_node_column_references_in_sorting_expressions(
                     ' sorting and try again.'
                 )
             )
+
+
+def merge_sql_statements(dialect: Dialect, sql_statements: List[str]) -> List[str]:
```
jansentom, line 182: nice abstraction

bach/tests/functional/bach/test_df_sample.py
```diff
@@ -35,6 +34,7 @@ def test_get_sample(engine, unique_table_test_name):
                                   sample_percentage=50)
 
 
+@pytest.mark.skip_athena('We do not support the seed parameter for athena')
```
jansentom, line 37: Does Athena not support this? It is very useful ...

jansentom (approved):
Small comments, in general LGTM

bach/tests/functional/bach/test_df_sample.py
thijs-obj: I tried and it didn't work. Documentation also doesn't mention it https://prestodb.io/docs/current/sql/select.html#tablesample

bach/bach/utils.py
KathiaBarahona: maybe renaming it to `reduce_sql_statements`? I think I'll expect just a `str` based on the name.

bach/sql_models/util.py
```diff
@@ -78,6 +82,27 @@ def quote_identifier(dialect: Dialect, name: str) -> str:
     raise DatabaseNotSupportedException(dialect)
 
 
+def ddl_quote_identifier(dialect: Dialect, name: str) -> str:
```
KathiaBarahona, line 85: :+1:

KathiaBarahona (approved):
lgtm, just a comment

bach/bach/utils.py
thijs-obj: I agree, that the name is not really perfect. But I don't really like `reduce` either for the exact same reason. A `reduce()` function usually also gives back a single value and not an array with a single element (examples: [1], [2]). So that would be confusing in the exact same way, to me at least.

[1] https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce#:~:text=The%20final%20result%20of%20running%20the%20reducer%20across%20all%20elements%20of%20the%20array%20is%20a%20single%20value.
[2] https://docs.python.org/3/library/functools.html#functools.reduce:~:text=so%20as%20to%20reduce%20the%20iterable%20to%20a%20single%20value

PR #1322, opened 2022-10-03
ModelHub: Location stack visualization

hendrik-obj:
- [x] use funnel and explode to visualize location stacks
- [x] add tests

PR #1297, opened 2022-09-26
ModelHub[Athena]: Remove `skip_athena_todo` marker

KathiaBarahona:
Modelhub now works with Athena!! :partying_face: 

Dependencies:
#1284 
#1289 
#1293
#1294

PR #1348, opened 2022-10-12
Revert "ModelHub[Athena]: Remove `skip_athena_todo` marker"

thijs-obj:
Reverts objectiv/objectiv-analytics#1297  The PR https://github.com/objectiv/objectiv-analytics/pull/1322/files was merged in parallel, and that still requires the marker that #1297 removed.

This is a temporary revert. We need to get rid of that marker again


thijs-obj:
Just going to merge this without review, as it's currently broken on main, and merely reverts an earlier PR.

PR #1354, opened 2022-10-13
Bach: Set default format when casting SeriesAbstractDateTime to SeriesString 

KathiaBarahona:


bach/bach/series/series_datetime.py
```diff
@@ -348,11 +357,27 @@ def dtype_to_expression(cls, dialect: Dialect, source_dtype: str, expression: Ex
         if source_dtype not in cls.supported_source_dtypes:
             raise ValueError(f'cannot convert {source_dtype} to {cls.dtype}')
 
-        if expression.is_constant and source_dtype == 'string':
-            return cls._constant_expression_to_dtype_expression(dialect, expression)
+        if source_dtype == 'string':
+            if expression.is_constant:
+                return cls._constant_expression_to_dtype_expression(dialect, expression)
```
thijs-obj, line 362: I realize this is not part of the PR, but am a bit curious about this part. It just seems very error prone. `_constant_expression_to_dtype_expression` loops over the expression and tries to convert every string token in the expression to a date/time/timestamp. Non-string tokens are ignored. If there happen to be multiple string tokens we convert them one by one. I cannot imagine in what valid scenarios we have multiple string tokens and other tokens, and then that algorithm gives the correct result. Additionally, I don't see any tests for this. Have we tested this in anyway?

Additionally, I think we are not 100% consistent in marking expressions as constants when they are constant. But that's not the fault of this code.

bach/bach/series/series_datetime.py
```diff
@@ -348,11 +357,27 @@ def dtype_to_expression(cls, dialect: Dialect, source_dtype: str, expression: Ex
...
+                    "COALESCE(try(date_parse({}, '%H:%i:%S.%f')), try(date_parse({}, '%H:%i:%S')))",
+                    expression,
+                    expression,
+                )
+                expression = Expression.construct('to_unixtime({})', expression)
 
         return Expression.construct(f'cast({{}} as {cls.get_db_dtype(dialect)})', expression)
 
+    def astype(self, dtype: Union[str, Type]) -> Series:
+        if dtype != SeriesString.dtype and dtype != str:
```
thijs-obj, line 376:
```suggestion
        if get_series_type_from_dtype(dtype) is not SeriesString:
```
SeriesString also has a dtype_alias. that we should treat the same as the dtype. Writing it this way checks both the normal dtype and the aliasses, I think

bach/bach/series/series_datetime.py
```diff
@@ -348,11 +357,27 @@ def dtype_to_expression(cls, dialect: Dialect, source_dtype: str, expression: Ex
...
+            return super().astype(dtype=dtype)
+
+        return self.dt.strftime(self._FINAL_DATE_TIME_FORMAT)
```
thijs-obj, line 379: Question: Is there a reason that we don't do this in `SeriesString.dtype_to_expression()`?

Overriding `astype()` means that there are now two places where casting logic can be. In `astype()` and in `dtype_to_expression()`. That just seems more complex than always having it in the same place.

bach/tests/functional/bach/test_df_describe.py
```diff
@@ -81,8 +80,6 @@ def test_df_numerical_describe(engine) -> None:
     )
 
 
-@pytest.mark.skip_athena_todo()
-@pytest.mark.skip_bigquery_todo()
```
thijs-obj, line 85: Very nice to see all these go :+1:

bach/bach/series/series_datetime.py
```diff
@@ -765,6 +787,21 @@ def get_db_dtype(cls, dialect: Dialect) -> Optional[str]:
             return SeriesFloat64.get_db_dtype(dialect)
         return super().get_db_dtype(dialect)
 
+    def astype(self, dtype: Union[str, Type]) -> Series:
```
thijs-obj, line 790: Same comments/questions here as for astype() above

thijs-obj (commented):
Left some questions and comments

bach/bach/series/series_datetime.py
KathiaBarahona: It's mainly due to the logic we need to follow when formatting SeriesTimeDelta, I didn't wanted to mix all that stuff in SeriesString

bach/bach/series/series_datetime.py
KathiaBarahona: mmmmh I blame myself for this one, I think it was intended for avoiding errors when trying to parse values with unsupported precision. Else athena raises an error if the string has an unexpected format :/

bach/bach/series/series_datetime.py
KathiaBarahona: From `dtype_to_expression` in SeriesString? In that case it means we sort of have to expose the formatting in a public method just for SeriesTimeDelta and looks(IMO) weird

bach/bach/series/series_datetime.py
KathiaBarahona: fixed here https://github.com/objectiv/objectiv-analytics/pull/1354/commits/6c71c17aaa8a85eb6c6c832658845ea57a15ba93

bach/bach/series/series_datetime.py
thijs-obj: > From `dtype_to_expression` in SeriesString?
Yes. That what I was thinking.

But I get the feeling you don't agree :)

bach/bach/series/series_datetime.py
KathiaBarahona: actually we do test this, but yup. https://github.com/objectiv/objectiv-analytics/blob/f4bd49a4a709ef746cb099b6eca9b0f75b4df11f/bach/tests/functional/bach/test_series_timestamp.py#L92-L94 Should be revisited

Issue #1033, opened 2022-07-18
Fix BQ tests.functional.bach.test_df_sort_values.test_sort_values_expression


thijs-obj:
* Make code changes if needed
* Make test use `engine` instead of `pg_engine`

PR #1355, opened 2022-10-13
Bach: Using column reference instead of expression in ORDER BY (BQ and Athena)

KathiaBarahona:
This fixes issue: https://github.com/objectiv/objectiv-analytics/issues/1033


thijs-obj (approved):
lgtm

PR #1360, opened 2022-10-14
Bach: make ffill propagation on windows possible

hendrik-obj:
- [x] make ffill method possible on df with window (only in `ValuePropagation` for now, maybe later add it to `fillna`, `ffill` and `bfill`)
- [x] fix materialization of fillna with method ffill/bfill


bach/tests/functional/bach/test_df_fillna.py
```diff
@@ -91,6 +99,29 @@ def test_fillna_w_methods_against_pandas(engine) -> None:
     )
 
 
+def test_ffill_propagation(engine) -> None:
+    pdf = pd.DataFrame(DATA, columns=list("ABCDEFG"))
+    pdf['groups'] = pdf.B>0
+    df = DataFrame.from_pandas(engine=engine, df=pdf, convert_objects=True).reset_index().sort_values('_index_0')
+    pdf = pdf.reset_index(names='_index_0').sort_values('_index_0')
```
vard-obj, line 106: Shall we change it to this?
```suggestion
    pdf = pdf.reset_index().rename(columns={'index': '_index_0'})
```

bach/bach/operations/value_propagation.py
```diff
@@ -73,17 +76,21 @@ def propagate(
...
 
         # sorting is lost due to materialization, still requires it
         df = df.sort_values(**self._row_sorting)
-        df = self._propagate_first_value_per_partition(df)
+        df = self._propagate_first_value_per_partition(df, window_group=window_group)
 
         # sort values, this way we respect the provided order by
         df = df.sort_values(by=self.ROW_NUMBER_SERIES_NAME)
-        return df[self._df.data_columns]
+        return df[self._df.data_columns].materialize(node_name='nafilled')
```
hendrik-obj, line 87: This is done so you can select from the data frame later. See the test that is added:
```
fillna_check_ffill_expected[fillna_check_ffill_expected.F=='f']
```

bach/bach/operations/value_propagation.py
```diff
@@ -53,7 +53,10 @@ def __init__(self, df: DataFrame, method: str):
         }
 
     def propagate(
-        self, sort_by: Optional[Union[str, Sequence[str]]] = None, ascending: Union[bool, List[bool]] = True,
+        self,
+        sort_by: Optional[Union[str, Sequence[str]]] = None,
+        ascending: Union[bool, List[bool]] = True,
+        window_group: Optional[Union[str, Sequence[str]]] = None
```
KathiaBarahona, line 59: **suggestion** Maybe providing a `WrappedWindow` instead? this is how we normally apply window functions

bach/tests/functional/bach/test_df_fillna.py
```diff
@@ -91,6 +99,29 @@ def test_fillna_w_methods_against_pandas(engine) -> None:
...
+    df = DataFrame.from_pandas(engine=engine, df=pdf, convert_objects=True).reset_index().sort_values('_index_0')
+    pdf = pdf.reset_index().rename(columns={'index': '_index_0'})
+
+    pd.testing.assert_frame_equal(
+        pdf, df.to_pandas(), check_index_type=False, check_names=False,
+    )
+
+    from bach.operations.value_propagation import ValuePropagation
+    vp = ValuePropagation(df, method='ffill')
+    df_ffill = vp.propagate(window_group='groups')
```
KathiaBarahona, line 114: shouldn't it be the same format as pandas? This means we need to expose `ValuePropagation` in bach's API.

bach/tests/functional/bach/test_df_fillna.py
```diff
@@ -91,6 +99,29 @@ def test_fillna_w_methods_against_pandas(engine) -> None:
...
 
+def test_ffill_propagation(engine) -> None:
+    pdf = pd.DataFrame(DATA, columns=list("ABCDEFG"))
+    pdf['groups'] = pdf.B>0
+    df = DataFrame.from_pandas(engine=engine, df=pdf, convert_objects=True).reset_index().sort_values('_index_0')
+    pdf = pdf.reset_index().rename(columns={'index': '_index_0'})
+
+    pd.testing.assert_frame_equal(
+        pdf, df.to_pandas(), check_index_type=False, check_names=False,
+    )
```
KathiaBarahona, line 110: I think it's not necessary to test this, else we will be doing 2 query executions and might take some time for Athena and BQ
```suggestion
```

bach/bach/operations/value_propagation.py
```diff
@@ -73,17 +76,21 @@ def propagate(
 
         # sort values by row number
         df = df.sort_values(**self._row_sorting)
-        df = self._add_partition_per_data_columns(df)
+        df = self._add_partition_per_data_columns(df, window_group=window_group)
```
KathiaBarahona, line 79: Can't comment on the line, but seems to me we should check if `df` already has a groupby in line 73. Seems to me it will break otherwise

KathiaBarahona (changes requested):
As I mentioned in other comment. I think that the correct format if we want to apply propagation is grouping by the df first and later call fillna (pandas syntax). (`df.groupby(by='x').fillna(method='ffill')`)

I think a way to do this is just extracting the index columns from the df and use that as `window_group` instead:
``` python
    def __init__(self, df: DataFrame, method: str):
        self._df = df
        self._method = ValuePropagationMethod.get_method(method)
        
    def propagate(
        self,
        sort_by: Optional[Union[str, Sequence[str]]] = None,
        ascending: Union[bool, List[bool]] = True,
    ):
        df = self._df.copy()

        gb = []
        if df.group_by:
            gb = list(df.group_by.index.keys())
            df = df.copy_override(
                series={s.name: s.copy_override(group_by=None) for s in df.data.values()},
                group_by=None,
            )
        if sort_by:
            df = df.sort_values(by=gb + sort_by, ascending=ascending)
        elif gb:
            df = df.sort_values(by=gb, ascending=ascending)
        elif not df.order_by:
            raise Exception('dataframe must be sorted in order to apply forward or backward fill.')

        # Number each sorted row this way we avoid issues when sorting NULL values
        # since bfill is ffill with reversed sort (this might generate different results)
        base_series = df[df.data_columns[0]]
        # order nulls last in window function, else numbering will be wrong
        window = df.groupby(by=gb).window(na_position='last')
        df[self.ROW_NUMBER_SERIES_NAME] = base_series.window_row_number(window)
        df = df.materialize(node_name='numbered_fillna')

        # sort values by group_by indexes and row number
        by = gb + [self.ROW_NUMBER_SERIES_NAME]
        ascending = self._method == ValuePropagationMethod.FORWARD_FILL

        df = df.sort_values(by=by, ascending=ascending)
        df = self._add_partition_per_data_columns(df, window_group=gb)

        # sorting is lost due to materialization, still requires it
        df = df.sort_values(by=by, ascending=ascending)
        df = self._propagate_first_value_per_partition(df, window_group=gb)

        # sort values, this way we respect the provided order by
        df = df.sort_values(by=by)
        return df[self._df.data_columns].materialize(node_name='nafilled')
```

Current implementation breaks if the df is already grouped, so I think its better to fix that now since this PR intends to consider the groupby anyways

bach/bach/operations/value_propagation.py
KathiaBarahona: nvm, I think this param is better to be removed. I explain why in another comment

hendrik-obj:
ok so now I implemented a groupby parameter to ffill/bfill. Couldn't mimic the regular pandas syntax (df.groupby('some_column').ffill()) without big modifications to core Bach, as df.groupby('some_column') doesn't have the same index as df (which would be needed for the correct output of ffill with groups). 

Therefore this pr is back to draft, let's discuss.

KathiaBarahona (approved):
had to approve my own changes as I cannot unblock the PR

hendrik-obj (commented):
changes are good for me. more generic approach providing a `WrappedPartition`

vard-obj (approved):
LGTM

PR #1362, opened 2022-10-14
Use `sphinx-exec-code` for SQL output generation in example notebooks

ivarpruijn:
This PR is a prototype to use extension [`sphinx-exec-code`](https://sphinx-exec-code.readthedocs.io/en/latest/configuration.html) ([repo](https://github.com/spacemanspiff2007/sphinx-exec-code)) to simply show the output of `display_sql_as_markdown()` in example notebooks. This way, we don't have to update the SQL output on most Bach updates, as it's basically already tested in the `doctests` before it.

Downsides of this approach: 
- Some duplicate code/execution: as `sphinx-exec-code` cannot use the preceeding `doctest` output, the model to output has to be re-executed from scratch.
- Maintenance: If the model to output is changed earlier on in the notebook, this should also be done for the output block, which can be forgotten.

I think this does way up against having to update all SQL outputs for Bach updates though.

Alternative options using `sphinx-doctest`, which will take more time to prototype:

- Subclass doctest's [DocTestRunner or OutputChecker classes](https://docs.python.org/3/library/doctest.html#doctestrunner-objects), and overriding their methods to write results to the document, based on some custom `optionflag`. Much like [sphinx-exec-code](https://github.com/spacemanspiff2007/sphinx-exec-code/blob/main/src/sphinx_exec_code/sphinx_exec.py) does.
- Write a custom `doctest` implementation that does the above: https://github.com/pytest-dev/pytest/blob/main/src/_pytest/doctest.py.


bach/docs/source/example-notebooks/explore-data.rst
```diff
@@ -260,111 +260,22 @@ The SQL for any analysis can be exported with one command, so you can use models
...
-	            WHERE
-	       ORDINALITY = jsonb_array_length("location_stack")
-	          ) || (CASE WHEN jsonb_array_length("location_stack") > 1 THEN ' located at ' || (SELECT string_agg(replace(regexp_replace(value ->> '_type', '([a-z])([A-Z])', '\1 \2', 'g'), ' Context', '') || ': ' || (value ->> 'id'), ' => ') FROM jsonb_array_elements("location_stack") WITH ORDINALITY WHERE ORDINALITY < jsonb_array_length("location_stack") ) ELSE '' END),
-	          "event_type"
-	<BLANKLINE>
+.. exec_code::
+	:language: jupyter-notebook
+	:language_output: jupyter-notebook-out
+
+	# --- hide: start ---
```
jansentom, line 267: It's unfortunate that this has to redo everything already prepared above, but I see no way around it, as sphinx-exec-code runs in a new interpreter, and has no way to share code with the rest of the file. 
What you could do is write the markup-SQL to a temporary file in the tests above in a `.. testcleanup::` section, and print that stuff in `.. exec_code::`. That makes sure there is no discrepancy between both (and I'd really like that!)
This does require that doctests run and complete before exec-code.

ivarpruijn: Tried your last suggestion but unfortunately failed in https://github.com/objectiv/objectiv-analytics/pull/1362/commits/2ec84d6a315a190226f80de1345f64a5ab5b59fa. 

The problem is that if we run the doctests before exec-code (which runs with the Sphinx/Docusaurus build), for some reason the custom directives we use there to set frontmatter (for the documents' title, position, videos, etc.) get ignored, messing up titles, ordering in the menu, etc. I think the doctests strip out the custom directives or something, but I haven't been able to debug that yet.

Wdyt about using this current solution (with a small risk of having discrepancies in the future), and logging an issue to try the alternatives I wrote in the description of this PR?

notebooks/model-hub-demo-notebook.ipynb
```diff
@@ -333,7 +333,8 @@
     "df_selection = df[(df.event_type == 'PressEvent') &\n",
     "                  (df.user_id.isin(converted_users))]\n",
     "# calculate the number of PressEvents before conversion per session\n",
-    "presses_per_session = df_selection[df_selection.conversion_count == 0].groupby('session_id').session_hit_number.count()"
+    "presses_per_session = df_selection[df_selection.conversion_count == 0].groupby('session_id').session_hit_number.count()\n",
+    "presses_per_session_pdf = presses_per_session.to_pandas()"
```
ivarpruijn, line 337: Improves performance by not having to re-execute the query later on (see below)

ivarpruijn:
@jansentom Please review. I'll apply this to the marketing-analytics, product-analytics and funnel-discovery notebooks in different PRs.

ivarpruijn:
> 

@

> @jansentom Please review. I'll apply this to the marketing-analytics, product-analytics and funnel-discovery notebooks in different PRs.

@jansentom I lied, I also applied this to the last three notebooks in this PR. Please review, thanks.

PR #1366, opened 2022-10-17
Funnel conversion

vard-obj:


modelhub/modelhub/aggregate.py
```diff
@@ -530,3 +531,202 @@ def drop_off_locations(data: bach.DataFrame,
...
+    #     result_df['full_conversion_rate'] = (result_df['n_users_completed_step'] / n_users_start).round(2)
+    #
+    #     columns = ['step', 'n_users', 'step_conversion_rate',
+    #                'full_conversion_rate', 'dropoff_percentage']
+    #
+    #     return result_df[columns]
+
+    def funnel_conversion(self,
+                          data: bach.DataFrame,
+                          groupby: Union[List[Union[str, Series]], str, Series] = 'user_id',
```
hendrik-obj, line 633: not sure if we should now support anything other that user_id (also below there is some user_id hardcoded in the function)

hendrik-obj (changes requested):
I think there can be some optimizations here and there, but left two main comments about functionality. If those are addressed, and the docs are clear on how to use this function, the results look good.

modelhub/modelhub/aggregate.py
```diff
@@ -530,3 +531,202 @@ def drop_off_locations(data: bach.DataFrame,
...
+        column = location_stack or data['location_stack']
+        if type(column) == str:
+            column = data[column]
+        data['location'] = column
+        if type(column) == SeriesLocationStack:
+            # extract the nice name per event
+            data['location'] = column.ls.nice_name
+        location = 'location'
+
+        df_root_user = data.drop_duplicates(subset=[location, 'user_id'], keep='first')
```
hendrik-obj, line 675: should be sorted by session_id and session_hit_number

modelhub/modelhub/aggregate.py
```diff
@@ -530,3 +531,202 @@ def drop_off_locations(data: bach.DataFrame,
...
+
+        # full_conversion_rate
+        n_users_start = result.sort_values('n_users', ascending=False).\
+            materialize(limit=1)['n_users'].head(1).iloc[0]
+        result['full_conversion_rate'] = result['n_users_completed_step'] / n_users_start
+        result['full_conversion_rate'] = cast(SeriesFloat64, result['step_conversion_rate']).round(2)
+
+        columns = [location, 'n_users', 'n_users_completed_step', 'step_conversion_rate',
+                   'full_conversion_rate', 'dropoff_rate']
+        return result[columns]
```
hendrik-obj, line 732: sort descending by users that entered each step

modelhub/modelhub/aggregate.py
```diff
@@ -530,3 +531,93 @@ def drop_off_locations(data: bach.DataFrame,
...
+        # merging n_users with n_users completed_step and n_users for drop-offs
+        result = step_visitors_df.merge(step_completed_df,
+                                        left_on=location,
+                                        right_on=f'{location}_step_1',
+                                        how='left').reset_index(drop=True)
+        result['n_users_completed_step'] = result['n_users_completed_step'].fillna(0)
+
+        # n_users drop-offs
+        result = result.merge(dropoff_df, left_on=location, right_on='location', how='left')
+        result['dropoff'] = result['dropoff'].fillna(0)
```
hendrik-obj, line 607: 'drop-off' can be calculated as 'n_users' - 'n_users_completed_step'. Then 'drop-off' can be divided by the 'n_users_start'. If we do that, drop_off_locations and therefore this merge is not necessary anymore.

modelhub/modelhub/aggregate.py
```diff
@@ -530,3 +531,92 @@ def drop_off_locations(data: bach.DataFrame,
             result['percentage'] *= 100
             result = result.sort_values(by='percentage', ascending=False)
         return result
+
+    def funnel_conversion(self,
+                          data: bach.DataFrame,
+                          location_stack: LocationStackType = None) -> bach.DataFrame:
```
hendrik-obj, line 537: shall we call it `funnel_steps_column`, or something like that, a column that can be a SeriesLocationStack or SeriesString (but `funnel_steps_column` itself can also be just a `str`)?

Also maybe we should not accept `None` for this function, as I think it does not make a lot of sense in most cases.

modelhub/modelhub/aggregate.py
```diff
@@ -530,3 +531,92 @@ def drop_off_locations(data: bach.DataFrame,
...
+        """
+        Calculates conversion numbers for all locations stacks in the `data`.
+        N.B. Filter the dataframe beforehand if you don't want to include all location stacks .
```
hendrik-obj, line 540:
```suggestion
        N.B. Filter the dataframe beforehand to filter down to the funnel locations.
```

modelhub/modelhub/aggregate.py
```diff
@@ -530,3 +531,92 @@ def drop_off_locations(data: bach.DataFrame,
...
+        column = location_stack or data['location_stack']
+        if type(column) == str:
+            column = data[column]
+        data['location'] = column
+        if type(column) == SeriesLocationStack:
+            # extract the nice name per event
+            data['location'] = column.ls.nice_name
+        location = 'location'
+
+        df_root_user = data.sort_values(['session_id', 'session_hit_number']).\
```
hendrik-obj, line 582: can we use `check_objectiv_dataframe` or the decorator to check if all columns that are needed for this function are in `data`?

hendrik-obj (commented):
functionally all good. Some small comments regarding names and docs. Tests should pass. Then good

modelhub/modelhub/aggregate.py
```diff
@@ -530,3 +531,92 @@ def drop_off_locations(data: bach.DataFrame,
...
+        the number of unique users who completed the step (the step is considered
+        completed if the user went to any other step of the funnel),
+        the conversion rate to completing the step, the conversion rate to completing the step
+        when looking at all users who started the funnel (= the 'full' conversion rate),
+        and the fraction of the users in the funnel dropping out at the given step.
+
+        N.B. We assumed that the funnel direction is always the same.
+        The implementation of VisibleEvents makes for the most accurate calculation
+        of the conversion numbers, as the number of users as well as the conversion rate
+        is based on events on each location stack.
```
ivarpruijn, line 552: Does it make sense to include this here?

modelhub/modelhub/aggregate.py
```diff
@@ -530,3 +531,92 @@ def drop_off_locations(data: bach.DataFrame,
...
+            - A Series with the same base node as `data`.
+
+            If its value is `None`, the whole location stack is taken.
+
+        :returns: :py:class:`bach.DataFrame` with the following columns: `step` (the location considered as a
+            step, e.g. a feature or root location), `n_users` (number of unique users starting the step),
+            `n_users_completed_step` (number of unique users completing the step),
+            `step_conversion_rate` (number of users completing the step / `n_users`), `full_conversion_rate`
+            (number of users completing the step / number of users starting the funnel), and
+            `dropoff_share` (fraction of the users dropping out at the given step).
```
ivarpruijn, line 568: Think this should be updated

modelhub/modelhub/aggregate.py
vard-obj: If the value is None it takes the `feature_nice_name`. I kind of followed the 'standard', as we have the same parameter name in the `drop_off_locations`. In that case, shall we change the parameter name in both functions?

PR #1367, opened 2022-10-17
Improve performance of doctests for marketing analytics notebook

ivarpruijn:
Shortens timeframe to start on August 1st, removes some sections that were not highly necessary, and updates performance for some others. Brings down runtime for doctests from about 4.5 minutes to ~1 minute on my machine.


notebooks/marketing-analytics.ipynb
```diff
@@ -30,8 +30,8 @@
    "outputs": [],
    "source": [
     "# set the timeframe of the analysis\n",
-    "start_date = '2022-06-01'\n",
-    "end_date = None"
+    "start_date = '2022-08-01'\n",
+    "end_date = '2022-08-20'"
```
vard-obj, line 34: Do we really want to fix the end date?

ivarpruijn: No you're right, we don't, good catch! Will fix.

ivarpruijn:
> The `img/docs/example-notebooks/marketing-analytics-funnel-converted.png` file was not updated.

Sharp. I actually didn't update it because the new version doesn't look great, and it's more for illustration purposes in these example notebooks anyway. But I'll have another look if we should update.

notebooks/marketing-analytics.ipynb
ivarpruijn: Fixed in https://github.com/objectiv/objectiv-analytics/pull/1367/commits/e85fd413e33a425d86de5a04cf0ea8c4e30646ae

vard-obj:
I agree, in the end, it is fine for me to leave it as it is.

bach/docs/source/example-notebooks/marketing-analytics.rst
```diff
@@ -345,7 +322,7 @@ Daily conversion rate from marketing
 
 	>>> # calculate daily conversion rate from marketing campaigns overall
 	>>> # divide conversions from campaigns by total daily number of people coming from campaigns 
-	>>> conversion_rate_from_marketing = (conversions_from_marketing_daily / users_from_marketing_daily) * 100
+	>>> conversion_rate_from_marketing = ((conversions_from_marketing_daily / users_from_marketing_daily) * 100).fillna(0.0)
 	>>> conversion_rate_from_marketing.sort_index(ascending=False).fillna(0.0).head(10)
```
hendrik-obj, line 326:
```suggestion
	>>> conversion_rate_from_marketing.sort_index(ascending=False).head(10)

```

hendrik-obj (commented):
1 small comment (rest lgtm) and a suggestion:

We can speed up things quite drastically if where we do both showing the output table _and_ the graphs. Currently that is done in two steps like this, for one example. 
```
users_from_marketing_daily.head()
```
and
```
users_from_marketing_daily.sort_index(ascending=True).to_pandas().plot(kind='bar', figsize=[15,5], title='Daily #users from marketing', xlabel='Day')
```

If we change that to:
```
users_from_marketing_daily_pdf = users_from_marketing_daily.to_pandas()
users_from_marketing_daily_pdf.head()
```
and
```
users_from_marketing_daily_pdf.sort_index(ascending=True).plot(kind='bar', figsize=[15,5], title='Daily #users from marketing', xlabel='Day')
```
The query gets executed only once instead of twice.

ivarpruijn:
> 1 small comment (rest lgtm) and a suggestion:
> 
> We can speed up things quite drastically if where we do both showing the output table _and_ the graphs. Currently that is done in two steps like this, for one example.
> 
> ```
> users_from_marketing_daily.head()
> ```
> 
> and
> 
> ```
> users_from_marketing_daily.sort_index(ascending=True).to_pandas().plot(kind='bar', figsize=[15,5], title='Daily #users from marketing', xlabel='Day')
> ```
> 
> If we change that to:
> 
> ```
> users_from_marketing_daily_pdf = users_from_marketing_daily.to_pandas()
> users_from_marketing_daily_pdf.head()
> ```
> 
> and
> 
> ```
> users_from_marketing_daily_pdf.sort_index(ascending=True).plot(kind='bar', figsize=[15,5], title='Daily #users from marketing', xlabel='Day')
> ```
> 
> The query gets executed only once instead of twice.

Great idea, fixed that in https://github.com/objectiv/objectiv-analytics/pull/1367/commits/0ea5802d15a9db8a20b5ccef9b93d91a8d91b2dc.

With a quick benchmark this brings down execution time by about 20 seconds on my machine.

vard-obj (approved):

PR #1380, opened 2022-10-19
Bach get dtypes from model

thijs-obj:
Changes:
* `get_dtypes_from_model()` execute sql-model with `limit 0`, and look at meta-data from cursor to determine the types. This works across all databases
* Enable tests that now work

### DIFF
--- a/bach/bach/from_database.py
+++ b/bach/bach/from_database.py
@@ -1,7 +1,7 @@
 """
 Copyright 2022 Objectiv B.V.
 """
-from typing import Dict, Tuple, Mapping, Optional, List, Union
+from typing import Dict, Tuple, Mapping, Optional, List, Union, NamedTuple
 
 from sqlalchemy.engine import Engine, Dialect
 
@@ -40,6 +40,11 @@ _DB_TYPE_CODE_TO_DB_TYPE: Mapping[DBDialect, Mapping[Union[int, str], str]] = {
 }
 
 
+class _NameType(NamedTuple):
+    column_name: str
+    db_dtype: str
+
+
 def get_dtypes_from_model(
         engine: Engine,
         node: SqlModel,
@@ -67,7 +72,7 @@ def get_dtypes_from_model(
         description = res.cursor.description
 
     rows = [
-        (row[0], type_code_mapping.get(row[1], row[1]))
+        _NameType(column_name=row[0], db_dtype=type_code_mapping.get(row[1], row[1]))
         for row in description
     ]
     result = _get_dtype_from_db_type(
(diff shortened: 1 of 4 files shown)


bach/bach/from_database.py
```diff
@@ -13,39 +13,69 @@
 from sql_models.sql_generator import to_sql
 from sql_models.util import is_postgres, DatabaseNotSupportedException, is_bigquery, is_athena
 
+_DB_TYPE_CODE_TO_DB_TYPE: Mapping[DBDialect, Mapping[Union[int, str], str]] = {
```
thijs-obj, line 16: For now I think it makes most sense to have this mapping here, close to where it is used.

We could also make it part of the `Series` and support this in `types.py`, but this was simpler. And I don't think we really need to flexibility to have this defined per series or something like that. We can always change things in the future.

bach/bach/from_database.py
```diff
@@ -13,39 +13,69 @@
...
-    return _get_dtypes_from_information_schema_query(
-        engine=engine,
-        query=sql,
+    with engine.connect() as conn:
+        sql = escape_parameter_characters(conn, select_statement)
+        res = conn.execute(sql)
+        # See https://peps.python.org/pep-0249/#description for information what is in description.
+        # Unfortunately, the type_code is not the same as the database type on all databases. Therefore, we
+        # need to map it to the actual database type
+        description = res.cursor.description
```
KathiaBarahona, line 67: smart :rocket:

bach/bach/from_database.py
```diff
@@ -126,13 +156,31 @@ def _get_dtypes_from_information_schema_query(
     If name_to_column_mapping is incomplete, the column names that the query renders will be assumed to be
     series-names.
     """
-    column_to_name_mapping = {column: name for name, column in name_to_column_mapping.items()}
     with engine.connect() as conn:
         sql = escape_parameter_characters(conn, query)
         res = conn.execute(sql)
-        rows = res.fetchall()
+        rows = [(row[0], row[1]) for row in res.fetchall()]
```
KathiaBarahona, line 162: **suggestion**, Maybe a NamedTuple for this?

bach/bach/from_database.py
```diff
@@ -126,13 +156,31 @@ def _get_dtypes_from_information_schema_query(
...
-    db_dialect = DBDialect.from_engine(engine)
+def _get_dtype_from_db_type(
+        dialect: Dialect,
+        rows: List[Tuple[str, str]],
+        name_to_column_mapping: Mapping[str, str]
+) -> Dict[str, StructuredDtype]:
+    """
+    Return dictionary mapping series names to dtypes.
+    :param dialect: dialect
+    :param rows: List of tuples. Each tuple: [0] column name, [1] db_dtype
```
KathiaBarahona, line 179: **suggestion** Maybe a NamedTuple for this?

KathiaBarahona (approved):
lgtm! nice improvement

bach/bach/from_database.py
thijs-obj: Yeah I think that's cleaner. Changed in 172cba7f5c3f1ce6a1135f8f58e2de6b3b650e02

PR #1383, opened 2022-10-19
Migrate all imports of assert_equals_data to use the new location

thijs-obj:
change:
* import for `assert_equals_data` previous: `from tests.functional.bach.test_data_and_utils`, new: `from bach.testing`
  * removed the import from `test.functional.bach`, so there is no way that there are any forgotten imports as those would be broken now.
* small tweak of Makefile

### DIFF
--- a/bach/tests/functional/bach/test_data_and_utils.py
+++ b/bach/tests/functional/bach/test_data_and_utils.py
@@ -16,7 +16,6 @@ from bach import DataFrame, Series
 from sql_models.constants import DBDialect
 from sql_models.util import is_bigquery, quote_identifier
 from tests.unit.bach.util import get_pandas_df
-from bach.testing import assert_equals_data
 
 # Three data tables for testing are defined here that can be used in tests
 # 1. cities: 3 rows (or 11 for the full dataset) of data on cities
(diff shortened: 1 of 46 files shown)


modelhub/Makefile
```diff
@@ -11,15 +11,15 @@ tests-bigquery:
 # against Postgres
 	mypy modelhub
 	pycodestyle modelhub
-	pytest --all -n 8 --dist loadgroup --big-query tests_modelhub
+	pytest --big-query -n 8 --dist loadgroup tests_modelhub
```
thijs-obj, line 14: This is a completely unrelated fix. But I had to put it in some PR, so it became this one.

bach/tests/functional/bach/test_data_and_utils.py
```diff
@@ -16,7 +16,6 @@
 from sql_models.constants import DBDialect
 from sql_models.util import is_bigquery, quote_identifier
 from tests.unit.bach.util import get_pandas_df
-from bach.testing import assert_equals_data
```
KathiaBarahona, line 19: :+1:

PR #1384, opened 2022-10-19
Bach[Athena]: Trim trailing zeros after decimal point (Float64 -> String)

KathiaBarahona:


bach/tests/functional/bach/test_series_numeric.py
```diff
@@ -367,3 +367,26 @@ def test_exp(engine) -> None:
         decimal=14
     )
 
+
+def test_astype_to_string(engine):
```
thijs-obj, line 371: Might fit better in `test_df_astype.py`, but I can see how it fits here too.

thijs-obj (commented):
In principle this looks good to me. But something is up with a modelhub test

KathiaBarahona:
> In principle this looks good to me. But something is up with a modelhub test
seems `rtrim` is not a good idea :/ as double values are not consistent. Modelhub was generating some double values with truncated decimals, so had to use a regex instead. Also, splitted the failing test in modelhub, as there were multiple scenarios being tested and it was taking me too long for debugging

thijs-obj:
> seems `rtrim` is not a good idea :/ as double values are not consistent. Modelhub was generating some double values with truncated decimals, so had to use a regex instead. 

After having another look I think it's not so much the inconsistency of the double values, it's just that `rtrim` does not what we want. I am guessing (because the documentation doesn't say anything about it at all [1]) that `rtrim(x, '.0')` removes all `0` and `.` characters at the end of the string. So if the string is `'100000'`, all `'0'` are the end are removed, which gives `'1'`. Similarly `'20.0'` becomes `'2'`


[1] https://prestodb.io/docs/current/functions/string.html

modelhub/tests_modelhub/functional/modelhub/test_model_hub_retention_matrix.py
```diff
@@ -10,17 +10,17 @@
 from bach.testing import assert_equals_data
 
 
-def test_retention_matrix(db_params):
-
+@pytest.mark.parametrize('event_type', ['ClickEvent', None])
+def test_retention_matrix_yearly(db_params, event_type):
```
thijs-obj, line 14: I like splitting these tests, and splitting out the unit-testable parts.

I do have one worry: We now call `get_objectiv_dataframe_test()` much more often. Every time we call that function, takes about 1.5 seconds for BQ and 2.5 seconds for Athena. So the overall tests become quite a bit slower by splitting this. **Question:** Would it be an idea to cache that somehow? We can of course create an issue for that and do it as a separate issue.

thijs-obj (commented):
Code changes look good to me. Did leave a question regarding a concern, but that doesn't have to block merging. Considering most changes are in modelhub test now, perhaps @hendrik-obj or @vard-obj should have a look too?

modelhub/tests_modelhub/functional/modelhub/test_model_hub_retention_matrix.py
KathiaBarahona: I agree, the easiest change is just generating the dataframe based on [TEST_DATA_OBJECTIV](https://github.com/objectiv/objectiv-analytics/blob/f5be440538de1a46879711163a95edc86cf25d4b/modelhub/tests_modelhub/data_and_utils/data_objectiv.py#L63) instead of reading directly from the DBs (this is already tested by the pipelines anyways)

KathiaBarahona: https://github.com/objectiv/objectiv-analytics/issues/1387

vard-obj (approved):
lgtm

PR #1386, opened 2022-10-19
fix import

thijs-obj:
This broke after I merged https://github.com/objectiv/objectiv-analytics/pull/1383, because a bit before that https://github.com/objectiv/objectiv-analytics/pull/1366 was merged which relied on the old import existing.


thijs-obj:
Don't think this really requires review as it's a one line import change. So will merge once test pass. That way the main branch will be OK asap. Let me know if there are any review comments and I will fix it in another branch.

PR #1388, opened 2022-10-19
fix _ types in iglu

borft:
- add _types to old base schema
- support arrays in iglu generator
- regenerate context iglu schemas

### DIFF
--- a/schema/base_schema.json5
+++ b/schema/base_schema.json5
@@ -29,6 +29,13 @@
           "description": "String containing the name of the event type. (eg. ClickEvent).",
           "type": "string"
         },
+        "_types": {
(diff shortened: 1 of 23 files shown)


schema/base_schema.json5
```diff
@@ -143,6 +150,13 @@
         "_type": {
           "description": "A string literal used during serialization. Should always match the Context interface name.",
           "type": "string"
+        },
```
borft, line 153: @sdirosa this breaks the CI for the TS factories... I suggest to just remove the test, wdyt?

sdirosa: I will fix these tests in a separate PR, go ahead and comment them out 👍🏼

borft: cool! merging this

PR #1389, opened 2022-10-19
[Bach/MH] Bump Bach and Modelhub version to v0.0.26

jansentom:


ivarpruijn (approved):
With the change in https://github.com/objectiv/objectiv-analytics/pull/1390, LGTM

PR #1396, opened 2022-10-21
Generators: Iglu schema

sdirosa:
This generator aims at replicating the old one to produce all the files in objectiv_backend/snowplow/iglu


sdirosa:
@borft Let's review this together coming Monday. I've left out one schema that has properties that do not exist in our base schema and we need to think about how, and if, to embed mix/max rules in our base schema or leave it at this.

borft:
@sdirosa I've removed the old iglu schema. Also added a PR here: https://github.com/objectiv/objectiv.io/pull/589 to update the docs. For some reason the diff is not showing all deleted files, but they are all gone in the PR.

sdirosa:
@borft Looks good, I'll merge this for now. We can finalize the docs in the other PR you create.

PR #1397, opened 2022-10-21
ModelHub: Speed up functional tests

KathiaBarahona:
* Created `objectiv_df` fixture, this one reads from hardcoded TEST_SESSIONIZED_DATA instead of calling the pipelines
* Deleted `get_objectiv_dataframe_test` and migrated all functional tests (except pipelines) to use `objectiv_df` fixture instead.

Tests should be much faster as we avoid doing queries for getting db types per tests and any complex window function for context extraction and sessionized data. TEST_SESSIONIZED_DATA is just the result of the pipelines based on `TEST_DATA_OBJECTIV`

Without speed up, 89 tests:
BQ: **~54s**
image

Athena: **~1m 21s**
image

With speed up, 92 tests:
BQ: **~55s**
image

Athena: **~1m 5s**
image


modelhub/tests_modelhub/functional/modelhub/test_model_hub_funnel_discovery.py
```diff
@@ -305,22 +306,23 @@ def test_get_navigation_paths_filtered(db_params) -> None:
     )
 
 
-def test_filter_navigation_paths_conversion(db_params) -> None:
-    df, modelhub = get_objectiv_dataframe_test(db_params, global_contexts=['application'])
+@pytest.mark.parametrize("objectiv_df", ([['application']]), indirect=True)
```
jansentom, line 309: hmm, did not know this!

modelhub/tests_modelhub/functional/conftest.py
```diff
@@ -5,16 +5,39 @@
...
 from _pytest.tmpdir import TempPathFactory
 from sqlalchemy import create_engine
 
 from tests_modelhub.conftest import get_postgres_db_params
-from tests_modelhub.data_and_utils.utils import setup_db
+from tests_modelhub.data_and_utils.utils import setup_db, create_engine_from_db_params
+from tests_modelhub.data_and_utils.data_objectiv import TEST_SESSIONIZED_DATA, GLOBAL_CONTEXTS_IN_CURRENT_TEST_DATA
 
 
+@pytest.fixture()
```
jansentom, line 21: Since a dataframe is immutable (at least for the purpose of the tests here), could we scope it to `session`?

modelhub/tests_modelhub/functional/conftest.py
```diff
@@ -5,16 +5,39 @@
...
+        engine=engine,
+        df=pandas.DataFrame(TEST_SESSIONIZED_DATA),
+        convert_objects=True,
+    )
+    sessionized_df['stack_event_types'] = sessionized_df['stack_event_types'].astype('json')
+    sessionized_df['location_stack'] = sessionized_df['location_stack'].astype('objectiv_location_stack')
+    for gc in GLOBAL_CONTEXTS_IN_CURRENT_TEST_DATA:
+        sessionized_df[gc] = sessionized_df[gc].astype('objectiv_global_context')
+
+    global_contexts = request.param if hasattr(request, 'param') else []
```
jansentom, line 34: this is quite ugly, but it works hey?

jansentom (approved):
cool improvement!

hendrik-obj (approved):

modelhub/tests_modelhub/functional/conftest.py
KathiaBarahona: I don't think its possible, since db_params is function scoped

modelhub/tests_modelhub/functional/conftest.py
KathiaBarahona: I know :/ but there is not nicer way for this based on pytest's docs. But somehow works :woman_shrugging:

PR #1403, opened 2022-10-25
Checklock: Read AWS env variables

KathiaBarahona:
For output comparison we read and write into AWS bucket, this PR just includes enabling aws env variables in checklock


modelhub/tests_modelhub/checklock_holmes/.env_sample
```diff
@@ -17,3 +17,9 @@ METABASE_PASSWORD = "metabase1"
 METABASE_DASHBOARD_ID = '1' # 1-model-hub
 METABASE_DATABASE_ID = '2' # objectiv database
 METABASE_COLLECTION_ID = '2' # 2-object
+
+# AWS
```
jansentom, line 21: Bit confusing with athena above? What about making it slightly more descriptive,
REPORT__AWS_BUCKET_NAME etc maybe?

jansentom (approved):
Just a comment on naming. LGTM otherwise!

modelhub/tests_modelhub/checklock_holmes/.env_sample
KathiaBarahona: true true

KathiaBarahona: renamed here: https://github.com/objectiv/objectiv-analytics/pull/1403/commits/c85e55b18c85b4e02de3fb990835120ab0c77f36

PR #1408, opened 2022-10-25
Validation docs taxonomy

ivarpruijn:
Adds a Docusaurus writer for the refactored validation to generate the taxonomy docs.

To do:

- [x] Differentiate Global & Location Contexts with their own booleans (instead of just ‘isContext’), so we can create proper links to their pages.
- [x] Clean up code.
- [x] Generate overview.md pages (use the Abstracts for this).
- [x] Use proper slugs for overview.md pages.
- [x] Generate reference/overview.md page.
- [x] Documentation on how to update the docs (via the schema).

Related objectiv.io changes: https://github.com/objectiv/objectiv.io/pull/590


schema/generators/tsconfig.json
```diff
@@ -5,6 +5,9 @@
     "inlineSourceMap": true,
     "removeComments": true,
     "resolveJsonModule": true,
-    "esModuleInterop": true
+    "esModuleInterop": true,
+    "lib": [
+      "ES2020.String"
+    ]
```
sdirosa, line 11: Is this necessary?

ivarpruijn: I'm using a `replaceAll` function on line 139 of the docs template, which TS complaints about, and tried to fix it this way. Doesn't seem to work either now that I'm running it again. Shall I just remove this then?

schema/generators/templates/docs.template.ts
```diff
@@ -4,149 +4,244 @@
...
+	name: string;
+	description: string;
+	type: string;
+	internal: boolean;
+	optional: boolean;
+	nullable: boolean;
+	items: {
+		type: string;
+	};
+};
```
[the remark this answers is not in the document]
ivarpruijn: What's off? Does this fix it? https://github.com/objectiv/objectiv-analytics/pull/1408/commits/03f2ce8a20914f50d2f50771374abdf99a293339

schema/generators/tsconfig.json
sdirosa: You could use `.replace` with a regex, something like:

```ts
let name = '**' + p.name.replace(/_/g, '\\_')
```

schema/generators/templates/docs.template.ts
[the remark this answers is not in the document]
sdirosa: Doesn't look like it's prettier compliant. Tabs instead of spaces, different indents, etc

In the /generators folder you can run:
```
yarn prettier && yarn generate
```

schema/generators/tsconfig.json
ivarpruijn: Done in https://github.com/objectiv/objectiv-analytics/pull/1408/commits/b99faa0e1bd24b51065fb6ed8b518f31a617e677

schema/generators/templates/docs.template.ts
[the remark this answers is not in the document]
ivarpruijn: right, I actually switched to tabs over spaces, that was the preferred way. Ran the prettifier here: https://github.com/objectiv/objectiv-analytics/pull/1408/commits/88de763c082529579e4ddc1d5ebc50fc8458c7f5

PR #1414, opened 2022-10-26
temp fix for generated location stack schema

borft:


schema/generators/generated/snowplow/io.objectiv/location_stack/jsonschema/1-0-0
```diff
@@ -7,10 +7,16 @@
...
+          "minItems": 1
+        }
+    },
+    "additionalProperties": true,
     "type": "object",
-    "properties": {},
-    "additionalProperties": false,
     "required": [
-        null
+        "location_stack"
```
borft, line 20: @sdirosa this was broken, fixed it manually for now

sdirosa: Thanks, I forgot to double check this one, as I was mainly focused on the contexts. Will check the generator as soon as possible.

schema/generators/generated/snowplow/io.objectiv/location_stack/jsonschema/1-0-0
```diff
@@ -7,10 +7,16 @@
         "format": "jsonschema",
         "version": "1-0-0"
     },
+    "properties": {
+        "location_stack": {
+          "type": "array",
+          "description": "The location stack is an ordered list (stack), that contains a hierarchy of location contexts that deterministically describes where an event took place from global to specific. The whole stack (list) is needed to exactly pinpoint where in the UI the event originated.",
+          "minItems": 1
+        }
+    },
```
sdirosa, line 16: @miike I've modified the generator to produce a valid location stack definition for iglu. See https://github.com/objectiv/objectiv-analytics/pull/1420 .

It is slightly different in syntax, but equivalent - I think.

Main difference being that location_stack.type uses the array notation (e.g ["string"] instead of "string") and the "items" attribute is present as well.

Here is how it looks now.

```json
{
    "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
    "description": "The location stack is an ordered list (stack), that contains a hierarchy of location contexts that deterministically describes where an event took place from global to specific. The whole stack (list) is needed to exactly pinpoint where in the UI the event originated.",
    "self": {
        "vendor": "io.objectiv",
        "name": "location_stack",
        "format": "jsonschema",
        "version": "1-0-0"
    },
    "type": "object",
    "properties": {
        "location_stack": {
            "type": [
                "array"
            ],
            "description": "The location stack is an ordered list (stack), that contains a hierarchy of location contexts that deterministically describes where an event took place from global to specific. The whole stack (list) is needed to exactly pinpoint where in the UI the event originated.",
            "items": [
                "object"
            ],
            "minItems": 1
        }
    },
    "additionalProperties": false,
    "required": [
        "location_stack"
    ]
}
```

PR #1415, opened 2022-10-26
Checklock: Saves and reads history locally 

KathiaBarahona:
Reads locally if AWS Bucket variables are not found


jansentom (approved):
Could not reproduce the problem I mentioned during standup. LGTM!

PR #1416, opened 2022-10-26
Move the logic for create Athena urls to a function

thijs-obj:
Add helper function to construct Athena urls

### DIFF
--- a/bach/bach/utils.py
+++ b/bach/bach/utils.py
@@ -1,6 +1,8 @@
 import base64
 import re
 from typing import NamedTuple, Dict, List, Set, Iterable
+from urllib.parse import quote_plus
+
 from sqlalchemy.engine import Connection, Dialect
 
 from bach.expression import Expression, ColumnReferenceToken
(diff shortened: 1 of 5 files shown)


bach/bach/utils.py
```diff
@@ -188,3 +190,38 @@ def merge_sql_statements(dialect: Dialect, sql_statements: List[str]) -> List[st
         return sql_statements
     combined = '; '.join(sql_statements)
     return [combined]
+
+
+def athena_construct_engine_url(
+    *,
+    aws_access_key_id: str,
+    aws_secret_access_key: str,
```
thijs-obj, line 198: Actually just had a short discussion with Mike. These parameters might not be required. As in some cases (e.g. when running on AWS) it makes more sense to let the library automagically handle this. Will look into this in a bit.

bach/bach/utils.py
```diff
@@ -188,3 +190,56 @@ def merge_sql_statements(dialect: Dialect, sql_statements: List[str]) -> List[st
...
+
+def athena_construct_engine_url(
+    *,
+    aws_access_key_id: str = None,
+    aws_secret_access_key: str = None,
+    region_name: str,
+    schema_name: str,
+    s3_staging_dir: str,
+    athena_work_group: Optional[str] = None,
+    catalog_name: Optional[str] = None
```
KathiaBarahona, line 203: maybe `location` param will be nice to have

KathiaBarahona (approved):

bach/bach/utils.py
thijs-obj: I did an experiment with adding `location`, but if I then did `create table` with no location specified (through `DataFrame.database_create_table()`, then it did not write to the location parameter in the connection string, but to the staging_dir. So I think adding `location` doesn't work? Or am I doing it wrong

thijs-obj:
Still a review thread open. But think OK to merge like this, also because https://github.com/objectiv/objectiv.io/pull/596 relies on this. I'll monitor replies here and make follow-up PR if needed.

PR #1418, opened 2022-10-27
WIP Prototype: taxonomy docs generation with types and rules

ivarpruijn:


schema/generators/templates/parser.ts
```diff
@@ -425,3 +425,122 @@ export function getEvents(options?: {
...
+
+/**
+ * A global map for all supported types, by their name.
+ */
+ const typesMap = new Map();
+
+/**
+ * Process each supported entity and fill the typesMap with an enriched version of them.
+ */
+entityNames.forEach((entityName) => {
```
ivarpruijn, line 437: Could merge this with the loop starting on line 26, but seemed different enough to create a new one just for Types.

schema/base_schema.json
```diff
@@ -129,15 +129,15 @@
 
   "LocationStack": {
     "documentation": [
-      { "description": "The location stack is an ordered list (stack), that contains a hierarchy of location contexts that \ndeterministically describes where an event took place from global to specific. \nThe whole stack (list) is needed to exactly pinpoint where in the UI the event originated.", "type": "text", "target": "primary" }
+      { "description": "The LocationStack is an ordered list (a stack) containing a hierarchy of [LocationContexts](/taxonomy/reference/location-contexts/overview.md), which deterministically describes where in the UI of an application an [Event](/taxonomy/reference/events/overview.md) took place.", "type": "text", "target": "primary" }
```
ivarpruijn, line 132: Not sure if using Markdown in these descriptions creates any conflicts with other generators like for the Validator?

schema/generators/templates/docs.template.ts
```diff
@@ -159,31 +166,86 @@ export type PropertiesDefinition = {
...
+		}
 		
 		// final notes
 		docsWriter.writeLine(admonitionDescription);
 	});
 });
 
-// generate reference/overview.md for all the relevant Abstracts
+// Generate pages for Types (i.e. LocationStack and GlobalContexts)
+[...getTypes()].forEach((type) => {
```
ivarpruijn, line 195: Could merge this with the loop starting on line 26 that parses all Contexts & Events, but the formatting of pages for Types is different enough that I thought a new loop made sense.

schema/generators/templates/docs.template.ts
```diff
@@ -205,51 +267,108 @@ Generator.generate({ outputFile: `${destination}/${outputFile}` }, (writer: Text
...
+	docsWriter.writeLine();
+});
+
+const getRuleSummaries = (rule, ruleSummaries, entity = null) => {
+	// only show if it's about the current Context entity or one of its requiredContexts
+	if(!rule._inheritedFrom) {
+		switch (rule.type) {
+			case 'MatchContextProperty':
+				// TODO: show InputValueContext as an optional Context for InputChangeEvent (and PressEvent?), 
+				// and show its rules on the InputValueContext page as well
```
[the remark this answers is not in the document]
ivarpruijn: Agree, actually just wrote this down in our common doc with cases. Of course any Global Context can be added to it, but especially InputValueContext is relevant, and its Validation Rules do say that ‘InputContext.id should equal InputValueContext.id’, which seems a bit strange since it’s not referenced anywhere.

Maybe we should consider introducing some sort of 'OptionalGlobalContext' relation?

schema/base_schema.json
sdirosa: We shouldn't edit primary text descriptions, instead we should add a new markdown description with the desired target.

schema/generators/templates/docs.template.ts
sdirosa: Agreed. I think having a loop block for different concerns is better. We don't care about speed in these generators, readability is much more important.

schema/generators/templates/parser.ts
sdirosa: I'm not sure I understand the purpose of these additions.
getEntities already supports querying types, the code below is basically the same we already have in the main parser. What's the added value?

schema/base_schema.json
ivarpruijn: I'll fix that up

schema/generators/templates/docs.template.ts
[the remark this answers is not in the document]
sdirosa: > which seems a bit strange since it’s not referenced anywhere.
This requirement is expressed via a validation rule, so you could parse that to produce some output mentioning the requirement

> Maybe we should consider introducing some sort of 'OptionalGlobalContext' relation?
I think we need something like that. Not sure on the naming, perhaps `allowed` instead of `optional`.  Seems also quite strange to allow InputValueContext in any event that's not InputChangeEvent.

You can also swap the condition around and list the events a GlobalContext can be used in combination with. Basically, restrict which events a Global Context supports.

schema/generators/templates/parser.ts
ivarpruijn: Totally right simplified this in https://github.com/objectiv/objectiv-analytics/pull/1418/commits/eb06be6265b517afce997cb88f66d4ce3ef083e1

schema/generators/templates/docs.template.ts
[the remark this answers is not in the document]
ivarpruijn: Let's put this on our list of potential future improvements, to have some form of `AllowedGlobalContext` (and `AllowedLocationContext` that we can specify for an Event.

schema/base_schema.json
ivarpruijn: Done in https://github.com/objectiv/objectiv-analytics/pull/1418/commits/6d427f49f276882cfd4e0e6f59c2286caee2713f

sdirosa: Very nice

schema/generators/templates/docs.template.ts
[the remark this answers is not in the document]
sdirosa: Sounds good 👍🏼

schema/generators/templates/parser.ts
```diff
@@ -425,3 +425,38 @@ export function getEvents(options?: {
...
+      types.push(type);
+    }
+  }
+
+  if (sortBy) {
+    return types.sort((a, b) => a[sortBy].localeCompare(b[sortBy]));
+  }
+
+  return types;
+}
```
sdirosa, line 462: I think we should embed this functionality in getEntities instead of duplicating the filtering, inclusion and exclusion logic.

I can do that for you and push to this PR if that's ok

ivarpruijn: Good idea. Please go ahead!

sdirosa: Just pushed the change here: https://github.com/objectiv/objectiv-analytics/pull/1418/commits/1c2103b57e8669eb13d1e05692518150f14ad180

Also rerun prettier, so there are some extra formatting changes in unrelated files.

PR #1419, opened 2022-10-27
Modelhub tweaks

thijs-obj:
Two changes:
1. Give an earlier error if `get_dtypes_from_table()` returns an empty dictionary. This usually means that the table doesn't exist, but the error we gave was about missing columns. That was not really helpful imo.
2. No longer require `bq_credentials` or `bq_credentials_path` if `db_url` is a BigQuery url. This behaviour was inconsistent, if the environment variable `DSN` was set to a BigQuery url we didn't require the additional parameters, but if the same url was specified as `db_url` we did require it.

The second change was triggered by a question from @borft about why we didn't support 'the normal' way of authenticating. I think he had a point :)

### DIFF
--- a/modelhub/modelhub/modelhub.py
+++ b/modelhub/modelhub/modelhub.py
@@ -89,19 +89,17 @@ class ModelHub:
                        bq_credentials: Optional[str] = None) -> Engine:
         """
         returns db_connection based on db_url.
-        If db_url is for BigQuery, bq_credentials_path or bq_credentials must be provided.
+
+        If db_url is for BigQuery, bq_credentials_path or bq_credentials can be provided.
         When both are given, bq_credentials wins.
         """
         kwargs: Dict[str, Any] = {}
 
         if db_url and re.match(r'^bigquery://.+', db_url):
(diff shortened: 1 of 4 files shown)


modelhub/modelhub/modelhub.py
```diff
@@ -128,6 +126,17 @@ def get_objectiv_dataframe(
         transformations and sets the right data types for all columns. As such, the models from the model hub
         can be applied to a DataFrame created with this method.
 
+        For all databases, except BigQuery, the credentials can be specified as part of `db_url`. For
+        BigQuery the credentials can be set with either `bq_credentials` (primary) or `bq_credentials_path`.
+        Additionally, for all databases it's possible to specify credentials as part of the environment,
+        either as variables, files, or some other method. For more information on specifying the credentials
+        as part of the environment, check the documentation of the specific database vendor:
```
jansentom, line 133: very nice

jansentom (approved):
:+1:

KathiaBarahona (approved):

Issue #1405, opened 2022-10-25
bach `from_pandas()` make `convert_objects=True`


thijs-obj:
We virtually always set `convert_objects=True` when calling `from_pandas()`. 

* Make `convert_objects=True` the default, perhaps even remove the parameter.
* Update all usages to remove (now) useless parameter

See also https://github.com/objectiv/objectiv-analytics/issues/418#issuecomment-1288644385

PR #1423, opened 2022-10-31
Bach simplify `from_pandas()` by making `convert_objects=True` by default

thijs-obj:
addresses issue https://github.com/objectiv/objectiv-analytics/issues/1405


thijs-obj:
assigned Tom as reviewer instead of Kathia. As I guess Kathia is working on Athena support

thijs-obj:
@jansentom said over another channel that he liked this, and was okay to merge. So will go ahead with that

PR #1427, opened 2022-11-01
Bach: NULL UUIDs fix when doing `to_pandas`

KathiaBarahona:
Doing `uuid.UUID(None)` raises the following exception:

```
    raise TypeError('one of the hex, bytes, bytes_le, fields, '
TypeError: one of the hex, bytes, bytes_le, fields, or int arguments must be given
```
I did not applied this fix for empty strings, as Postgres actually raises an exception when having a non UUID string


vard-obj (approved):
lgtm

PR #1429, opened 2022-11-01
SnowPlow transport

sdirosa:
A transport module meant to be used in place of the default transport we ship with our SDKs. It will repackage events in Snowplow's format and send them over using their javascript API.


tracker/core/tracker/src/TrackerEvent.ts
```diff
@@ -42,12 +43,13 @@ export class TrackerEvent implements AbstractEvent, Contexts {
...
    */
   constructor(
-    { __instance_id, _type, _types, id, time, ...otherProps }: TrackerEventConfig,
+    { __instance_id, _type, _types, _schema_version, id, time, ...otherProps }: TrackerEventConfig,
     ...contextConfigs: ContextsConfig[]
   ) {
     this.__instance_id = __instance_id;
     this._type = _type;
     this._types = _types;
+    this._schema_version = _schema_version ?? '1.0.0';
```
sdirosa, line 52: We want to expose the _schema_version, so the SnowplowTransport may access it to populate iglu schema strings

PR #1432, opened 2022-11-02
Add BQ check in the notebooks for temp tables

vard-obj:


KathiaBarahona (approved):
lgtm, tho the `marketing-analytics` notebook takes A LOT to execute and sometimes fails

Issue #1215, opened 2022-09-04
[Bach] support week numbers in `strftime()`


thijs-obj:
See this discussion: https://github.com/objectiv/objectiv-analytics/pull/1204#discussion_r960983964

* Support weeknumbers in `strftime()`
  * Sync with data on which weeknumber to implement (preferably `%W`, see linked discussion?). Check what each databases supports out of the box!
  * Add the supported codes in docstring of `strftime()`
  * Make sure this works the same on all databases.
* Add to existing test
* Add test that combines all edge cases:
  * Years that start on each of the 7 days of the week
  * leap-years and non-leap years
  * dates before and after February 28th/29th
 

Probably mostly a problem on Athena and perhaps Postgres

PR #1434, opened 2022-11-02
Bach strftime support `%V`

thijs-obj:
Added support for `%V` in `strftime()`. This addresses issue https://github.com/objectiv/objectiv-analytics/issues/1215

# Changes
* Postgres, BigQuery: simply added `%V` to the allowed codes
* Athena:
  * Split format string on `%V`
  * non-%V parts generate expressions that use `date_format({}, {})`
  * The sub strings that are `%V` generate expressions that use `format_datetime({}, 'ww')` to get the equivalent value
  * Resulting expressions are concatenated
* Added test that checks %V (plus some other codes) for a lot of dates

# Discussion
Implemented `%V`. The issue mentioned `%W`, but `%V` was easier to do, seemed just as powerful, and was actually also suggested in the earlier discussion [here](https://github.com/objectiv/objectiv-analytics/pull/1204#discussion_r960983964)

### DIFF
--- a/bach/bach/series/series_datetime.py
+++ b/bach/bach/series/series_datetime.py
@@ -74,8 +74,8 @@ class DateTimeOperation:
         code semantics: https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes
 
         The subset of codes that are supported across all databases is:
-            `%A`, `%B`, `%F`, `%H`, `%I`, `%M`, `%R`, `%S`, `%T`, `%Y`, `%a`, `%b`, `%d`, `%j`, `%m`, `%y`,
-            `%%`
+            `%A`, `%B`, `%F`, `%H`, `%I`, `%M`, `%R`, `%S`, `%T`, `%V`, %Y`, `%a`, `%b`, `%d`, `%j`, `%m`,
+            `%y`, `%%`
 
         Additionally one specific combination is supported: `%S.%f`
 
--- a/bach/bach/series/series_datetime.py
+++ b/bach/bach/series/series_datetime.py
@@ -15,7 +15,7 @@ from bach.series import Series, SeriesString, SeriesBoolean, SeriesFloat64, Seri
 from bach.expression import Expression, join_expressions, StringValueToken
 from bach.series.series import WrappedPartition, ToPandasInfo
 from bach.series.utils.datetime_formats import parse_c_standard_code_to_postgres_code, \
-    parse_c_code_to_bigquery_code, parse_c_code_to_athena_code, warn_non_supported_format_codes
+    parse_c_code_to_bigquery_code, warn_non_supported_format_codes, parse_c_code_to_athena_expression
 from bach.types import DtypeOrAlias, StructuredDtype, value_to_series_type, get_series_type_from_dtype
 from sql_models.constants import DBDialect
 from sql_models.util import is_postgres, is_bigquery, DatabaseNotSupportedException, is_athena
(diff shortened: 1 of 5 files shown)


jansentom (approved):
LGTM

jansentom (approved):
cool to have this!

bach/bach/series/utils/datetime_formats.py
```diff
@@ -263,10 +272,84 @@ def parse_c_code_to_athena_code(date_format: str) -> str:
     return ''.join(result)
 
 
+def _split_format_string_V(date_format: str) -> List[str]:
```
KathiaBarahona, line 275: maybe?
```suggestion
def _split_format_string_iso_week_number(date_format: str) -> List[str]:
```

KathiaBarahona: also will be nice to include in docstring the reason why we do this

bach/bach/series/utils/datetime_formats.py
```diff
@@ -263,10 +272,84 @@ def parse_c_code_to_athena_code(date_format: str) -> str:
...
+    evaluate to a different string than the original date_format would with python's strftime() function.
+
+    :param series_expression: The expression of the Date/Time/Timestamp/Timedelta Series that needs to be
+        formatted
+    :param date_format: C date format string
+    """
+    # Athena docs
+    # format_datetime(): https://trino.io/docs/current/functions/datetime.html#java-date-functions
+    # date_format():     https://trino.io/docs/current/functions/datetime.html#mysql-date-functions
+    split_date_formats = _split_format_string_V(date_format)
```
KathiaBarahona, line 328: now that I think, why not just replacing `%V` with `ww` instead of splitting the dateformat? I mean we only use `_split_format_string_V` for athena

bach/bach/series/series_datetime.py
```diff
@@ -98,15 +98,12 @@ def strftime(self, format_str: str) -> SeriesString:
...
-            parsed_format_str = parse_c_code_to_athena_code(format_str)
             if isinstance(series, SeriesTime):
                 # for time series, we should convert the float values into a timestamp,
                 # otherwise date_format won't work
                 series = SeriesTimestamp.from_total_seconds(series.copy_override_type(SeriesFloat64))
 
-            expression = Expression.construct(
-                'date_format({}, {})', series, Expression.string_value(parsed_format_str)
-            )
+            expression = parse_c_code_to_athena_expression(series.expression, format_str)
```
KathiaBarahona, line 106: shouldn't we return expressions for the rest of the parsing functions?  It looks weird having one function returning an expression and the rest a string.

bach/bach/series/utils/datetime_formats.py
thijs-obj: We need to pass them to different formatting functions. `ww` works only with `format_datetime()`, all the other codes work with `date_format()`. That's why we need to split the string, and cannot do a simple replacement. We could do something like `format_datetime(date_format(format_string_with_replacements))`, but then we would need to do a lot of escaping to get `format_string_with_replacements` working for both formatting functions. There would be a lot of edge cases to consider and test.

Or am I misunderstanding what you are suggesting?

bach/bach/series/utils/datetime_formats.py
thijs-obj: Added comments in a8bb012cfe9dd1553e6ac0ad85176baea956b8b5

thijs-obj: I don't really like the proposed renaming. 'V' is a bit vague, but becomes clear from the comments very quickly. 'iso week number' is maybe more concrete, but I don't know of the top of my head that that's '%V'. So when looking at the code in for example `parse_c_code_to_athena_expression()` I'd see us call this function and then there is an if of the form `if sub_df == '%V'`. If this function has `V` in the name, I know that is logically connected, if this is called `iso_week_number` I don't know that.

With the current name anyone can understand the code without knowing what `%V` is, one just needs to know that `%V` is not supported (which is in the comments of the function where we call this function). When we rename this, then one also needs to know that `%V` is iso week number, for it to make sense.

bach/bach/series/series_datetime.py
thijs-obj: We could. Just opted for minimizing the diff and work, instead of maximizing consistency. Seemed best trade-off at the time to me, you think not?

thijs-obj:
Merging this, as it has been open in this state for a while. Feel free to reply to open comment threads, or open new threads. I'll fix things in a follow up PR in that case.

PR #1443, opened 2022-11-03
Notebook tweaks and fix conversion event in marketing analytics notebook

hendrik-obj:
- [x] make notebooks run on bq
- [x] select correct conversion event in marketing notebook
- [x] update docs accordingly

### DIFF
--- a/notebooks/model-hub-logistic-regression.ipynb
+++ b/notebooks/model-hub-logistic-regression.ipynb
@@ -41,7 +41,7 @@
    "source": [
     "# set the timeframe of the analysis\n",
     "start_date = '2022-03-01'\n",
-    "end_date = '2022-06-30'"
+    "end_date = None"
    ]
   },
   {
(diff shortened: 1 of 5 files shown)


notebooks/model-hub-logistic-regression.ipynb
```diff
@@ -41,7 +41,7 @@
    "source": [
     "# set the timeframe of the analysis\n",
     "start_date = '2022-03-01'\n",
-    "end_date = None"
```
jansentom, line 44: are you sure?

ivarpruijn: A while ago we wanted to not have an end_date, so that users who would take the notebook to test on their own data would use the latest data. And then we limit the timeframe only for doctests.

hendrik-obj: The model doesn't converge on our data set without an end date. So better to have a dataset that you have some degree of control over. But alas, also fine to leave it `None`.

jansentom: in that case, let's cap it indeed.

ivarpruijn (approved):
LGTM. Corresponding objectiv.io PR: https://github.com/objectiv/objectiv.io/pull/601

PR #1445, opened 2022-11-04
Better conversion

hendrik-obj:
modelhub contained data related to a data frame. This will be decoupled with this pr: adding a conversion events with `add_conversion_event` will return a SeriesBoolean (that can be set as a column to the data frame it is created from), and not add data to a property of the model hub instance.

### DIFF
--- a/modelhub/modelhub/aggregate.py
+++ b/modelhub/modelhub/aggregate.py
@@ -205,6 +205,7 @@ class Aggregate:
             # required by Map.conversions_in_time and Map.conversions_counter
             'session_id', 'moment', 'event_type',
         ],
+        include_series_from_params=['name'],
         required_global_contexts=['application']
     )
     def top_product_features_before_conversion(self,
--- a/modelhub/modelhub/map.py
+++ b/modelhub/modelhub/map.py
@@ -109,7 +109,7 @@ class Map:
         is_new_user_series = is_new_user_series.copy_override_type(bach.SeriesBoolean)
         return is_new_user_series.copy_override(name='is_new_user').materialize()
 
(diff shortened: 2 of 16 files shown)


modelhub/modelhub/aggregate.py
```diff
@@ -217,7 +218,7 @@ def top_product_features_before_conversion(self,
         combining several models from the model hub.
 
         :param data: :py:class:`bach.DataFrame` to apply the method on.
-        :param name: label of the conversion event.
+        :param name: the name of the SeriesBoolean conversion column.
```
KathiaBarahona, line 221: maybe `conversion_event_series_name`. IMO  `name` is too broad.

modelhub/modelhub/map.py
```diff
@@ -294,20 +291,11 @@ def _calculate_is_conversion_event(self, data: bach.DataFrame, name: str, **kwar
 
         series_to_calculate = _CalculatedConversionSeries.IS_CONVERSION_EVENT
         self._check_conversion_dependencies(data, series_to_calculate)
-
-        if name not in self._mh._conversion_events:
+        if name not in data.data_columns:
             raise KeyError(f"Key {name} is not labeled as a conversion")
```
KathiaBarahona, line 295: I don't think this message makes sense anymore

KathiaBarahona (approved):
just a couple of comments. Nice improvement :rocket: this removes lots of joins

ivarpruijn:
Updated the example notebooks in this PR, resulting output is in https://github.com/objectiv/objectiv.io/pull/606

